PR #3, opened 2026-04-18
specs: 000 — system overview

r12f:
## Summary

Adds the foundational `specs/000-overview.md` per parent epic **DPU-4** and issue **DPU-6**. This spec must merge before any feature spec.

It covers:

- **Mission, scope, non-goals** — DPU-resident flow telemetry consuming ERSPAN III mirrors; v1 explicitly excludes inline steering, L7 parsing, multi-DPU clustering, long-term storage, and a GUI.
- **Component map** — `infmon-backend` (C/C++ VPP plugin, gtest), `infmon-frontend` (Rust, cargo test), `infmon-cli` (Rust, cargo test), `tests/` (E2E pcap replay, not in CI), with an ASCII diagram.
- **Data flow** — ERSPAN-mirrored packets → backend match & flow-table update → frontend snapshot/aggregate → exporters.
- **Repo layout** and **language/test conventions** — clang-format + gtest for C/C++; rustfmt + clippy + cargo test for Rust; Conventional Commits + DCO sign-off.
- **Build & release model** — single `infmon_<ver>_arm64.deb` for v1 on BlueField-3 / Ubuntu 22.04 arm64.
- **Glossary** — ERSPAN III, VPP, DPDK, flow / flow tracker, snapshot, aggregate, exporter, collector, DCO, etc.
- **Spec template** — `# NNN — Title / Owners / Context / Goals & Non-goals / Design / Interfaces / Test plan / Open questions`.
- **Open questions** explicitly deferred to specs 002–005 with proposed defaults.

## Test plan

- [x] Spec is a markdown doc only — no code changes; CI build/test jobs are unaffected.

## Related

- Epic: DPU-4 (InFMon)
- Closes: DPU-6

cc @banidoru for sign-off.

### DIFF
--- a/specs/000-overview.md
+++ b/specs/000-overview.md
@@ -1,0 +1,344 @@
+# 000 — System Overview
+
+## Version history
+
+| Version | Date       | Author      | Changes |
+| ------- | ---------- | ----------- | ------- |
+| 0.1     | 2026-04-18 | Riff (r12f) | Initial draft. Establishes mission, scope, component map, repo layout, build/release model, glossary (incl. flow / flow-rule), and pointers to the canonical spec template at [`TEMPLATE.md`](TEMPLATE.md). |
+
+---
+
+## Context
+
+Modern DPU-equipped servers (BlueField-3 and similar) terminate large volumes
+of east-west and north-south traffic at line rate. Operators need per-flow
+visibility — packets, bytes, TCP flags, RTT estimates, drops — without
+sacrificing data-plane throughput. Existing tools (sFlow, IPFIX from kernel
+paths, vendor-specific telemetry) either sample too aggressively, run on the
+host CPU, or are tied to one ASIC.
+
+InFMon is a flow-telemetry service that runs **on the DPU itself**, on top of
+VPP/DPDK, consumes ERSPAN III mirrored copies of production traffic, builds
+flow records in user-space, and exposes them through a local frontend and
+standard exporters. Because it never sits inline, it cannot drop production
+traffic, and because it runs on the DPU's ARM cores, it offloads telemetry
+work entirely off the host.
+
+## Mission
+
+Provide a high-throughput, low-overhead, **DPU-resident flow telemetry
+service** that turns mirrored packets into accurate, queryable flow records
+and feeds them to operator tooling.
+
+## Scope
+
+In-scope for v1:
+
+- Ingest ERSPAN III mirrored packets via VPP on BlueField-3 ARM cores.
+- Parse L2–L4 (Ethernet, VLAN, IPv4/IPv6, TCP/UDP/ICMP).
+- Maintain a flow table keyed by 5-tuple (+ VRF/VNI when present — see note
+  below on extraction source).
+- Maintain per-flow counters: packets, bytes, first/last seen, TCP flag union,
+  and a deliberately narrow TCP RTT estimate for v1: handshake RTT only
+  (time delta between observed `SYN` and matching `SYN-ACK` on the reverse
+  direction of the same flow). Mid-stream / data-segment RTT estimation from
+  mirrored traffic is out of scope for v1 and is deferred to a later spec.
+
+> **Note on VRF/VNI source.** When the inbound packet is ERSPAN III, the
+> `VRF/VNI` field in the flow key is taken from the ERSPAN III header itself
+> (session ID / VRF-ID / SGT fields carried in the ERSPAN III shim). InFMon
+> v1 does **not** parse inner tunnel encapsulation (VXLAN, GENEVE, inner
+> GRE) to derive a VNI; doing so would exceed the L2–L4 scope above. If the
+> mirrored payload happens to be a VXLAN frame, the VNI is treated as part
+> of the opaque inner payload, not extracted into the flow key. This will be
+> revisited in spec 002 (parser).
+- Expose a snapshot/aggregate API consumed by `infmon-frontend`.
+- Ship exporters for at least one standard format (IPFIX or OpenTelemetry —
+  decided in spec 004).
+- A CLI (`infmon-cli`) for inspection, debugging, and admin.
+- Packaging as a `.deb` for **arm64 / aarch64**, single artifact for v1.
+
+Out-of-scope for v1 (non-goals):
+
+- Inline packet steering or modification.
+- Stateful application-layer parsing (TLS SNI, HTTP, gRPC, etc.).
+- Multi-DPU clustering / federated aggregation.
+- A long-term storage backend (InFMon is a producer, not a TSDB).
+- A GUI dashboard. The frontend exposes APIs; visualization is downstream.
+- x86 host builds (portable in principle; not a v1 deliverable).
+
+## Component Map
+
+```
+                +-----------------------------------------------+
+                |                  BlueField-3 DPU              |
+                |                                               |
+ERSPAN III ---> |  VPP graph                                    |
+mirrored        |     |                                         |
+packets         |     v                                         |
+                |  +-------------------+                        |
+                |  | infmon-backend    |                        |
+                |  | (C/C++ VPP plug)  |                        |
+                |  | - decap ERSPAN    |                        |
+                |  | - parse L2–L4     |                        |
+                |  | - flow table      |                        |
+                |  | - counters        |                        |
+                |  +---------+---------+                        |
+                |            |  shared rings / SHM              |
+                |            |  (snapshot/aggregate API,        |
+                |            |   backend -> frontend only)      |
+                |            v                                  |
+                |  +-------------------+    +---------------+   |
+                |  | infmon-frontend   |--->| exporters     |---+--> collectors
+                |  | (Rust)            |    | (IPFIX/OTel)  |   |     (off-DPU)
+                |  | - aggregate       |    +---------------+   |
+                |  | - serve API       |                        |
+                |  +---------+---------+                        |
+                |            ^                                  |
+                +------------|----------------------------------+
+                             |  local UDS, loopback only
+                       +-----+------+
+                       | infmon-cli |
+                       |  (Rust)    |
+                       +------------+
+```
+
+| Component         | Language | Test framework | Purpose                                      |
+|-------------------|----------|----------------|----------------------------------------------|
+| `infmon-backend`  | C/C++    | gtest          | VPP plugin: ERSPAN decap, parse, flow table  |
+| `infmon-frontend` | Rust     | cargo test     | Snapshot/aggregate, exporter drivers, API    |
+| `infmon-cli`      | Rust     | cargo test     | Operator/admin CLI over the frontend API     |
+| `tests/`          | mixed    | pytest + pcap  | E2E real-packet replay (see *E2E execution* below) |
+
+> **Backend language note.** "C/C++" here means: production VPP plugin code
+> (graph nodes, parser, flow-table) is **C11**, matching VPP's own conventions
+> and its C-only node registration macros (`VLIB_REGISTER_NODE` etc.). C++17
+> is allowed only inside `backend/tests/` for GoogleTest fixtures and helpers,
+> which link against the plugin's C ABI through `extern "C"` headers. No C++
+> runtime is loaded into the VPP process in production.
+
+> **Frontend → CLI API and security.** The frontend exposes its API to the
+> CLI over a Unix domain socket bound to a path under `/run/infmon/` with
+> `0660` permissions and an `infmon` group. Membership in that group is the
+> v1 authorization model: any local process whose user is in `infmon` may
+> query; others get `EACCES` at connect time. There is no network listener
+> on this socket — it is loopback/UDS only, never TCP. Stronger auth (token,
+> mTLS) is deferred and tracked as an open question; this is called out so
+> operators on a shared DPU understand the trust boundary.
+
+> **E2E execution.** The `tests/` suite is **not** run on every PR because
+> it requires either a physical BlueField-3 with a real ERSPAN source or an
+> emulated DPU rig with replayed pcaps. It is run in two situations:
+> (1) **release-candidate gating** — every `vX.Y.Z-rc*` tag must pass the
+> full E2E suite on the reference rig before promotion to a final tag, and
+> (2) **nightly** on the maintainers' rig against `main`, with failures
+> filed as issues. A manual `workflow_dispatch` job is also provided so a
+> reviewer can request an E2E run on a specific PR when warranted.
+
+## Data Flow
+
+1. **Mirror.** Upstream switch / host vSwitch mirrors selected traffic via
+   ERSPAN III, encapsulated in GRE, to an IP on the DPU.
+2. **Receive.** DPDK PMD on the DPU delivers packets to VPP.
+3. **Decap & dispatch.** A VPP node owned by `infmon-backend` recognizes
+   ERSPAN III, strips the outer GRE/ERSPAN headers, and reinjects the inner
+   frame **into a private InFMon-only sub-graph** for L2–L4 parsing and flow
+   accounting. The decapped inner frame is **never** returned to VPP's
+   standard L2/L3 forwarding graph: InFMon is a passive telemetry consumer
+   and must not cause the DPU to forward, route, or otherwise act on the
+   mirrored copy. Once parsing and counter updates complete, the buffer is
+   freed at the terminal node.
+4. **Match & key.** L2–L4 parser produces a flow key (5-tuple + VRF/VNI when
+   available) and per-packet metadata (length, TCP flags, timestamp).
+5. **Update.** Flow table lookup; on hit, update counters in place; on miss,
+   insert a new flow record (with eviction policy per spec 003 to be written).
+6. **Snapshot.** Frontend periodically reads the backend's exposed snapshot
+   (lock-free / RCU-style; details in spec 002) and builds an aggregate view.
+7. **Export.** Frontend pushes records to configured exporters and serves
+   `infmon-cli` queries against the aggregate.
+
+## Repo Layout
+
+```
+InFMon/
+├── README.md
+├── LICENSE                 # Apache-2.0
+├── specs/
+│   ├── 000-overview.md     # this file
+│   ├── TEMPLATE.md         # canonical copy of the spec skeleton
+│   └── NNN-<slug>.md       # one spec per accepted feature
+├── backend/                # infmon-backend (C/C++ VPP plugin)
+│   ├── src/
+│   ├── include/
+│   ├── tests/              # gtest unit tests
+│   └── CMakeLists.txt
+├── frontend/               # infmon-frontend (Rust)
+│   ├── src/
+│   ├── tests/
+│   └── Cargo.toml
+├── cli/                    # infmon-cli (Rust)
+│   ├── src/
+│   └── Cargo.toml
+├── tests/                  # E2E, real-packet replay (NOT in CI)
+│   ├── pcaps/
+│   └── scenarios/
+├── packaging/
+│   └── debian/             # aarch64 .deb build files
+└── .github/workflows/      # CI: build + unit tests for all components
+```
+
+## Language & Test Conventions
+
+- **C/C++ (backend):**
+  - C11 / C++17. Match VPP's existing style (4-space indent, snake_case
+    functions, `vlib_*` / `vnet_*` patterns).
+  - `clang-format` enforced in CI. Headers in `backend/include/`.
+  - Unit tests: GoogleTest (`gtest`). Built only when
+    `-DINFMON_BUILD_TESTS=ON`.
+- **Rust (frontend, cli):**
+  - Stable toolchain pinned via `rust-toolchain.toml`.
+  - `rustfmt` and `clippy -D warnings` enforced in CI.
+  - `cargo test` for unit + integration tests.
+  - Workspace `Cargo.toml` at repo root if/when shared crates appear.
+- **Commits:** Conventional Commits (`feat:`, `fix:`, `docs:`, `refactor:`,
+  `test:`, `ci:`, `chore:`). All commits include `Signed-off-by:` (DCO).
+- **Git history / merge style:** PR branches should be **rebased onto
+  `main`** (not merge-committed) before merge. PRs are merged via GitHub's
+  "Squash and merge" for single-logical-change PRs, or "Rebase and merge"
+  for series of self-contained, individually reviewable commits. Merge
+  commits on `main` are not used; `main` stays linear. Force-pushes are
+  allowed on a PR branch *up to* approval; after approval, prefer additive
+  fixup commits so reviewers can re-read deltas.
+- **Specs:** every feature begins life as `specs/NNN-<slug>.md`. A spec is
+  *accepted* when its PR merges into `main`. Implementation PRs reference the
+  spec number in their description.
+
+## Build & Release Model
+
+v1 target: a single Debian package for **aarch64** (BlueField-3 ARM cores
+running Ubuntu 22.04 / DOCA-supported distro).
+
+- Build host: ARM64 Linux with VPP development headers and a recent stable
+  Rust toolchain. Cross-compilation from x86 is a future nice-to-have, not a
+  v1 requirement.
+- Artifact:
+  - `infmon_<version>_arm64.deb` — installs the VPP plugin (`.so` into VPP's
+    plugin dir), the `infmon-frontend` daemon (systemd unit), and the
+    `infmon-cli` binary into `/usr/bin`.
+  - The package ships its plugin under VPP's standard plugin directory but
+    **does not** force-enable it. VPP's default behavior is to auto-load
+    every `.so` in the plugin dir; to keep InFMon opt-in on shared DPUs
+    that may not need telemetry, the package installs a drop-in VPP config
+    snippet at `/etc/vpp/startup.d/10-infmon.conf` that **disables**
+    autoload by default:
+    ```
+    plugins {
+        plugin infmon_plugin.so { disable }
+    }
+    ```
+    Operators flip `disable` → `enable` (or remove the snippet) on the
+    instances where InFMon should run. The systemd unit for
+    `infmon-frontend` is also installed disabled and must be `systemctl
+    enable --now`'d explicitly.
+- Versioning: SemVer. The first published artifact is `v1.0.0`, cut once
+  specs 000–005 are merged and their implementations are green in CI.
+  **No `v0.x.y` releases are planned**; pre-`v1.0.0` testing happens off
+  tagged builds (branch builds and `-rc` candidates of `v1.0.0`, e.g.
+  `v1.0.0-rc1`). The `v0.x.y` scheme is therefore intentionally not used
+  and should not appear on tags.
+- CI scope:
+  - **Per-PR (GitHub-hosted runners, x86_64):** `infmon-frontend` and
+    `infmon-cli` Rust builds + `cargo test` + `clippy` + `rustfmt`. Spec /
+    docs lint. The backend's C code is **syntax/format-checked** here
+    (`clang-format --dry-run`, header-only compile checks where feasible)
+    but its full build and `gtest` suite are *not* run on x86, because the
+    plugin links against VPP headers that target the deployment ABI.
+  - **Per-PR (self-hosted aarch64 runner with VPP dev headers):** full
+    `infmon-backend` build + `gtest` suite. This job is required for
+    merge. The aarch64 runner is the same class of host described under
+    "Build host" above.
+  - **Manual / nightly (reference rig):** the `tests/` E2E suite, as
+    described in *E2E execution* in the Component Map section.
+
+## Glossary
+
+- **DPU** — Data Processing Unit. Smart-NIC with general-purpose cores
+  (BlueField-3 has ARM Cortex-A78AE cores) plus accelerators.
+- **BlueField-3 (BF-3)** — NVIDIA's third-gen DPU. v1 target platform.
+- **VPP** — Vector Packet Processing. FD.io's user-space packet processing
+  framework built on DPDK; processes packets in vectors through a graph of
+  nodes.
+- **DPDK** — Data Plane Development Kit. User-space NIC drivers and packet
+  I/O primitives that VPP sits on top of.
+- **ERSPAN** — Encapsulated Remote Switched Port Analyzer. A traffic mirror
+  protocol that wraps mirrored frames in GRE so they can be sent across an
+  IP network. **ERSPAN III** is the version with a richer header (timestamps,
+  hardware ID, security group tag).
+- **Flow** — A unidirectional sequence of packets sharing the same key
+  (typically 5-tuple: src IP, dst IP, src port, dst port, protocol; extended
+  with VRF/VNI when present). Flows are produced by **flow-rules**: each
+  flow-rule observes mirrored traffic and generates one flow per distinct
+  key tuple it sees, with its own packet/byte counters.
+- **Flow-rule** — A configured matcher (key-set + limits) that selects
+  packets and emits flows. One flow-rule yields many flows (one per unique
+  key tuple). Flow-rules are managed via `infmon-cli flow-rule {add,rm,list,show}`;
+  the resulting flows are read with `infmon-cli flow {list,show}`.
+- **Flow table** — In-memory data structure that maps a flow key to its
+  current counters and metadata; the backend's core state.
+- **Snapshot** — A point-in-time, read-only view of the flow table that the
+  frontend consumes without blocking the data path.
+- **Aggregate** — Frontend-side reduction of snapshots over a window (e.g.
+  top-N talkers, per-VRF totals).
+- **Exporter** — Component that pushes flow records to an external collector
+  in a standard format (IPFIX, OpenTelemetry metrics/logs, etc.).
+- **Collector** — Off-DPU consumer of exported records.
+- **Frontend** — `infmon-frontend`, the Rust user-space service that owns
+  aggregation, exporters, and the API surface for the CLI.
+- **CLI** — `infmon-cli`, operator/admin tool that talks to the frontend.
+- **DCO / Signed-off-by** — Developer Certificate of Origin attestation
+  required on every commit (`git commit -s`).
+
+## Spec Template
+
+Every subsequent spec lives at `specs/NNN-<slug>.md` and follows the
+skeleton tracked at [`specs/TEMPLATE.md`](TEMPLATE.md). Copy that file
+verbatim when starting a new spec; do not hand-roll the header.
+
+Conventions enforced by `TEMPLATE.md` (and required for every spec):
+
+- **One PR = one row** in the `## Version history` table. When you push
+  fixes addressing review comments on the same PR, **amend the existing
+  row's `Changes` cell** instead of appending a new row per iteration.
+- The `Author` column uses the GitHub display name and handle of the
+  PR author, in the form `Name (handle)` — e.g. `Riff (r12f)`.
+- Cross-spec metadata (`Depends on`, `Related`, `Parent epic`, `Affects`)
+  is optional — include only what applies. **`Depends on` and `Related`
+  entries must be markdown links** to the target spec file (e.g.
+  `[002-flow-tracking-model](002-flow-tracking-model.md)`), never bare
+  names.
+- The following fields are **forbidden** and must not appear in any spec:
+  `Owner` / `Owners`, `Status`, `Reviewer` / `Reviewers`, `Last updated`,
+  `Tracking issue`. The PR itself plus the version-history table cover
+  ownership, status, and timestamps; tracking issues are not used for
+  spec docs.
+
+See [`TEMPLATE.md`](TEMPLATE.md) for the full skeleton and the inline
+explanations of each rule.
+
+## Open Questions
+
+1. **Exporter format for v1** — IPFIX, OpenTelemetry, or both? Decided in
+   spec 004. *Default:* IPFIX first (broadest collector support); OTel as a
+   fast follow.
+2. **Snapshot transport** — shared memory ring vs. local Unix-domain socket
+   vs. gRPC. Decided in spec 002. *Default:* shared-memory ring (lowest
+   overhead on DPU).
+3. **Flow table eviction policy** — LRU, time-based, or hybrid. Decided in
+   spec 003. *Default:* time-based with active/idle timeouts (IPFIX-aligned).
+4. **Distro target** — Ubuntu 22.04 only, or also DOCA's recommended base?
+   Decided in spec 005. *Default:* Ubuntu 22.04 arm64.
+5. **Frontend API authn/authz** — v1 ships with UNIX-group-based access on a
+   UDS (see Component Map note). Should we add a token or mTLS layer for
+   multi-tenant DPU scenarios? Tracked for a post-v1 spec. *Default:* keep
+   group-based for v1; revisit when a concrete multi-tenant requirement
+   lands.
--- a/specs/TEMPLATE.md
+++ b/specs/TEMPLATE.md
@@ -1,0 +1,73 @@
+# NNN — Title
+
+> Canonical spec skeleton. Copy this file to `specs/NNN-<slug>.md` for new
+> specs. Inline copies of this template in other docs (e.g. `000-overview.md`)
+> are for reading convenience only — **this file is authoritative**.
+
+## Version history
+
+| Version | Date       | Author       | Changes        |
+| ------- | ---------- | ------------ | -------------- |
+| 0.1     | YYYY-MM-DD | Riff (r12f)  | Initial draft. |
+
+> **Version-history rules.**
+> - **One PR = exactly one row.** When you push fixes addressing review
+>   comments on the same PR, **amend the existing row's `Changes` cell** —
+>   do **not** add a new row per review iteration.
+> - `Author` is the **GitHub display name and handle** of the PR author,
+>   in the form `Name (handle)` — e.g. `Riff (r12f)`. Co-authors,
+>   reviewers, and bots do not appear here.
+> - `Date` is the date the PR was first opened; it does not change on
+>   review-comment fixes.
+
+## Metadata (optional)
+
+Include only the fields that apply. All cross-spec references must be
+**markdown links** to the target spec file, not bare names.
+
+- **Parent epic:** `DPU-NN` (Multica issue, if any)
+- **Depends on:** [`NNN-<slug>`](NNN-<slug>.md), [`NNN-<slug>`](NNN-<slug>.md)
+- **Related:** [`NNN-<slug>`](NNN-<slug>.md)
+- **Affects:** `<component>`, `<component>` (e.g. `infmon-backend`, `infmon-cli`)
+
+> **Forbidden metadata.** Do **not** add any of the following — they will
+> be stripped on sight:
+> - `Owner` / `Owners` (the spec repo's commit history is the owner record)
+> - `Status` (a spec is "draft" while its PR is open and "accepted" once merged)
+> - `Reviewer` / `Reviewers` (GitHub PR reviewers are the source of truth)
+> - `Last updated` (the version-history table already records this)
+> - `Tracking issue` (specs are not tracked by issues; the PR itself is the unit of work)
+
+---
+
+## Context
+
+What problem are we solving and why now? What constraints apply
+(hardware, performance, compatibility)?
+
+## Goals & Non-goals
+
+- Goal: …
+- Goal: …
+- Non-goal: …
+
+## Design
+
+Detailed approach. Diagrams welcome. Data structures, algorithms,
+ownership, threading/locking model, failure modes.
+
+## Interfaces
+
+APIs, CLIs, file formats, wire formats, config knobs. Include
+exact signatures or schemas.
+
+## Test plan
+
+- Unit tests (what, where, which framework)
+- Integration / E2E (what scenarios, what packet captures)
+- Performance criteria, if any
+
+## Open questions
+
+Numbered list of unresolved items, each with a proposed default
+and the decision-maker.


specs/000-overview.md
```diff
@@ -0,0 +1,262 @@
...
+                |                                               |
+ERSPAN III ---> |  VPP graph                                    |
+mirrored        |     |                                         |
+packets         |     v                                         |
+                |  +-------------------+    shared rings/SHM    |
+                |  | infmon-backend    | <--------------------+ |
+                |  | (C/C++ VPP plug)  |                      | |
+                |  | - decap ERSPAN    |                      | |
+                |  | - parse L2–L4     |                      | |
+                |  | - flow table      |                      | |
```
banidoru: The `shared rings/SHM` arrow (`<--------------------+`) appears to flow from the exporters box back *into* the backend, but the data-flow section describes a unidirectional path (backend → frontend → exporters). This is confusing — consider reversing the arrow or restructuring so the SHM connection clearly flows backend → frontend only, with exporters as a downstream leaf.

specs/000-overview.md
```diff
@@ -0,0 +1,262 @@
+# 000 — System Overview
+
+**Owners:** @banidoru (review), DPU team
```
banidoru: The spec template (line 211) defines valid statuses as `Draft | Accepted | Superseded`, but here the status field contains a process constraint (`must merge before any feature spec`). Consider keeping the status as just `Draft` and moving the merge-ordering requirement into the **Context** section or a separate note.

specs/000-overview.md
```diff
@@ -0,0 +1,262 @@
...
+│   ├── src/
+│   └── Cargo.toml
+├── tests/                  # E2E, real-packet replay (NOT in CI)
+│   ├── pcaps/
+│   └── scenarios/
+├── packaging/
+│   └── debian/             # aarch64 .deb build files
+└── .github/workflows/      # CI: build + unit tests for all components
+```
+
```
banidoru, line 189: CI scope says "builds + unit tests for all three components on every PR", but the build host section (line 131) specifies ARM64 Linux with VPP dev headers. If GitHub Actions CI runs on x86 runners, the backend (a VPP ARM plugin) likely can't build or run its gtest suite there without cross-compilation or emulation. This should be clarified — are backend unit tests actually in the PR CI path, or only in a separate ARM-based job?

specs/000-overview.md
```diff
@@ -0,0 +1,262 @@
...
+- Ingest ERSPAN III mirrored packets via VPP on BlueField-3 ARM cores.
+- Parse L2–L4 (Ethernet, VLAN, IPv4/IPv6, TCP/UDP/ICMP).
+- Maintain a flow table keyed by 5-tuple (+ VRF/VNI when present).
+- Maintain per-flow counters: packets, bytes, first/last seen, TCP flag union,
+  basic RTT samples for TCP.
+- Expose a snapshot/aggregate API consumed by `infmon-frontend`.
+- Ship exporters for at least one standard format (IPFIX or OpenTelemetry —
+  decided in spec 004).
+- A CLI (`infmon-cli`) for inspection, debugging, and admin.
+- Packaging as a `.deb` for **arm64 / aarch64**, single artifact for v1.
```
banidoru, line 59: "VRF/VNI when present" appears in multiple places but it's never specified *where* this info is extracted from — the ERSPAN III header (which carries a platform-specific ID and security group tag) or the inner packet (e.g., VXLAN VNI). This distinction matters for the backend parser design in spec 002/003. Worth a one-liner here noting the expected source, even if details are deferred.

specs/000-overview.md
```diff
@@ -0,0 +1,262 @@
...
+│   ├── tests/
+│   └── Cargo.toml
+├── cli/                    # infmon-cli (Rust)
+│   ├── src/
+│   └── Cargo.toml
+├── tests/                  # E2E, real-packet replay (NOT in CI)
+│   ├── pcaps/
+│   └── scenarios/
+├── packaging/
+│   └── debian/             # aarch64 .deb build files
```
banidoru, line 186: Versioning defines `v0.x.y` for pre-1.0 and then jumps to `v1.0.0` once specs 000–005 merge. Are any `v0.x` releases planned for early testing/feedback, or is the first published artifact `v1.0.0`? If no pre-releases are intended, the `v0.x.y` scheme is dead weight and could be dropped for simplicity.

banidoru (commented):
- Architecture diagram arrow direction for `shared rings/SHM` is misleading — looks like exporters feed back into the backend.
- CI story for ARM64 backend builds is unaddressed — if CI runs on x86, unit-test coverage claim for the VPP plugin needs qualification.
- Status field uses a free-form process note instead of the enum defined in the spec template (`Draft | Accepted | Superseded`).
- VRF/VNI extraction source (ERSPAN III header vs. inner packet) should be called out explicitly to avoid ambiguity in downstream specs.
- Minor: no 0.x release plan described despite pre-1.0 versioning scheme being defined.

specs/000-overview.md
```diff
@@ -0,0 +1,262 @@
...
+                +------------|----------------------|-----------+
+                             |                      |
+                       +-----+-----+         +------v------+
+                       | infmon-cli |        | collectors  |
+                       |  (Rust)    |        | (off-DPU)   |
+                       +------------+         +-------------+
+```
+
+| Component         | Language | Test framework | Purpose                                      |
+|-------------------|----------|----------------|----------------------------------------------|
```
banidoru, line 107: The diagram shows  talking to the . On a shared DPU, this implies the frontend is serving an API. Is there any authentication or authorization on this API? If not, any process on the DPU could potentially query it, which might be a security concern or a resource-contention issue if the CLI is used heavily.

specs/000-overview.md
```diff
@@ -0,0 +1,262 @@
...
+| `infmon-backend`  | C/C++    | gtest          | VPP plugin: ERSPAN decap, parse, flow table  |
+| `infmon-frontend` | Rust     | cargo test     | Snapshot/aggregate, exporter drivers, API    |
+| `infmon-cli`      | Rust     | cargo test     | Operator/admin CLI over the frontend API     |
+| `tests/`          | mixed    | pytest + pcap  | E2E real-packet replay (not in CI)           |
+
+## Data Flow
+
+1. **Mirror.** Upstream switch / host vSwitch mirrors selected traffic via
+   ERSPAN III, encapsulated in GRE, to an IP on the DPU.
+2. **Receive.** DPDK PMD on the DPU delivers packets to VPP.
```
banidoru, line 142: The table notes that the  directory is for E2E tests that are "not in CI". This is also mentioned later. It would be valuable to state *when* or *how* these tests are run. Manually on release candidates? Nightly on a physical rig? The current description leaves a gap in the testing strategy.

specs/000-overview.md
```diff
@@ -0,0 +1,262 @@
...
+  top-N talkers, per-VRF totals).
+- **Exporter** — Component that pushes flow records to an external collector
+  in a standard format (IPFIX, OpenTelemetry metrics/logs, etc.).
+- **Collector** — Off-DPU consumer of exported records.
+- **Frontend** — `infmon-frontend`, the Rust user-space service that owns
+  aggregation, exporters, and the API surface for the CLI.
+- **CLI** — `infmon-cli`, operator/admin tool that talks to the frontend.
+- **DCO / Signed-off-by** — Developer Certificate of Origin attestation
+  required on every commit (`git commit -s`).
+
```
banidoru, line 300: The spec includes a great template for future specs. It might be helpful to also create a  file in the repo. This makes it easier for contributors to copy-paste and reduces the chance of the template in this overview doc becoming stale.

specs/000-overview.md
```diff
@@ -0,0 +1,262 @@
...
+  `test:`, `ci:`, `chore:`). All commits include `Signed-off-by:` (DCO).
+- **Specs:** every feature begins life as `specs/NNN-<slug>.md`. A spec is
+  *accepted* when its PR merges into `main`. Implementation PRs reference the
+  spec number in their description.
+
+## Build & Release Model
+
+v1 target: a single Debian package for **aarch64** (BlueField-3 ARM cores
+running Ubuntu 22.04 / DOCA-supported distro).
+
```
banidoru, line 220: The spec requires a  line, which is excellent for DCO. It might also be worth specifying the desired git history for PRs. For example, should developers rebase their branches to maintain a linear history, or are merge commits acceptable? Defining this now can prevent future inconsistencies.

banidoru (changes requested):
- The  connects directly to the frontend; consider if this API needs auth to prevent unauthorized queries/load.
-  test suite is noted as "not in CI." What is the trigger/cadence for this critical suite?
- Spec template is defined inline; consider creating  for easier reuse.
-  is required, which is great. Should also clarify if PRs must have linear history (rebase) or if merge commits are okay.

specs/000-overview.md
```diff
@@ -0,0 +1,262 @@
...
+## Scope
+
+In-scope for v1:
+
+- Ingest ERSPAN III mirrored packets via VPP on BlueField-3 ARM cores.
+- Parse L2–L4 (Ethernet, VLAN, IPv4/IPv6, TCP/UDP/ICMP).
+- Maintain a flow table keyed by 5-tuple (+ VRF/VNI when present).
+- Maintain per-flow counters: packets, bytes, first/last seen, TCP flag union,
+  basic RTT samples for TCP.
```
banidoru: The flow key mentions "5-tuple (+ VRF/VNI when present)" but the L2-L4 parsing scope (line 37) doesn't mention VXLAN/GRE inner-header parsing that would be needed to extract VNI. If VNI comes from the outer ERSPAN III header's session ID, that should be stated explicitly. If it requires parsing inner encapsulation, that contradicts the "L2-L4 only" scope.

specs/000-overview.md
```diff
@@ -0,0 +1,262 @@
...
+- Expose a snapshot/aggregate API consumed by `infmon-frontend`.
+- Ship exporters for at least one standard format (IPFIX or OpenTelemetry —
```
banidoru, line 56: "basic RTT samples for TCP" — this is a surprisingly non-trivial feature for a v1. Accurate RTT estimation from mirrored (not inline) traffic requires matching request/response packets across flows, handling retransmissions, and dealing with asymmetric mirroring. Consider either scoping this more precisely (e.g., SYN→SYN-ACK only) or deferring it to a later spec.

specs/000-overview.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Data Flow
+
+1. **Mirror.** Upstream switch / host vSwitch mirrors selected traffic via
+   ERSPAN III, encapsulated in GRE, to an IP on the DPU.
+2. **Receive.** DPDK PMD on the DPU delivers packets to VPP.
+3. **Decap & dispatch.** A VPP node owned by `infmon-backend` recognizes
+   ERSPAN III, strips the outer GRE/ERSPAN headers, and reinjects the inner
+   frame into the parsing path.
+4. **Match & key.** L2–L4 parser produces a flow key (5-tuple + VRF/VNI when
```
banidoru, line 151: The component table says `infmon-backend` is "C/C++" but the conventions section (line 127) says "C11 / C++17". VPP plugins are typically pure C. Mixing C++ into a VPP plugin has real implications (name mangling, exception safety, VPP's C-only node registration macros). Worth clarifying whether the plugin itself is C-only with gtest wrappers in C++, or if C++ is used in production code too.

specs/000-overview.md
```diff
@@ -0,0 +1,262 @@
...
+   available) and per-packet metadata (length, TCP flags, timestamp).
+5. **Update.** Flow table lookup; on hit, update counters in place; on miss,
+   insert a new flow record (with eviction policy per spec 003 to be written).
+6. **Snapshot.** Frontend periodically reads the backend's exposed snapshot
+   (lock-free / RCU-style; details in spec 002) and builds an aggregate view.
+7. **Export.** Frontend pushes records to configured exporters and serves
+   `infmon-cli` queries against the aggregate.
+
+## Repo Layout
```
banidoru, line 160: Step 3 says the backend "reinjects the inner frame into the parsing path" — this is ambiguous. Does it mean back into VPP's standard L2/L3 graph (which could cause routing/forwarding side effects), or into a dedicated InFMon-only graph path? For a passive telemetry tool, accidentally forwarding decapped packets would be a serious bug.

specs/000-overview.md
```diff
@@ -0,0 +1,262 @@
...
+├── frontend/               # infmon-frontend (Rust)
+│   ├── src/
+│   ├── tests/
+│   └── Cargo.toml
+├── cli/                    # infmon-cli (Rust)
+│   ├── src/
+│   └── Cargo.toml
+├── tests/                  # E2E, real-packet replay (NOT in CI)
+│   ├── pcaps/
+│   └── scenarios/
```
banidoru, line 184: The `.deb` installs the VPP plugin `.so` into VPP's plugin dir. This means VPP will auto-load it on startup. Should there be a note about load-on-demand configuration (`plugins { plugin infmon_plugin.so { enable } }`) to avoid loading it on VPP instances that don't need it?

banidoru (commented):
- VNI extraction contradicts L2-L4-only parsing scope — needs clarification on source
- TCP RTT estimation from mirrored traffic is complex enough to warrant its own scoping or deferral
- C vs C++ boundary in VPP plugin needs explicit statement (plugin C-only, tests C++?)
- "Reinjects inner frame into parsing path" is ambiguous and risks accidental forwarding
- VPP plugin auto-load behavior should be addressed in packaging section

r12f:

specs/000-overview.md
r12f: Thanks for the review! However, this comment is on `specs/000-overview.md`, which is not part of PR #5 — this PR only adds `specs/002-flow-tracking-model.md` (see the Files Changed tab). The overview spec lives on its own PR (spec 000). Could you re-post this feedback on that PR so it can be tracked against the right file? Resolving here as out-of-scope.

specs/000-overview.md
r12f: Thanks for the review! However, this comment is on `specs/000-overview.md`, which is not part of PR #5 — this PR only adds `specs/002-flow-tracking-model.md` (see the Files Changed tab). The overview spec lives on its own PR (spec 000). Could you re-post this feedback on that PR so it can be tracked against the right file? Resolving here as out-of-scope.

specs/000-overview.md
r12f: Thanks for the review! However, this comment is on `specs/000-overview.md`, which is not part of PR #5 — this PR only adds `specs/002-flow-tracking-model.md` (see the Files Changed tab). The overview spec lives on its own PR (spec 000). Could you re-post this feedback on that PR so it can be tracked against the right file? Resolving here as out-of-scope.

specs/000-overview.md
r12f: Thanks for the review! However, this comment is on `specs/000-overview.md`, which is not part of PR #5 — this PR only adds `specs/002-flow-tracking-model.md` (see the Files Changed tab). The overview spec lives on its own PR (spec 000). Could you re-post this feedback on that PR so it can be tracked against the right file? Resolving here as out-of-scope.

specs/000-overview.md
r12f: Thanks for the review! However, this comment is on `specs/000-overview.md`, which is not part of PR #5 — this PR only adds `specs/002-flow-tracking-model.md` (see the Files Changed tab). The overview spec lives on its own PR (spec 000). Could you re-post this feedback on that PR so it can be tracked against the right file? Resolving here as out-of-scope.

specs/000-overview.md
r12f: Thanks for the review! However, this comment is on `specs/000-overview.md`, which is not part of PR #5 — this PR only adds `specs/002-flow-tracking-model.md` (see the Files Changed tab). The overview spec lives on its own PR (spec 000). Could you re-post this feedback on that PR so it can be tracked against the right file? Resolving here as out-of-scope.

specs/000-overview.md
r12f: Thanks for the review! However, this comment is on `specs/000-overview.md`, which is not part of PR #5 — this PR only adds `specs/002-flow-tracking-model.md` (see the Files Changed tab). The overview spec lives on its own PR (spec 000). Could you re-post this feedback on that PR so it can be tracked against the right file? Resolving here as out-of-scope.

specs/000-overview.md
r12f: Thanks for the review! However, this comment is on `specs/000-overview.md`, which is not part of PR #5 — this PR only adds `specs/002-flow-tracking-model.md` (see the Files Changed tab). The overview spec lives on its own PR (spec 000). Could you re-post this feedback on that PR so it can be tracked against the right file? Resolving here as out-of-scope.

specs/000-overview.md
r12f: Thanks for the review! However, this comment is on `specs/000-overview.md`, which is not part of PR #5 — this PR only adds `specs/002-flow-tracking-model.md` (see the Files Changed tab). The overview spec lives on its own PR (spec 000). Could you re-post this feedback on that PR so it can be tracked against the right file? Resolving here as out-of-scope.

specs/000-overview.md
r12f: Thanks for the review! However, this comment is on `specs/000-overview.md`, which is not part of PR #5 — this PR only adds `specs/002-flow-tracking-model.md` (see the Files Changed tab). The overview spec lives on its own PR (spec 000). Could you re-post this feedback on that PR so it can be tracked against the right file? Resolving here as out-of-scope.

specs/000-overview.md
r12f: Thanks for the review! However, this comment is on `specs/000-overview.md`, which is not part of PR #5 — this PR only adds `specs/002-flow-tracking-model.md` (see the Files Changed tab). The overview spec lives on its own PR (spec 000). Could you re-post this feedback on that PR so it can be tracked against the right file? Resolving here as out-of-scope.

specs/000-overview.md
r12f: Thanks for the review! However, this comment is on `specs/000-overview.md`, which is not part of PR #5 — this PR only adds `specs/002-flow-tracking-model.md` (see the Files Changed tab). The overview spec lives on its own PR (spec 000). Could you re-post this feedback on that PR so it can be tracked against the right file? Resolving here as out-of-scope.

specs/000-overview.md
r12f: Thanks for the review! However, this comment is on `specs/000-overview.md`, which is not part of PR #5 — this PR only adds `specs/002-flow-tracking-model.md` (see the Files Changed tab). The overview spec lives on its own PR (spec 000). Could you re-post this feedback on that PR so it can be tracked against the right file? Resolving here as out-of-scope.

specs/000-overview.md
r12f: Thanks for the review! However, this comment is on `specs/000-overview.md`, which is not part of PR #5 — this PR only adds `specs/002-flow-tracking-model.md` (see the Files Changed tab). The overview spec lives on its own PR (spec 000). Could you re-post this feedback on that PR so it can be tracked against the right file? Resolving here as out-of-scope.

specs/000-overview.md
r12f: Thanks — noting the SHM-arrow direction concern. This comment targets `specs/000-overview.md`, which is not modified by PR #4 (this PR only adds `specs/001-ci-and-precommit.md`). Capturing the suggestion to be addressed in the spec-000 PR / a follow-up so the diagram fix lands in the right file. Resolving here as out-of-scope-to-this-PR.

specs/000-overview.md
r12f: Agreed in principle (Status field should match the template's allowed values; merge-ordering belongs in Context). This comment is on `specs/000-overview.md`, which is not part of PR #4. Will be applied in the spec-000 PR. Resolving as out-of-scope here.

specs/000-overview.md
r12f: Good catch — backend gtest on x86 GH runners vs. ARM64+VPP build host is a real tension. Note that PR #4's `cpp-test.yml` (specs/001 §4.3) runs `cmake --build` + `ctest` inside `ligato/vpp-base:24.02` on amd64 (VPP headers are arch-independent), and `cross-build.yml` (§4.4) handles ARM64 as compile-only. The spec-000 wording ("builds + unit tests for all three components") should be reconciled with that. This thread is on `specs/000-overview.md` which isn't in PR #4's diff; flagging for the spec-000 PR. Resolving here.

specs/000-overview.md
r12f: Agreed a one-liner specifying the VRF/VNI source (ERSPAN III session ID vs. inner VXLAN VNI) would help downstream parser specs. Comment is on `specs/000-overview.md`, not modified by PR #4 — will be addressed in the spec-000 PR. Resolving as out-of-scope here.

specs/000-overview.md
r12f: Reasonable question on `v0.x.y` vs. starting at `v1.0.0`. Comment targets `specs/000-overview.md`, not part of PR #4. Recording for the spec-000 PR. Resolving as out-of-scope here.

specs/000-overview.md
r12f: Authn/authz on the frontend's local API on a shared DPU is a real security concern worth a sentence in the overview. Comment is on `specs/000-overview.md`, not part of PR #4. Flagging for the spec-000 PR / a security-model spec. Resolving here.

specs/000-overview.md
r12f: Agreed — specs/001 §6 in this PR documents the manual workflow (`make e2e` on `r12f-bf3`, future self-hosted nightly out-of-scope). The matching one-liner in `specs/000-overview.md` is what this comment targets; that file isn't in PR #4's diff. Will mirror the §6 wording into spec-000 in its own PR. Resolving here.

specs/000-overview.md
r12f: Good idea — a `specs/_template.md` (or `.github/SPEC_TEMPLATE.md`) avoids drift. Comment is on `specs/000-overview.md`, not modified by PR #4. Filing for the spec-000 PR / a small repo-hygiene change. Resolving here.

specs/000-overview.md
r12f: Note: PR #4's spec/001 §7 already mandates `Require linear history: on` and `Allow force pushes: off`, which implicitly requires rebase (no merge commits on `main`). The DCO/sign-off line that prompted this comment lives in `specs/000-overview.md`, which isn't in PR #4's diff; an explicit cross-reference can be added to spec-000 in its own PR. Resolving here.

specs/000-overview.md
r12f: Agreed this needs to be made explicit (likely ERSPAN III session ID, not inner-encap parsing — otherwise the L2-L4 scope is contradicted). Comment is on `specs/000-overview.md`, not part of PR #4. Flagging for the spec-000 PR. Resolving here.

specs/000-overview.md
r12f: Valid scoping concern for v1 RTT. Comment is on `specs/000-overview.md`, not part of PR #4 — will be addressed in the spec-000 PR (likely scoped to SYN→SYN-ACK or deferred). Resolving here.

specs/000-overview.md
r12f: Real concern (VPP node registration is C-only; mixing C++ has ABI/exception-safety implications). Comment is on `specs/000-overview.md`, not in PR #4's diff. Flagging for the spec-000 PR — likely resolution: plugin in C11, gtest harness in C++17. Resolving here.

specs/000-overview.md
r12f: Important correctness concern — "reinjects into the parsing path" must mean an InFMon-private graph node, never VPP's L2/L3 forwarding graph, since this is a passive telemetry tool. Comment is on `specs/000-overview.md`, not part of PR #4. Flagging for the spec-000 PR. Resolving here.

specs/000-overview.md
r12f: Agreed — the `.deb` should not auto-enable the plugin; load-on-demand via `plugins { plugin infmon_plugin.so { enable } }` is the right default for a shared VPP. Comment is on `specs/000-overview.md`, not in PR #4's diff. Flagging for the spec-000 PR / packaging spec. Resolving here.

specs/000-overview.md
r12f: Good catch — the diagram was misleading. Redrew the Component Map so the SHM/snapshot link is shown unambiguously as backend → frontend (single downward arrow), with exporters as a downstream leaf and the CLI on a separate local UDS path. Fixed in e435ce2.

specs/000-overview.md
r12f: Marking resolved — the offending status line no longer exists in the current version of the file (this comment is on an outdated diff). The current spec has no process constraint embedded in a Status field.

specs/000-overview.md
r12f: Right — the original wording over-promised. Split the CI scope into (a) per-PR x86 GitHub-hosted job that runs Rust builds/tests + clippy/rustfmt + docs lint and only does syntax/format checks for the backend C code, and (b) per-PR self-hosted aarch64 job (with VPP dev headers) that does the full backend build + gtest suite and is required for merge. E2E remains on the reference rig. Fixed in e435ce2.

specs/000-overview.md
r12f: Agreed. Added an explicit note in Scope: VRF/VNI is taken from the ERSPAN III header itself (session ID / VRF-ID / SGT shim fields). v1 does not parse inner tunnel encapsulation (VXLAN/GENEVE/inner GRE) to derive a VNI, since that would exceed the L2–L4 scope. Detailed parser design deferred to spec 002. Fixed in e435ce2.

specs/000-overview.md
r12f: Picked 'drop v0.x.y' over 'keep both schemes' because no v0 releases are actually planned and the dual scheme just adds confusion. Spec now says: first published artifact is v1.0.0; pre-1.0 testing happens off branch builds and v1.0.0-rcN candidates; v0.x.y is intentionally not used. Fixed in e435ce2.

specs/000-overview.md
r12f: Important point. Added an explicit security note: the frontend exposes its API only over a UNIX domain socket under /run/infmon/ with mode 0660 and an 'infmon' group; non-members get EACCES at connect; there is no TCP listener. That is the v1 authz model. Stronger auth (token / mTLS) is now tracked as Open Question #5 for a post-v1 spec since it would be a real architectural addition rather than a doc tweak. Fixed in e435ce2.

specs/000-overview.md
r12f: Agreed, that was a real gap. Added an 'E2E execution' note: the suite runs (1) as a release-candidate gate on every vX.Y.Z-rc tag against the reference rig, (2) nightly against main on the maintainers' rig with failures filed as issues, and (3) on demand via a manual workflow_dispatch job a reviewer can trigger on a specific PR. Fixed in e435ce2.

specs/000-overview.md
r12f: Good idea. Added specs/TEMPLATE.md with the same skeleton, and updated the Spec Template section to point at it as the authoritative copy (with the inline copy kept as a reading aid). Fixed in e435ce2.

specs/000-overview.md
r12f: Added a 'Git history / merge style' bullet to Language & Test Conventions: rebase PR branches onto main (no merge commits into PRs), merge via Squash-and-merge for single-logical-change PRs or Rebase-and-merge for clean multi-commit series, keep main linear, force-push allowed pre-approval but prefer additive fixups after. Fixed in e435ce2.

specs/000-overview.md
r12f: Same fix as the line-46 thread: the spec now explicitly states VNI/VRF for the flow key comes from the ERSPAN III header and that inner-tunnel parsing (which would contradict the L2–L4 scope) is out of scope for v1. Fixed in e435ce2.

specs/000-overview.md
r12f: Picked 'narrow scope to handshake-only' over 'defer entirely' because handshake RTT (SYN → SYN-ACK delta on the same flow's reverse direction) is feasible from mirrored traffic without cross-flow matching or retransmit handling, and is genuinely useful as a baseline. Mid-stream / data-segment RTT estimation is now explicitly deferred. Fixed in e435ce2.

specs/000-overview.md
r12f: Added a Backend language note: production plugin code (graph nodes, parser, flow table) is C11 to match VPP's conventions and its C-only registration macros; C++17 is used only inside backend/tests/ for gtest fixtures, linking the plugin's C ABI through extern "C" headers. No C++ runtime in the VPP process in production. Fixed in e435ce2.

specs/000-overview.md
r12f: Agreed, this was the most safety-critical ambiguity. Reworded step 3 to make clear the decapped inner frame is reinjected into a *private InFMon-only sub-graph* for parsing/accounting only, never returned to VPP's standard L2/L3 forwarding graph, and the buffer is freed at the terminal node. Fixed in e435ce2.

specs/000-overview.md
r12f: Yes — added a packaging note: the .deb installs the .so under VPP's plugin dir but also installs a drop-in /etc/vpp/startup.d/10-infmon.conf that *disables* autoload by default, so InFMon is opt-in on shared DPUs. The infmon-frontend systemd unit ships disabled too, requiring an explicit 'systemctl enable --now'. Fixed in e435ce2.

banidoru (commented):
Re-review (iteration 2):

- All 14 prior review threads already resolved — no unaddressed concerns from the previous round.
- Typo in glossary rename: `flow flow-rule` should be `flow-rule` (duplicate word).
- `specs/TEMPLATE.md` still uses `Owners / Status / Last updated` header fields, but the main spec (000) replaced those with a Version History table — consider aligning the template.
- All other changes (VRF/VNI clarification, RTT scoping, diagram fix, CI split, security model, E2E execution, packaging, versioning, git history, TEMPLATE.md addition) look good.

specs/000-overview.md
```diff
@@ -0,0 +1,353 @@
...
+- **DPDK** — Data Plane Development Kit. User-space NIC drivers and packet
+  I/O primitives that VPP sits on top of.
+- **ERSPAN** — Encapsulated Remote Switched Port Analyzer. A traffic mirror
+  protocol that wraps mirrored frames in GRE so they can be sent across an
+  IP network. **ERSPAN III** is the version with a richer header (timestamps,
+  hardware ID, security group tag).
+- **Flow** — A unidirectional sequence of packets sharing the same key
+  (typically 5-tuple: src IP, dst IP, src port, dst port, protocol; extended
+  with VRF/VNI when present).
+- **Flow table / flow flow-rule** — In-memory data structure that maps a flow
```
banidoru: Typo from the rename: `Flow table / flow flow-rule` — should be `Flow table / flow-rule` (extra `flow` word).

specs/TEMPLATE.md
```diff
@@ -0,0 +1,31 @@
+# NNN — Title
+
+**Owners:** @handle1, @handle2
+**Status:** Draft | Accepted | Superseded
```
banidoru: The main spec (`000-overview.md`) dropped `Owners / Status / Last updated` in favor of a Version History table. Since this template is now the canonical copy, consider updating it to match — otherwise new specs created from the template will use a format the project already moved away from.

banidoru:
This typo is still present in the latest version of the file.

banidoru:
The template is now out of sync with the main spec. This should be addressed.

banidoru (changes requested):
- The typo in  remains.\n- The new  is out of sync with .

specs/000-overview.md
banidoru: Still open — the typo `flow flow-rule` remains on this line in the latest commit (042e504). Should be `flow-rule` (remove the extra `flow`).

specs/TEMPLATE.md
banidoru: Still open — `specs/TEMPLATE.md` still uses the `Owners / Status / Last updated` header format while the main spec has moved to a Version History table. The template should be updated to match.

banidoru (commented):
Re-review (iteration 2) of the new commits (dbaf00d → 042e504):

- **Two prior threads remain unresolved:** (1) typo `flow flow-rule` in Glossary line 281, (2) TEMPLATE.md header format doesn't match the main spec's Version History table. Replied on both threads noting they're still open.
- The new content is solid — VRF/VNI source note, narrowed RTT scope, redrawn diagram, backend language note, UDS security model, E2E execution policy, CI scope split, opt-in packaging, and linear-history policy all address prior review feedback well.
- No new issues found in the diff beyond the two outstanding items above.

r12f:

specs/000-overview.md
r12f: The `flow flow-rule` artifact was fixed in 075a26d as part of the glossary rewrite — the line is now split into separate **Flow-rule** and **Flow table** entries with no leftover `flow flow-rule` token. Resolving.

specs/TEMPLATE.md
r12f: Done in fb82188 — replaced the `Owners / Status / Last updated` header block with the same `Version history` table the main spec uses. New specs cloned from TEMPLATE.md will now match.

specs/000-overview.md
```diff
@@ -0,0 +1,360 @@
...
+   overhead on DPU).
+3. **Flow table eviction policy** — LRU, time-based, or hybrid. Decided in
+   spec 003. *Default:* time-based with active/idle timeouts (IPFIX-aligned).
+4. **Distro target** — Ubuntu 22.04 only, or also DOCA's recommended base?
+   Decided in spec 005. *Default:* Ubuntu 22.04 arm64.
+5. **Frontend API authn/authz** — v1 ships with UNIX-group-based access on a
+   UDS (see Component Map note). Should we add a token or mTLS layer for
+   multi-tenant DPU scenarios? Tracked for a post-v1 spec. *Default:* keep
+   group-based for v1; revisit when a concrete multi-tenant requirement
+   lands.
```
banidoru: Nit: file is missing a trailing newline (`\ No newline at end of file`). Most editors and linters expect one — consider adding it.

specs/000-overview.md
```diff
@@ -0,0 +1,360 @@
+# 000 — System Overview
+
+## Version history
+
+| Version | Date       | Author      | Changes        |
+| ------- | ---------- | ----------- | -------------- |
+| 0.1     | 2026-04-18 | r12f        | Initial draft. |
+| 0.2     | 2026-04-18 | r12f        | Rename `tracker` → `flow-rule`, `bucket` → `flow`. |
+| 0.3     | 2026-04-18 | r12f       | Clarify flow-rule generates one flow per distinct key tuple; fix glossary entry. |
```
banidoru: Nit: column alignment is off by one space compared to the v0.1/v0.2 rows (`r12f       |` vs `r12f        |`). Cosmetic only.

banidoru (commented):
Iteration 3 re-review — all prior threads resolved, new commit looks good.

- All 15 previous review threads have been addressed and resolved. Nice work on the thorough responses.
- Glossary split (Flow / Flow-rule / Flow table) is clean and removes the `flow flow-rule` typo.
- TEMPLATE.md now matches the version-history format used in the main spec — good consistency.
- Minor: `000-overview.md` is missing a trailing newline (the diff shows `\ No newline at end of file`). Not a blocker but worth fixing.
- Minor: version history v0.3 row has slightly off column alignment (`r12f` has one fewer padding space than v0.1/v0.2). Cosmetic only.

specs/000-overview.md
banidoru: Still open — the diff shows `\\ No newline at end of file` is still present on the last line of `specs/000-overview.md` in commit 0a830fb. Please add a trailing newline.

specs/000-overview.md
banidoru: Still open — the v0.3 row still has one fewer space before the pipe than v0.1/v0.2 rows in commit 0a830fb.

banidoru (commented):
- Glossary split (Flow / Flow-rule / Flow table) is clean and removes the prior typo — good.
- TEMPLATE.md header update to match the version-history table format used in 000 — good.
- Two prior nits still open: missing trailing newline (line 360) and column misalignment in the v0.3 row (line 9). Replied on both threads.
- No new issues found in the 042e504→0a830fb diff.

specs/TEMPLATE.md
```diff
@@ -0,0 +1,73 @@
...
+## Metadata (optional)
+
+Include only the fields that apply. All cross-spec references must be
+**markdown links** to the target spec file, not bare names.
+
+- **Parent epic:** `DPU-NN` (Multica issue, if any)
+- **Depends on:** [`NNN-<slug>`](NNN-<slug>.md), [`NNN-<slug>`](NNN-<slug>.md)
+- **Related:** [`NNN-<slug>`](NNN-<slug>.md)
+- **Affects:** `<component>`, `<component>` (e.g. `infmon-backend`, `infmon-cli`)
+- **Repo:** `r12f/InFMon`
```
r12f: Remove this.

specs/TEMPLATE.md
```diff
@@ -0,0 +1,73 @@
...
+
+| Version | Date       | Author | Changes        |
+| ------- | ---------- | ------ | -------------- |
+| 0.1     | YYYY-MM-DD | r12f   | Initial draft. |
+
+> **Version-history rules.**
+> - **One PR = exactly one row.** When you push fixes addressing review
+>   comments on the same PR, **amend the existing row's `Changes` cell** —
+>   do **not** add a new row per review iteration.
+> - The `Author` column is always `r12f` (the spec repo owner). Co-authors,
```
r12f: Author is the GitHub account id and name of PR author, e.g. Riff (r12f)

specs/000-overview.md
banidoru: Still open — `specs/000-overview.md` still has no trailing newline in commit 070b5f1.

specs/000-overview.md
banidoru: Fixed — the version history table was consolidated to a single v0.1 row, eliminating the alignment inconsistency.

specs/TEMPLATE.md
banidoru: Addressed — TEMPLATE.md has been rewritten with the new structure; the old content this referred to is gone.

specs/TEMPLATE.md
banidoru: Addressed — the template now uses `Riff (r12f)` as the example author, and the version-history rules block documents the `Name (handle)` convention.

banidoru (commented):
Iteration 4 re-review:

- **3 previously-open threads resolved:** column alignment fix (version history consolidated), TEMPLATE.md cleanup, and author format convention — all addressed in this push.
- **1 thread still open:** `specs/000-overview.md` is still missing a trailing newline.
- **Version history consolidation looks good** — folding v0.1/v0.2/v0.3 into a single row with expanded Changes cell matches the new "one PR = one row" convention from TEMPLATE.md.
- **TEMPLATE.md rewrite is solid** — version-history rules, forbidden-metadata list, and metadata conventions are clear and actionable.
- **One remaining nit:** please add the trailing newline to `specs/000-overview.md` to close out the last open thread.

banidoru (changes requested):
The file `specs/000-overview.md` is still missing a trailing newline.

specs/000-overview.md
banidoru: Confirmed — `specs/000-overview.md` still lacks a trailing newline in commit 070b5f1. Still open.

banidoru (commented):
- **Trailing newline still missing** on `specs/000-overview.md` — only open issue remaining from prior reviews.
- New diff (0a830fb→070b5f1): version-history consolidation and TEMPLATE.md rewrite look clean.
- Author format (`Riff (r12f)`) consistently applied.
- Forbidden-metadata rules and version-history conventions are well-documented in the template.
- No new issues found in the latest changes.

specs/000-overview.md
r12f: Fixed — added trailing newline.

debian/infmon-frontend.logrotate
```diff
@@ -0,0 +1,13 @@
...
+    daily
+    rotate 14
+    compress
+    delaycompress
+    missingok
+    notifempty
+    create 0640 infmon infmon
+    sharedscripts
+    postrotate
+        systemctl reload infmon-frontend.service >/dev/null 2>&1 || true
```
banidoru: This calls `systemctl reload infmon-frontend.service`, but `debian/infmon-frontend.service` has no `ExecReload=` directive. Without one, systemd sends SIGHUP by default. If the frontend binary doesn't handle SIGHUP for log-file reopening, this will either silently do nothing or terminate the process.

Consider adding `ExecReload=/bin/kill -USR1 $MAINPID` (or whichever signal the frontend will use for log reopen) to the service unit, and matching the signal here.

tools/sync-debian-changelog.py
```diff
@@ -0,0 +1,249 @@
...
+from typing import Dict, List, Tuple
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+CHANGELOG_MD = REPO_ROOT / "CHANGELOG.md"
+DEBIAN_CHANGELOG = REPO_ROOT / "debian" / "changelog"
+
+PKG = "infmon"
+MAINTAINER = os.environ.get(
+    "DEBFULLNAME_EMAIL",
+    "InFMon maintainers <r12f.code@gmail.com>",
```
banidoru: `DEBFULLNAME_EMAIL` is non-standard. Debian packaging tools use two separate env vars: `DEBFULLNAME` and `DEBEMAIL`. Consider:
```python
MAINTAINER = (
    f"{os.environ['DEBFULLNAME']} <{os.environ['DEBEMAIL']}>"
    if 'DEBFULLNAME' in os.environ and 'DEBEMAIL' in os.environ
    else "InFMon maintainers <r12f.code@gmail.com>"
)
```
This way `dpkg-buildpackage`'s standard env setup works out of the box.

debian/control
```diff
@@ -0,0 +1,79 @@
...
+Description: InFMon flow telemetry — command-line client
+ InFMon is a high-performance network flow monitoring system built around
+ a VPP-based dataplane and a Rust control plane.
+ .
+ This package contains the infmonctl command-line client. It talks to
+ the infmon-frontend over its REST API and does not require VPP to be
+ installed locally.
+
+Package: infmon
+Architecture: arm64
```
banidoru: Meta packages that only pull in dependencies are conventionally `Architecture: all` rather than a specific arch. Since the `infmon` meta package has no binaries of its own (only `${misc:Depends}` + the three sub-packages), `all` avoids needing a cross-build to produce it and is the Debian convention for this pattern.

tools/sync-debian-changelog.py
```diff
@@ -0,0 +1,249 @@
...
+    cur_subsection: str = "Changed"
+    cur_bullets: List[Tuple[str, str]] = []
+
+    def flush():
+        if cur_ver is not None:
+            sections.append((cur_ver, cur_date, list(cur_bullets)))
+
+    for line in md.splitlines():
+        m = SECTION_RE.match(line)
+        if m:
```
banidoru: `BULLET_RE` only matches single-line bullets. Multi-line bullets in CHANGELOG.md (like the ones currently in the file — e.g. the first entry spans 3 lines) will have their continuation lines silently dropped, producing truncated debian/changelog entries. Consider accumulating continuation lines (non-blank, non-header lines following a bullet) into the current bullet's text.

banidoru (commented):
Re-review (iteration 5) of the packaging PR:

- All 18 prior review threads are resolved — nice cleanup.
- Logrotate `postrotate` calls `systemctl reload` but the systemd unit has no `ExecReload=` — this will send SIGHUP by default, which may crash the frontend if unhandled.
- `DEBFULLNAME_EMAIL` is a non-standard env var; Debian tooling uses `DEBFULLNAME` + `DEBEMAIL` separately.
- Meta package `infmon` uses `Architecture: arm64` — `all` is conventional for dependency-only packages.
- `sync-debian-changelog.py` drops continuation lines of multi-line CHANGELOG.md bullets (only first line captured).

banidoru (changes requested):
- The four outstanding review threads from the previous review cycle remain unaddressed in the latest commits.
- `logrotate` configuration attempts to reload a service that has no `ExecReload` directive, which will not work as intended.
- A script is using non-standard environment variables for Debian packaging, which will break standard build workflows.
- The `infmon` meta-package has an incorrect architecture specified (`amd64` instead of `all`).
- A script for syncing changelogs has a bug that truncates multi-line entries.

These issues need to be addressed before the PR can be approved.

banidoru (approved):
Re-review after rebase (iteration 5):

- All 17 prior review threads on the spec files are resolved — feedback was addressed thoroughly.
- The only delta from the previous review is the trailing-newline fix on `specs/000-overview.md`, which resolves the last outstanding nit.
- 4 orphaned threads reference `debian/` files that are no longer part of this PR (moved to a separate branch). These are not actionable here.
- Spec content is solid: clear scope, good component/data-flow diagrams, well-defined glossary, and the TEMPLATE.md is clean.
- No new issues found. LGTM.

debian/infmon-frontend.logrotate
r12f: The packaging files have been removed from this branch — they now belong to PR #13. This comment is no longer applicable.

tools/sync-debian-changelog.py
r12f: The packaging files have been removed from this branch — they now belong to PR #13. This comment is no longer applicable.

debian/control
r12f: The packaging files have been removed from this branch — they now belong to PR #13. This comment is no longer applicable.

tools/sync-debian-changelog.py
r12f: The packaging files have been removed from this branch — they now belong to PR #13. This comment is no longer applicable.

banidoru (approved):
- All 22 prior review threads are resolved — fixes verified across the board.
- `specs/000-overview.md` is comprehensive and well-structured: scope, component map, data flow, repo layout, build/release model, glossary, and open questions all read cleanly.
- `specs/TEMPLATE.md` now matches the main spec's version-history format with clear rules for authors and forbidden metadata.
- Security model (UDS + group perms), E2E execution strategy, and backend C/C++ language boundary are all explicitly documented — addresses earlier ambiguities well.
- No new issues found. LGTM.

banidoru (approved):
- All prior review threads resolved; fixes verified in latest commits.
- Spec 000 is comprehensive: scope, component map, data flow, repo layout, build/release, glossary, and open questions are all well-structured.
- VRF/VNI extraction source, backend C/C++ split, UDS security model, and E2E execution cadence are clearly documented.
- TEMPLATE.md is clean and consistent with the conventions established in 000.
- No new issues found in the diff from 9f8a573 to aa0f6b4.

banidoru (approved):
All reviewers approved. LGTM.

PR #5, opened 2026-04-18
specs: 002 — flow tracking model

r12f:
## Summary

Adds `specs/002-flow-tracking-model.md`, the spec for the flow-tracker
abstraction that sits between the packet parser (Spec 003) and the OTLP
exporter (Spec 006).

Highlights:

- **Tracker model** — a tracker is `(name, ordered field list, max_keys, eviction_policy)`,
  immutable for its lifetime. Keys are the tuple of field values in declared order;
  buckets are per-(tracker, key).
- **v1 field set (L3, inner only)** — `src_ip`, `dst_ip` (both stored as 16 B with
  IPv4 mapped to v6 for layout uniformity), `ip_proto`, `dscp`. Fixed-width
  canonical key encoding capped at 64 B per tracker so v2 L4 fields fit without
  revisiting the cap.
- **Config schema** — static file at backend start, TOML canonical with YAML accepted;
  all-or-nothing validation on load/reload, full rule list documented (§5.3).
- **CRUD API surface** consumed by `infmon-cli flow {add,rm,list,show}` with
  synchronous semantics, atomic swap against the data plane, and a small closed
  error set. `add` does not replace by name on purpose — `rm` then `add` is the only
  mutation path, which preserves the immutability promise the exporter's label set
  depends on.
- **Eviction (v1)** — a single policy, `lru_drop`: drop the LRU bucket, insert
  the new key, bump `infmon_tracker_evictions_total`. Drop-on-failure surfaces
  via a separate `_drops_total{reason}` counter. Rationale for drop-not-flush
  is in §6.1.
- **Observability** — five per-tracker metrics exported via Spec 006, enough for
  an operator to size `max_keys` against observed cardinality.
- **Future extension hooks** (non-normative) for L4 fields, RoCEv2 BTH/AETH
  fields, additional eviction policies, and per-tracker sampling. Dynamic field
  add/reorder is explicitly out of scope and the rationale is recorded.

## Verification

Spec-only PR; no code/tests touched. Reviewed for internal consistency with the
cross-referenced specs (000 system overview, 001 CI, 003 parser, 004 backend,
006 OTLP, 007 CLI).

Closes DPU-8.

### DIFF
--- a/specs/002-flow-tracking-model.md
+++ b/specs/002-flow-tracking-model.md
@@ -288,7 +288,7 @@ spec defines only the operations and their semantics.
   fields, the total `max_keys` budget in rule 6 — are reported up
   front), then issues `add` calls one at a time. Each `add` also
   re-checks the running budget, so the same `budget_exceeded` error
-  (§7.2) can be produced by a CLI `flow add` that would push the
+  (§7.2) can be produced by a CLI `flow-rule add` that would push the
   cumulative `max_keys` past rule 6's limit. This means file-order does
   **not** affect whether a valid file loads, but it does affect the
   order of error reporting if the validator is bypassed (e.g. a future
@@ -304,7 +304,7 @@ A small, closed set:
   reason.
 - `budget_exceeded` — would exceed the backend's total `max_keys`
   budget (§5.3 rule 6). Returned both at config-file load and from a
-  CLI `flow add` whose `max_keys` would push the cumulative total past
+  CLI `flow-rule add` whose `max_keys` would push the cumulative total past
   the budget; runtime `add` therefore enforces the same invariant as
   the file validator and operators cannot silently overcommit between
   reloads.
--- a/specs/002-flow-tracking-model.md
+++ b/specs/002-flow-tracking-model.md
@@ -2,21 +2,13 @@
 
 ## Version history
 
-| Version | Date       | Author      | Changes        |
-| ------- | ---------- | ----------- | -------------- |
-| 0.1     | 2026-04-18 | r12f        | Initial draft. |
-| 0.2     | 2026-04-18 | r12f        | Rename tracker->flow-rule, bucket->flow. |
-| 0.3     | 2026-04-18 | r12f        | Add `mirror_src_ip` to v1 field set as the only outer-header field allowed in a flow-rule key. |
-| 0.4     | 2026-04-18 | r12f        | Address PR #5 review: tighten `name` regex to ≥ 2 chars and motivate 31-char cap; `max_keys` positive (not non-negative); spell out DSCP extraction as `(tos>>2)&0x3F`; promote 64-byte cap to MUST and cross-reference §5.3; document LRU recency on insert and tie-breaking; clarify `show` returns `last_seen_ns` (max across resident flows); spell out file-load vs per-`add` budget enforcement; remove `budget_exceeded_runtime` drop reason for v1 (single-threaded CRUD); note TOML/YAML singular/plural is syntactic only. |
-| 0.5     | 2026-04-18 | r12f       | Add §2.4 mental-model section (one flow-rule generates one flow per distinct key tuple); align CLI examples with the `flow-rule` / `flow` verb split (Spec 007); rename `infmon_tracker_*` metrics to `infmon_flow_rule_*` (and `_buckets`→`_flows`, `bucket_count`→`flow_count`); promote `mirror_src_ip` to part of the recommended default flow-rule (still opt-in for custom flow-rules). |
-| 0.6     | 2026-04-18 | r12f       | Add `mirror_src_ip` canonical encoding to §4; fix `token-flow` rename artifact in §9.4 → `token-bucket`. |
+| Version | Date       | Author       | Changes |
+| ------- | ---------- | ------------ | ------- |
+| 0.1     | 2026-04-18 | Riff (r12f)  | Initial draft of the flow-rule data model and lifecycle. Establishes flow-rule (matcher) vs flow (one per distinct key tuple), `mirror_src_ip` as the only outer-header field allowed in a key, name regex / `max_keys` / 64-byte caps, LRU recency on insert, file-load vs per-`add` budget enforcement, drop-reason set, `infmon_flow_rule_*` metric names, and §2.4 mental-model section. CLI examples align with the `flow-rule` / `flow` verb split (Spec 007). |
 
-Parent epic: DPU-4 (EPIC: InFMon — flow telemetry service on BF-3)
-Depends on: 000-overview (system overview), 003-erspan-and-packet-parsing
-           (defines the parsed inner-packet record this spec keys off of)
-Related: 004-backend-architecture (consumes this model in the VPP plugin),
-         007-cli (exposes the flow-rule CRUD as `flow-rule {add,rm,list,show}`
-         and live flow inspection as `flow {list,show}`)
+- **Parent epic:** `DPU-4` (EPIC: InFMon — flow telemetry service on BF-3)
+- **Depends on:** [`000-overview`](000-overview.md), [`003-erspan-and-packet-parsing`](003-erspan-and-packet-parsing.md)
+- **Related:** [`004-backend-architecture`](004-backend-architecture.md), [`007-cli`](007-cli.md)
 
 ## 1. Purpose
 
--- a/specs/002-flow-tracking-model.md
+++ b/specs/002-flow-tracking-model.md
@@ -5,6 +5,7 @@
 | Version | Date       | Author      | Changes        |
 | ------- | ---------- | ----------- | -------------- |
 | 0.1     | 2026-04-18 | bf3 (agent) | Initial draft. |
+| 0.2     | 2026-04-18 | bf3 (agent) | Rename `tracker` → `flow-rule`, `bucket` → `flow`. |
 
 Tracking issue: DPU-8 (project InFMon)
 Parent epic: DPU-4 (EPIC: InFMon — flow telemetry service on BF-3)
@@ -15,24 +16,24 @@ Related: 004-backend-architecture (consumes this model in the VPP plugin),
 
 ## 1. Purpose
 
-Define the data model and lifecycle of a **flow tracker** in InFMon: the
+Define the data model and lifecycle of a **flow flow-rule** in InFMon: the
 named, configurable thing that turns parsed inner-packet records into
-counter buckets. The flow tracker is the single seam between
+counter flows. The flow flow-rule is the single seam between
 *"what the wire just gave us"* (Spec 003) and
 *"what we export"* (Spec 006, OTLP).
 
-A v1 deployment will typically run with a handful of trackers (≤ 16) and
-millions of keys per tracker; this spec sets the contract that lets us
+A v1 deployment will typically run with a handful of flow-rules (≤ 16) and
+millions of keys per flow-rule; this spec sets the contract that lets us
 keep that bounded and predictable.
 
 ## 2. Concepts
 
-### 2.1 Tracker
+### 2.1 Flow-rule
 
-A **tracker** is:
+A **flow-rule** is:
 
 ```
-tracker := (name, ordered_field_list, max_keys, eviction_policy)
+flow-rule := (name, ordered_field_list, max_keys, eviction_policy)
 ```
 
 - **name** — short identifier, `^[a-z0-9][a-z0-9_-]{0,30}$`. Used in
@@ -40,32 +41,32 @@ tracker := (name, ordered_field_list, max_keys, eviction_policy)
 - **ordered_field_list** — non-empty, ordered tuple of fields drawn from
   the v1 field set (§3). Order is significant: it defines the byte
   layout of the key (§4) and therefore the hash. Reordering produces a
-  different tracker even with the same field set.
+  different flow-rule even with the same field set.
 - **max_keys** — non-negative integer; upper bound on the number of
-  distinct keys held simultaneously for this tracker. Required.
+  distinct keys held simultaneously for this flow-rule. Required.
 - **eviction_policy** — what to do when a new key arrives at a full
-  tracker. v1 supports a single policy (§6); the field exists so future
+  flow-rule. v1 supports a single policy (§6); the field exists so future
   policies can be added without a config break.
 
-A tracker is *static* in v1: its field list and max_keys are fixed at
-configuration time. CRUD operations replace the whole tracker; partial
+A flow-rule is *static* in v1: its field list and max_keys are fixed at
+configuration time. CRUD operations replace the whole flow-rule; partial
 mutation (e.g. add a field) is explicitly out of scope to keep the key
-layout immutable for the lifetime of any one tracker.
+layout immutable for the lifetime of any one flow-rule.
 
 ### 2.2 Key
 
-A **key** is the concrete value emitted by a packet for a given tracker:
-the tuple of field values, in the tracker's declared order. Two packets
+A **key** is the concrete value emitted by a packet for a given flow-rule:
+the tuple of field values, in the flow-rule's declared order. Two packets
 produce the same key iff every listed field is bytewise equal after the
 canonicalisation rules in §4.
 
-### 2.3 Bucket
+### 2.3 Flow
 
-A **bucket** is the per-key counter state owned by the tracker. v1
-buckets carry the minimum needed by the OTLP exporter (Spec 006):
-`packets`, `bytes`, `first_seen_ns`, `last_seen_ns`. The exact bucket
+A **flow** is the per-key counter state owned by the flow-rule. v1
+flows carry the minimum needed by the OTLP exporter (Spec 006):
+`packets`, `bytes`, `first_seen_ns`, `last_seen_ns`. The exact flow
 layout is owned by Spec 004 (backend); this spec only requires that
-*there is one bucket per (tracker, key)*.
+*there is one flow per (flow-rule, key)*.
 
 ## 3. Field set v1 (L3, inner)
 
@@ -80,12 +81,12 @@ intentionally not part of v1.
 | `ip_proto` | 1 B                     | inner IPv4.protocol / IPv6.next_header (after extension headers) | 0–255         |
 | `dscp`     | 1 B (low 6 bits used)   | inner IPv4.tos>>2 / IPv6.tc>>2 | upper 2 bits zero                                     |
 
-A tracker MUST list at least one field. There is no implicit field; if
-you want a tracker keyed only on `dscp`, configure it that way.
+A flow-rule MUST list at least one field. There is no implicit field; if
+you want a flow-rule keyed only on `dscp`, configure it that way.
 
 ### 3.1 Why IPv4-mapped-IPv6
 
-So a single tracker can carry both address families with one fixed-width
+So a single flow-rule can carry both address families with one fixed-width
 key. A v4 packet and a v6 packet that happen to encode the same address
 *will* collide; this is by design — operators who want them separated
 should add `ip_proto` (or, post-v1, an explicit `ip_version` field).
@@ -94,13 +95,13 @@ should add `ip_proto` (or, post-v1, an explicit `ip_version` field).
 
 The packet parser (Spec 003) is responsible for delivering a normalised
 record with each v1 field already extracted, validated, and in
-host-endian where applicable. Trackers do **not** re-parse packets.
+host-endian where applicable. Flow-rules do **not** re-parse packets.
 Malformed / truncated packets are dropped upstream and never reach the
-tracker (they bump a parser counter, not a tracker counter).
+flow-rule (they bump a parser counter, not a flow-rule counter).
 
 ## 4. Key layout & canonicalisation
 
-For a tracker `T = [f1, f2, ..., fn]` the key is the concatenation of
+For a flow-rule `T = [f1, f2, ..., fn]` the key is the concatenation of
 each field's canonical encoding, in order, with no padding between
 fields:
 
@@ -114,16 +115,16 @@ Canonical encodings (v1):
 - `ip_proto`: 1 byte.
 - `dscp`: 1 byte, value in `0..=63`, upper bits zeroed.
 
-Total key width is fixed per tracker and computable from the field list
+Total key width is fixed per flow-rule and computable from the field list
 alone. Implementations SHOULD reject configurations whose key width
 exceeds 64 bytes (room for L4 fields in v2 without revisiting this cap).
 
-The hash function used to index the bucket store is owned by Spec 004;
+The hash function used to index the flow store is owned by Spec 004;
 this spec only fixes the *input* (the bytes above, in this order).
 
 ## 5. Configuration schema
 
-Trackers are configured via a static file loaded at backend start.
+Flow-rules are configured via a static file loaded at backend start.
 TOML is the canonical format; YAML is accepted but converted to the same
 internal representation. The CLI (Spec 007) is a thin wrapper that
 mutates this same schema and asks the backend to reload.
@@ -133,13 +134,13 @@ mutates this same schema and asks the backend to reload.
 ```toml
 # /etc/infmon/flows.toml
 
-[[tracker]]
+[[flow-rule]]
 name             = "by_5tuple_l3"
 fields           = ["src_ip", "dst_ip", "ip_proto", "dscp"]
 max_keys         = 1_048_576           # 2^20
 eviction_policy  = "lru_drop"          # only value supported in v1
 
-[[tracker]]
+[[flow-rule]]
 name             = "by_dscp"
 fields           = ["dscp"]
 max_keys         = 64
@@ -149,7 +150,7 @@ eviction_policy  = "lru_drop"
 ### 5.2 YAML (equivalent)
 
 ```yaml
-trackers:
+flow-rules:
   - name: by_5tuple_l3
     fields: [src_ip, dst_ip, ip_proto, dscp]
     max_keys: 1048576
@@ -165,38 +166,38 @@ trackers:
 The backend MUST reject the configuration (and refuse to start, or
 refuse the reload) if any of the following hold:
 
-1. Two trackers share a name.
-2. A tracker's `fields` list is empty, contains an unknown field, or
+1. Two flow-rules share a name.
+2. A flow-rule's `fields` list is empty, contains an unknown field, or
    contains duplicates.
 3. `max_keys` is missing, negative, or zero.
 4. `eviction_policy` is not in the supported set (v1: `{"lru_drop"}`).
 5. The computed key width exceeds 64 bytes.
 6. The total of all `max_keys` exceeds the backend's compile-time
-   budget (set by Spec 004; default 16 Mi keys across all trackers).
+   budget (set by Spec 004; default 16 Mi keys across all flow-rules).
 
 Validation is **all-or-nothing** per reload. A failed reload leaves the
-previously running tracker set untouched.
+previously running flow-rule set untouched.
 
 ## 6. Eviction policy (v1)
 
 v1 ships exactly one policy: **`lru_drop`**.
 
-Behaviour, when a packet would create a new key in a tracker that is
+Behaviour, when a packet would create a new key in a flow-rule that is
 already at `max_keys`:
 
-1. Evict the least-recently-updated key from the tracker. Its bucket
+1. Evict the least-recently-updated key from the flow-rule. Its flow
    contents are **dropped**, not flushed — Spec 006 (OTLP) flushes
    continuously, so the loss is bounded by the export interval.
-2. Insert the new key with a fresh bucket and apply the current packet.
-3. Increment the per-tracker counter `infmon_tracker_evictions_total`
-   (label: `tracker=<name>`).
+2. Insert the new key with a fresh flow and apply the current packet.
+3. Increment the per-flow-rule counter `infmon_tracker_evictions_total`
+   (label: `flow-rule=<name>`).
 
 If the eviction itself fails (e.g. data structure invariant violation),
 the packet is dropped and `infmon_tracker_drops_total` is incremented
-instead — the tracker never silently corrupts.
+instead — the flow-rule never silently corrupts.
 
 "Recently used" means *most recently updated by an incoming packet*.
-Recency state lives with the bucket and is maintained on every hit; the
+Recency state lives with the flow and is maintained on every hit; the
 exact data structure is owned by Spec 004 (a likely choice is a
 segmented LRU keyed off the existing hash table, but that's an
 implementation detail).
@@ -217,10 +218,10 @@ spec defines only the operations and their semantics.
 
 | Op       | CLI                              | Input                                     | Output                              | Notes |
 |----------|----------------------------------|-------------------------------------------|-------------------------------------|-------|
-| `add`    | `infmon-cli flow add <spec>`     | full tracker definition (name, fields, max_keys, eviction_policy) | created tracker, or error           | Fails if name exists. |
-| `rm`     | `infmon-cli flow rm <name>`      | tracker name                              | ok / `not_found`                    | Drops all buckets for that tracker. |
-| `list`   | `infmon-cli flow list`           | —                                         | array of tracker definitions        | Cheap; no bucket data. |
-| `show`   | `infmon-cli flow show <name>`    | tracker name                              | tracker definition + live stats: `bucket_count`, `evictions_total`, `drops_total`, `last_packet_ns` | Stats are best-effort snapshots. |
+| `add`    | `infmon-cli flow add <spec>`     | full flow-rule definition (name, fields, max_keys, eviction_policy) | created flow-rule, or error           | Fails if name exists. |
+| `rm`     | `infmon-cli flow rm <name>`      | flow-rule name                              | ok / `not_found`                    | Drops all flows for that flow-rule. |
+| `list`   | `infmon-cli flow list`           | —                                         | array of flow-rule definitions        | Cheap; no flow data. |
+| `show`   | `infmon-cli flow show <name>`    | flow-rule name                              | flow-rule definition + live stats: `bucket_count`, `evictions_total`, `drops_total`, `last_packet_ns` | Stats are best-effort snapshots. |
 
 ### 7.1 Semantics
 
@@ -228,13 +229,13 @@ spec defines only the operations and their semantics.
   call returns only after the backend has applied (or rejected) the
   change.
 - `add` and `rm` are atomic with respect to the data plane: an
-  in-flight packet either sees the old tracker set or the new one,
+  in-flight packet either sees the old flow-rule set or the new one,
   never a partial state. Implementation note: Spec 004 will likely
-  use RCU-style swap of the tracker table.
-- `add` does **not** support replace-by-name. To change a tracker,
-  `rm` then `add`. This preserves the rule from §2.1 that a tracker's
+  use RCU-style swap of the flow-rule table.
+- `add` does **not** support replace-by-name. To change a flow-rule,
+  `rm` then `add`. This preserves the rule from §2.1 that a flow-rule's
   field list is immutable for its lifetime, and forces operators to
-  acknowledge that buckets will be dropped.
+  acknowledge that flows will be dropped.
 - `list` and `show` are read-only and lock-free on the data plane.
 - The static config file (§5) is applied via the same `add`/`rm`
   primitives at startup; there is no separate "bulk load" path.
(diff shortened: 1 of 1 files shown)


specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+- **name** — short identifier, `^[a-z0-9][a-z0-9_-]{0,30}$`. Used in
+  exported metric labels and CLI references. Unique per backend instance.
+- **ordered_field_list** — non-empty, ordered tuple of fields drawn from
+  the v1 field set (§3). Order is significant: it defines the byte
+  layout of the key (§4) and therefore the hash. Reordering produces a
+  different tracker even with the same field set.
+- **max_keys** — non-negative integer; upper bound on the number of
+  distinct keys held simultaneously for this tracker. Required.
```
banidoru: "non-negative integer" includes zero, but §5.3 rule 3 explicitly rejects zero (`missing, negative, or zero`). This is a contradiction — should be "positive integer" here.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,314 @@
...
+TOML is the canonical format; YAML is accepted but converted to the same
+internal representation. The CLI (Spec 007) is a thin wrapper that
+mutates this same schema and asks the backend to reload.
+
+### 5.1 TOML
+
+```toml
+# /etc/infmon/flows.toml
+
+[[tracker]]
```
banidoru: TOML uses `[[tracker]]` (singular) but the YAML equivalent uses `trackers:` (plural). These should match to avoid confusion when implementing the config parser — pick one naming convention.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,314 @@
...
+- The static config file (§5) is applied via the same `add`/`rm`
+  primitives at startup; there is no separate "bulk load" path.
+
+### 7.2 Errors
+
+A small, closed set:
+
+- `name_exists` — `add` with a duplicate name.
+- `not_found` — `rm` / `show` against an unknown name.
+- `invalid_spec` — any §5.3 validation failure; carries a human-readable
```
banidoru, line 303: `budget_exceeded_runtime` appears as a drop reason here but isn't explained anywhere in the spec. §5.3 rule 6 and §7.2's `budget_exceeded` error cover config-time validation — when exactly would a runtime budget exceedance occur that bypasses those checks? If it's for races during concurrent `add` calls, that should be documented. If it can't actually happen in v1 (single-threaded CRUD), consider removing it.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,314 @@
...
+
+If the eviction itself fails (e.g. data structure invariant violation),
+the packet is dropped and `infmon_tracker_drops_total` is incremented
+instead — the tracker never silently corrupts.
+
+"Recently used" means *most recently updated by an incoming packet*.
+Recency state lives with the bucket and is maintained on every hit; the
+exact data structure is owned by Spec 004 (a likely choice is a
+segmented LRU keyed off the existing hash table, but that's an
+implementation detail).
```
banidoru, line 247: §7.1 states the static config file is applied via the same `add`/`rm` primitives at startup with no separate bulk path. But §5.3 rule 6 validates total `max_keys` budget. If trackers are added sequentially, does each `add` check the running total? If so, the order of `[[tracker]]` entries in the file matters — a large tracker added first could block subsequent ones even if the whole set fits. Consider clarifying: is the file validated as a whole first (§5.3), then applied via `add`? Or is each `add` independently validated?

banidoru (commented):
Solid spec overall — well-structured, good cross-references, and the rationale sections are appreciated. A few items to tighten up:

- **`max_keys` definition contradicts validation rule** — §2.1 says "non-negative" but §5.3 rejects zero. Pick one.
- **`budget_exceeded_runtime` drop reason is unexplained** — §8 introduces it but no section defines when runtime budget exceedance happens vs. the config-time check in §5.3/§7.2.
- **TOML/YAML key inconsistency** — TOML uses `[[tracker]]` (singular), YAML uses `trackers:` (plural). Will cause confusion for config parsers.
- **Startup sequencing gap** — §7.1 says config file is applied via sequential `add` calls, but §5.3 rule 6 validates total budget. Unclear if validation is per-op or whole-file.
- **`show` stats missing `eviction_policy`** — §7 table shows `show` returns definition + stats, but the definition fields listed don't include the eviction policy in the output.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,314 @@
...
+### 3.1 Why IPv4-mapped-IPv6
+
+So a single tracker can carry both address families with one fixed-width
+key. A v4 packet and a v6 packet that happen to encode the same address
+*will* collide; this is by design — operators who want them separated
+should add `ip_proto` (or, post-v1, an explicit `ip_version` field).
+
+### 3.2 Parser handoff
+
+The packet parser (Spec 003) is responsible for delivering a normalised
```
banidoru, line 122: Nit: the 64-byte cap is described as a SHOULD ("implementations SHOULD reject") but §5.3 rule 5 makes it a MUST ("The backend MUST reject"). These should be consistent — I'd go with MUST everywhere since §5.3 is the normative validation list.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,314 @@
...
+### 2.1 Tracker
+
+A **tracker** is:
+
+```
+tracker := (name, ordered_field_list, max_keys, eviction_policy)
+```
+
+- **name** — short identifier, `^[a-z0-9][a-z0-9_-]{0,30}$`. Used in
+  exported metric labels and CLI references. Unique per backend instance.
```
banidoru: The regex `^[a-z0-9][a-z0-9_-]{0,30}$` allows a single-character name. Is that intentional? A 1-char tracker name in metric labels (`infmon_tracker_buckets{tracker="x"}`) and CLI output could be error-prone. Consider a minimum length of 2–3.

Also, the 31-char cap isn't motivated — is it aligned with a downstream constraint (OTLP attribute length, VPP name table, etc.)? If so, worth a brief note; if not, documenting "picked arbitrarily" is fine too.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,314 @@
...
+```yaml
+trackers:
+  - name: by_5tuple_l3
+    fields: [src_ip, dst_ip, ip_proto, dscp]
+    max_keys: 1048576
+    eviction_policy: lru_drop
+  - name: by_dscp
+    fields: [dscp]
+    max_keys: 64
+    eviction_policy: lru_drop
```
banidoru, line 198: §5.3 rule 6 validates the total `max_keys` budget at config load, but §7 doesn't mention budget validation on `add`. If a sequence of CLI `flow add` calls can push the sum-of-max_keys past the compile-time budget without rejection, operators won't discover the overcommit until the next restart/reload fails.

Should §7.2's `budget_exceeded` error be explicitly linked to rule 6 here, confirming that `add` enforces this check at runtime too?

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,314 @@
...
+refuse the reload) if any of the following hold:
+
+1. Two trackers share a name.
+2. A tracker's `fields` list is empty, contains an unknown field, or
+   contains duplicates.
+3. `max_keys` is missing, negative, or zero.
+4. `eviction_policy` is not in the supported set (v1: `{"lru_drop"}`).
+5. The computed key width exceeds 64 bytes.
+6. The total of all `max_keys` exceeds the backend's compile-time
+   budget (set by Spec 004; default 16 Mi keys across all trackers).
```
banidoru: "Evict the least-recently-updated key" — what's the initial recency value for a key created by the first packet hitting a fresh tracker? If it's "now" (insertion time), that's fine, but it's worth stating explicitly.

More interesting edge case: if a tracker is created via `add` with `max_keys = N` and then receives a burst of N+1 distinct keys in the same batch/poll cycle, does the eviction order among keys that all have the same `last_seen_ns` become arbitrary? If so, that's probably fine but should be called out as implementation-defined.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,314 @@
...
+  call returns only after the backend has applied (or rejected) the
+  change.
+- `add` and `rm` are atomic with respect to the data plane: an
+  in-flight packet either sees the old tracker set or the new one,
+  never a partial state. Implementation note: Spec 004 will likely
+  use RCU-style swap of the tracker table.
+- `add` does **not** support replace-by-name. To change a tracker,
+  `rm` then `add`. This preserves the rule from §2.1 that a tracker's
+  field list is immutable for its lifetime, and forces operators to
+  acknowledge that buckets will be dropped.
```
banidoru: `show` returns `last_packet_ns` but the bucket fields in §2.3 are `first_seen_ns` and `last_seen_ns`. Is `last_packet_ns` the `last_seen_ns` of the most-recently-updated bucket, or a separate tracker-level timestamp?

If it's a tracker-level field, it should be defined somewhere (it's not in §2.1 or §2.3). If it's derived from bucket data, the derivation should be stated. The naming mismatch will confuse implementors.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,314 @@
...
+All v1 fields are extracted from the **inner** packet headers — i.e.
+after ERSPAN decapsulation per Spec 003. Outer/transport fields are
+intentionally not part of v1.
+
+| Field      | Type / width            | Source                       | Notes                                                   |
+|------------|-------------------------|------------------------------|---------------------------------------------------------|
+| `src_ip`   | 16 B (v4 mapped to v6)  | inner IPv4 / IPv6 SA         | IPv4 stored as `::ffff:a.b.c.d` for layout uniformity   |
+| `dst_ip`   | 16 B (v4 mapped to v6)  | inner IPv4 / IPv6 DA         | same mapping rule                                       |
+| `ip_proto` | 1 B                     | inner IPv4.protocol / IPv6.next_header (after extension headers) | 0–255         |
+| `dscp`     | 1 B (low 6 bits used)   | inner IPv4.tos>>2 / IPv6.tc>>2 | upper 2 bits zero                                     |
```
banidoru: Nit: `IPv6.tc>>2` is correct for extracting DSCP, but the description doesn't mention masking. The TOS/TC byte is `(DSCP << 2) | ECN`, so the extraction is `(tc >> 2) & 0x3F` (or equivalently, since it's 6 bits, the `& 0x3F` is redundant after a u8 shift — but only if the input is guaranteed u8). The "upper 2 bits zero" note in the table covers the output contract but not the extraction procedure. Consider: `dscp = (tos >> 2) & 0x3F` to be unambiguous.

banidoru (commented):
- Name regex allows 1-char names and has no stated rationale for the 31-char cap — worth aligning with downstream label constraints or at least documenting why.
- Budget validation (§5.3 rule 6) isn't clearly enforced at `add`-time via the CRUD API — could allow CLI-driven overcommit.
- LRU eviction semantics undefined for keys that have never received a packet (initial recency = ?). Edge case matters when trackers start cold.
- `last_packet_ns` in `show` output (§7) doesn't match any bucket field name from §2.3 (`first_seen_ns`, `last_seen_ns`) — naming inconsistency.
- DSCP extraction description could be more precise about mask-vs-shift ordering for the ECN bits.

banidoru (approved):
All reviewers approved. LGTM.

r12f:

specs/002-flow-tracking-model.md
r12f: Clarified in a8f351c: §7.1 now spells out that the file loader runs the §5.3 rules over the whole file first (so order-independent failures including the rule-6 total budget are reported up front), then issues per-rule `add` calls; each `add` also re-checks the running budget so file-order does not affect whether a valid file loads.

specs/002-flow-tracking-model.md
r12f: Agreed — in v1 the CRUD plane is single-threaded so cumulative-budget overcommit can only happen at `add` time (rejected synchronously via `budget_exceeded`). Removed `budget_exceeded_runtime` from §8 in a8f351c and noted it is reserved for a future revision with concurrent admin paths.

specs/002-flow-tracking-model.md
r12f: Agreed — promoted to MUST in §4 in a8f351c, with an explicit cross-reference to §5.3 rule 5 as the normative restatement.

specs/002-flow-tracking-model.md
r12f: Good catch — fixed in a8f351c. §2.1 now reads 'positive integer (≥ 1)' to match §5.3 rule 3.

specs/002-flow-tracking-model.md
r12f: Acknowledged. The singular/plural difference is intentional (idiomatic TOML `[[flow-rule]]` array-of-tables vs idiomatic YAML `flow-rules:` list); the parser maps both to the same internal collection. Added an explicit note at the top of §5 in a8f351c so implementors don't read it as two schemas.

specs/002-flow-tracking-model.md
r12f: Made unambiguous in a8f351c. §3's table now states the source is `inner IPv4.tos` / `IPv6.tc`, with extraction `(tos >> 2) & 0x3F` (resp. `(tc >> 2) & 0x3F`), making the mask explicit even though it is redundant for an unsigned 8-bit shift.

specs/002-flow-tracking-model.md
r12f: Linked explicitly in a8f351c. §7.2's `budget_exceeded` now references §5.3 rule 6 and states that a CLI `flow add` whose `max_keys` would push the cumulative total past the budget is rejected at runtime as well, so operators cannot silently overcommit between reloads. Matching wording added in §7.1.

specs/002-flow-tracking-model.md
r12f: Naming mismatch fixed in a8f351c. `show` now returns `last_seen_ns` (defined as the maximum `last_seen_ns` across this flow-rule's resident flows; absent if there are none), so it is derived from §2.3 flow fields rather than introducing a new tracker-level timestamp.

specs/002-flow-tracking-model.md
r12f: Both edge cases documented in §6 in a8f351c: recency for a freshly-inserted key is set to the inserting packet's timestamp; if multiple resident keys share the same `last_seen_ns` the eviction choice is implementation-defined (v1 imposes no tie-breaking ordering).

specs/002-flow-tracking-model.md
r12f: Both points addressed in a8f351c. Tightened the regex to `^[a-z0-9][a-z0-9_-]{1,30}$` (2–31 chars) so single-character names are rejected, and added a sentence motivating the 31-char cap as a conservative ceiling under common downstream attribute-length limits (OTLP attributes, VPP node-name tables) — to be tightened if a concrete downstream limit demands it.

banidoru: Still open: §2.1 regex remains `{0,30}` (allows single-char names), not `{1,30}` as described in the reply. The 31-char cap motivation note is also absent from the current text.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,367 @@
...
+
+Tracking issue: DPU-8 (project InFMon)
+Parent epic: DPU-4 (EPIC: InFMon — flow telemetry service on BF-3)
+Depends on: 000-overview (system overview), 003-erspan-and-packet-parsing
+           (defines the parsed inner-packet record this spec keys off of)
+Related: 004-backend-architecture (consumes this model in the VPP plugin),
+         007-cli (exposes the CRUD surface as `flow {add,rm,list,show}`)
+
+## 1. Purpose
+
```
banidoru, line 14: "a **flow flow-rule**" and "The flow flow-rule is" — mechanical rename artifact. Should be just "**flow-rule**" and "The flow-rule is".

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,367 @@
...
+flow-rule := (name, ordered_field_list, max_keys, eviction_policy)
+```
+
+- **name** — short identifier, `^[a-z0-9][a-z0-9_-]{1,30}$` (i.e. 2–31
+  characters). Used in exported metric labels and CLI references. Unique
+  per backend instance. The 31-character cap is chosen to stay well
+  under common downstream attribute-length limits (OTLP attribute
+  values, VPP node-name tables) without being so short that it
+  constrains operator naming; tighten in a future rev if a concrete
+  downstream limit demands it.
```
banidoru, line 41: Still says `non-negative integer` — previous review flagged this as contradicting §5.3 rule 3 which rejects zero. The reply claimed this was fixed to 'positive integer (≥ 1)' but the change was not applied.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,367 @@
...
+keep that bounded and predictable.
+
+## 2. Concepts
+
+### 2.1 Flow-rule
+
+A **flow-rule** is:
+
+```
+flow-rule := (name, ordered_field_list, max_keys, eviction_policy)
```
banidoru, line 32: Regex is still `^[a-z0-9][a-z0-9_-]{0,30}$` — allows single-char names. The reply claimed it was tightened to `{1,30}` but the change was not applied.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,367 @@
...
+the packet is dropped and `infmon_tracker_drops_total` is incremented
+instead — the flow-rule never silently corrupts.
+
+"Recently used" means *most recently updated by an incoming packet*.
+Recency state lives with the flow and is maintained on every hit; the
+exact data structure is owned by Spec 004 (a likely choice is a
+segmented LRU keyed off the existing hash table, but that's an
+implementation detail).
+
+### 6.1 Why drop, not flush
```
banidoru, line 249: `show` output still lists `bucket_count` and `last_packet_ns`. Previous review agreed to rename `last_packet_ns` → `last_seen_ns`, and `bucket_count` should be `flow_count` to match the tracker→flow-rule rename. Neither fix was applied.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,367 @@
...
+  in-flight packet either sees the old flow-rule set or the new one,
+  never a partial state. Implementation note: Spec 004 will likely
+  use RCU-style swap of the flow-rule table.
+- `add` does **not** support replace-by-name. To change a flow-rule,
+  `rm` then `add`. This preserves the rule from §2.1 that a flow-rule's
+  field list is immutable for its lifetime, and forces operators to
+  acknowledge that flows will be dropped.
+- `list` and `show` are read-only and lock-free on the data plane.
+- The static config file (§5) is applied via the same `add`/`rm`
+  primitives at startup; there is no separate "bulk load" path.
```
banidoru, line 285: `budget_exceeded_runtime` is still listed as a drop reason. The previous review agreed to remove it (v1 is single-threaded, can't happen at runtime). The fix was not applied.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,367 @@
...
+```
+key(T, pkt) = enc(f1, pkt) || enc(f2, pkt) || ... || enc(fn, pkt)
+```
+
+Canonical encodings (v1):
+
+- `src_ip`, `dst_ip`: 16 bytes, network byte order, IPv4 mapped as in §3.
+- `ip_proto`: 1 byte.
+- `dscp`: 1 byte, value in `0..=63`, upper bits zeroed.
+
```
banidoru, line 144: `mirror_src_ip` was added to the field set (§3) but its canonical encoding is not listed here in §4. Add: `mirror_src_ip`: 16 bytes, network byte order, IPv4 mapped as in §3.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,367 @@
...
+- `infmon_tracker_buckets{flow-rule}` — gauge, current key count.
+- `infmon_tracker_evictions_total{flow-rule}` — counter (§6).
+- `infmon_tracker_drops_total{flow-rule, reason}` — counter; `reason` ∈
+  `{"eviction_failed"}`. (Earlier drafts also listed a
+  `budget_exceeded_runtime` reason; in v1 the CRUD plane is
+  single-threaded, so cumulative-budget overcommit can only happen at
+  `add` time and is rejected synchronously via `budget_exceeded`
+  (§7.2). The runtime drop reason is therefore not emitted in v1 and
+  is reserved for a future revision in which concurrent admin paths
```
banidoru, line 325: "1-in-N or token-flow" — bad rename artifact. Should be "token-bucket", not "token-flow".

banidoru (changes requested):
Re-review (iteration 2):

- **Mechanical rename artifacts**: find-replace of tracker→flow-rule and bucket→flow created broken phrases (`flow flow-rule`, `token-flow`, `bucket_count` in `show`) throughout the doc.
- **Prior fix not applied**: §2.1 regex still `{0,30}` (allows single-char names), not `{1,30}` as claimed. `max_keys` still says "non-negative" not "positive".
- **Prior fix not applied**: §7 `show` still returns `last_packet_ns`, not `last_seen_ns` as claimed fixed.
- **Prior fix not applied**: §8 still lists `budget_exceeded_runtime` as a drop reason, which was supposedly removed.
- **`mirror_src_ip` missing from §4**: new field added in §3 but its canonical encoding not listed in §4.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,391 @@
...
+| 0.5     | 2026-04-18 | bf3(agent)| Add §2.4 mental-model section (one flow-rule generates one flow per distinct key tuple); align CLI examples with the `flow-rule` / `flow` verb split (Spec 007); rename `infmon_tracker_*` metrics to `infmon_flow_rule_*` (and `_buckets`→`_flows`, `bucket_count`→`flow_count`); promote `mirror_src_ip` to part of the recommended default flow-rule (still opt-in for custom flow-rules). |
+
+Tracking issue: DPU-8 (project InFMon)
+Parent epic: DPU-4 (EPIC: InFMon — flow telemetry service on BF-3)
+Depends on: 000-overview (system overview), 003-erspan-and-packet-parsing
+           (defines the parsed inner-packet record this spec keys off of)
+Related: 004-backend-architecture (consumes this model in the VPP plugin),
+         007-cli (exposes the flow-rule CRUD as `flow-rule {add,rm,list,show}`
+         and live flow inspection as `flow {list,show}`)
+
```
banidoru, line 12: "a **flow flow-rule**" and "The flow flow-rule is" — mechanical rename artifact still present from iteration 1. Should be just "**flow-rule**".

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,391 @@
...
+keep that bounded and predictable.
+
+## 2. Concepts
+
+### 2.1 Flow-rule
+
+A **flow-rule** is:
+
+```
+flow-rule := (name, ordered_field_list, max_keys, eviction_policy)
```
banidoru, line 32: Regex is still `{0,30}` — allows single-char names. Previous review agreed to tighten to `{1,30}` (2–31 chars minimum). Not yet applied.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,391 @@
...
+```
+
+- **name** — short identifier, `^[a-z0-9][a-z0-9_-]{1,30}$` (i.e. 2–31
+  characters). Used in exported metric labels and CLI references. Unique
+  per backend instance. The 31-character cap is chosen to stay well
+  under common downstream attribute-length limits (OTLP attribute
```
banidoru, line 38: Still says "non-negative integer" which includes zero, contradicting §5.3 rule 3 (`missing, negative, or zero`). Should be "positive integer (≥ 1)" as agreed in iteration 1.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,391 @@
...
+
+`mirror_src_ip` is the documented exception to the "inner only" rule: it
+travels with the parser record (Spec 003 §4.5) so flow-rules can attribute
+flows to the device that mirrored them. All other outer fields remain
+unavailable.
+
+### 3.1 Why IPv4-mapped-IPv6
+
+So a single flow-rule can carry both address families with one fixed-width
+key. A v4 packet and a v6 packet that happen to encode the same address
```
banidoru, line 116: `mirror_src_ip` was added to the field set (§3) but its canonical encoding is not listed here. Add: `mirror_src_ip`: 16 bytes, network byte order, IPv4 mapped as in §3.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,391 @@
...
+Validation is **all-or-nothing** per reload. A failed reload leaves the
+previously running flow-rule set untouched.
+
+## 6. Eviction policy (v1)
+
+v1 ships exactly one policy: **`lru_drop`**.
+
+Behaviour, when a packet would create a new key in a flow-rule that is
+already at `max_keys`:
+
```
banidoru, line 224: `show` output still lists `bucket_count` and `last_packet_ns`. Previous review agreed to rename to `flow_count` and `last_seen_ns` to match the tracker→flow-rule/bucket→flow rename. Not yet applied.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,391 @@
...
+fast path predictable and surfaces the loss as a first-class metric so
+operators can size `max_keys` against observed eviction rate.
+
+## 7. CRUD API surface
+
+The backend exposes a small management API consumed by `infmon-cli`
+(Spec 007) and by the config loader. Transport is owned by Spec 004
+(likely a Unix socket carrying length-prefixed protobuf or JSON); this
+spec defines only the operations and their semantics.
+
```
banidoru, line 262: `budget_exceeded_runtime` is still listed as a drop reason. Previous review agreed to remove it (v1 is single-threaded CRUD, can't happen at runtime). Not yet applied.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,391 @@
...
+  (§7.2) can be produced by a CLI `flow add` that would push the
+  cumulative `max_keys` past rule 6's limit. This means file-order does
+  **not** affect whether a valid file loads, but it does affect the
+  order of error reporting if the validator is bypassed (e.g. a future
+  partial-reload path).
+
+### 7.2 Errors
+
+A small, closed set:
+
```
banidoru, line 300: "token-flow" — bad rename artifact. Should be "token-bucket".

banidoru (changes requested):
Iteration 2 review. The tracker→flow-rule rename and `mirror_src_ip` addition landed, but several issues flagged in previous reviews remain unfixed in the current text:

- §2.1 regex still `{0,30}` (allows single-char names); agreed fix was `{1,30}`
- §2.1 `max_keys` still says "non-negative integer" — contradicts §5.3 rule 3 rejecting zero
- §7 `show` output still references `bucket_count` and `last_packet_ns` instead of `flow_count` and `last_seen_ns`
- §8 still lists `budget_exceeded_runtime` as a drop reason (agreed to remove for v1)
- §4 missing `mirror_src_ip` canonical encoding
- §1 has "flow flow-rule" double-word artifact (lines 20, 22)

r12f:

specs/002-flow-tracking-model.md
r12f: Fixed in 3a48ce6 (the §2 mental-model rewrite). The current §2.x prose uses just "a **flow-rule**" / "The flow-rule is" — no "flow flow-rule" artifact remains around line 20.

specs/002-flow-tracking-model.md
r12f: Fixed in a8f351c — regex is now `^[a-z0-9][a-z0-9_-]{1,30}$` (2–31 chars), with a sentence motivating the 31-char upper bound. Resolving.

specs/002-flow-tracking-model.md
r12f: Done in dd9f1b6 — `token-flow` → `token-bucket` in §9.4.

specs/002-flow-tracking-model.md
r12f: Fixed in a8f351c — `max_keys` is now "positive integer (≥ 1); … Required.", consistent with §5.3 rule 3. Resolving.

specs/002-flow-tracking-model.md
r12f: Done in dd9f1b6 — added `mirror_src_ip: 16 bytes, network byte order, IPv4 mapped as in §3.` to the canonical-encodings list in §4.

specs/002-flow-tracking-model.md
r12f: Fixed in 3a48ce6 — `show` output now lists `flow_count` and `last_seen_ns` (with the explicit "max `last_seen_ns` across resident flows" note). Resolving.

specs/002-flow-tracking-model.md
r12f: Fixed in a8f351c — the only remaining mention of `budget_exceeded_runtime` is in §8 as a historical note explaining *why* it's not emitted in v1. The active drop-reason set is now just `{eviction_failed}`, and over-budget `add` is rejected synchronously via `budget_exceeded` (§7.2). Resolving.

banidoru (approved):
- All prior review threads resolved — regex, max_keys, naming, budget enforcement, metric renames, and artifact fixes all landed correctly.
- §2.4 mental-model section is a welcome addition; clearly separates the configured matcher from live flows.
- Default flow-rule now includes `mirror_src_ip` — good call for multi-source deployments; the opt-out path is documented.
- Minor: file ends without a trailing newline (`\ No newline at end of file`). Not blocking.
- Clean iteration — no new issues found.

banidoru (approved):
- All outstanding review comments have been addressed.
- The spec is now consistent and the naming (, ) has been applied throughout.
- The addition of the mental model section and clearer definitions for fields like  have improved the document.
- No new issues found.

banidoru (approved):
- All outstanding review comments have been addressed.
- The spec is now consistent and the naming (, ) has been applied throughout.
- The addition of the mental model section and clearer definitions for fields like  have improved the document.
- No new issues found.

banidoru (approved):
- All outstanding review comments have been addressed.
- The spec is now consistent and the naming ('flow-rule', 'flow') has been applied throughout.
- The addition of the mental model section and clearer definitions for fields like 'mirror_src_ip' have improved the document.
- No new issues found.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,392 @@
...
+  acknowledge that flows will be dropped.
+- `list` and `show` are read-only and lock-free on the data plane.
+- The static config file (§5) is applied via the same `add`/`rm`
+  primitives at startup; there is no separate "bulk load" path.
+  Concretely: the loader first runs the §5.3 validation rules over the
+  whole file (so order-independent failures — duplicate names, unknown
+  fields, the total `max_keys` budget in rule 6 — are reported up
+  front), then issues `add` calls one at a time. Each `add` also
+  re-checks the running budget, so the same `budget_exceeded` error
+  (§7.2) can be produced by a CLI `flow add` that would push the
```
banidoru: Should be `flow-rule add` (not `flow add`) — the CLI verb was renamed to `flow-rule` in this iteration. Same issue on line 315 in §7.2.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,392 @@
...
+## 10. Open questions
+
+- **Q1.** Should `show` include a top-N sample of live keys for
+  debugging, or is that the job of a separate `flow dump` command? —
+  Defer to Spec 007 (CLI) review.
+- **Q2.** Do we need a "drain" semantic on `rm` (flush flows to OTLP
+  before destroying)? Cheap to add later; v1 says no, to keep the
+  backend simple.
+- **Q3.** Hash collision behaviour — Spec 004 territory, mentioned
+  here so the cross-link exists.
```
banidoru: Missing trailing newline. Minor, but POSIX convention and avoids `\ No newline at end of file` noise in diffs.

banidoru (commented):
All prior review threads have been addressed and resolved — nice work on the comprehensive fixes.

New diff observations:
- Two instances of `flow add` in §7.1 and §7.2 should be `flow-rule add` to match the renamed CLI verb used everywhere else
- File is missing a trailing newline (POSIX convention, avoids noisy diffs)
- Version history entries v0.5/v0.6 have trailing whitespace in the Author column (cosmetic)
- Otherwise, all feedback from iterations 1–2 is correctly applied: regex tightened, positive integer, DSCP extraction explicit, budget enforcement clarified, metrics renamed, mental model section added, rename artifacts cleaned up

banidoru (approved):
All reviewers approved. LGTM.

specs/002-flow-tracking-model.md
banidoru: Still open — lines 291 and 307 still read `flow add` instead of `flow-rule add`.

specs/002-flow-tracking-model.md
banidoru: Still open — file still ends without a trailing newline.

banidoru (commented):
- Version history collapsed from 6 entries to 1 — cleaner, but loses the incremental change trail; fine if the team prefers squashed history
- Dependency/related links reformatted with relative markdown links — good improvement
- Two previously-flagged issues remain unaddressed: `flow add` → `flow-rule add` (lines 291, 307) and missing trailing newline
- No new issues introduced by this diff

specs/002-flow-tracking-model.md
banidoru: Still open — lines 291 and 307 still read `flow add` instead of `flow-rule add`.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,384 @@
...
+  acknowledge that flows will be dropped.
+- `list` and `show` are read-only and lock-free on the data plane.
+- The static config file (§5) is applied via the same `add`/`rm`
+  primitives at startup; there is no separate "bulk load" path.
+  Concretely: the loader first runs the §5.3 validation rules over the
+  whole file (so order-independent failures — duplicate names, unknown
+  fields, the total `max_keys` budget in rule 6 — are reported up
+  front), then issues `add` calls one at a time. Each `add` also
+  re-checks the running budget, so the same `budget_exceeded` error
+  (§7.2) can be produced by a CLI `flow add` that would push the
```
banidoru: Still reads `flow add` — should be `flow-rule add` to match the renamed CLI verb. Same issue on line 307.

banidoru (commented):
Iteration 4 re-review (version history cleanup commit):

- Version history consolidation into a single v0.1 entry looks clean
- Dependency links now use relative markdown refs — good improvement
- **Still open:** `flow add` → `flow-rule add` not fixed on lines 291 and 307
- **Still open:** Missing trailing newline at EOF (line 376)
- No new issues introduced by this commit

specs/002-flow-tracking-model.md
banidoru: Still open — file ends with `.` (0x2E), no trailing newline after this commit.

r12f: Fixed — added trailing newline.

specs/002-flow-tracking-model.md
r12f: Fixed — changed `flow add` to `flow-rule add` on lines 265, 291, and 307.

banidoru (approved):
- All previously open threads (CLI verb rename `flow add` → `flow-rule add`, trailing newline) are now fixed
- Diff between b99fd10 and 8d19c9a contains only those two targeted fixes — no regressions introduced
- No new issues found in the updated text
- Spec is consistent and ready to merge

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,384 @@
...
+  acknowledge that flows will be dropped.
+- `list` and `show` are read-only and lock-free on the data plane.
+- The static config file (§5) is applied via the same `add`/`rm`
+  primitives at startup; there is no separate "bulk load" path.
+  Concretely: the loader first runs the §5.3 validation rules over the
+  whole file (so order-independent failures — duplicate names, unknown
+  fields, the total `max_keys` budget in rule 6 — are reported up
+  front), then issues `add` calls one at a time. Each `add` also
+  re-checks the running budget, so the same `budget_exceeded` error
+  (§7.2) can be produced by a CLI `flow-rule add` that would push the
```
banidoru, line 291: This should be `flow-rule add` to match the renamed CLI verb in Spec 007.

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,384 @@
...
+## 10. Open questions
+
+- **Q1.** Should `show` include a top-N sample of live keys for
+  debugging, or is that the job of a separate `flow dump` command? —
+  Defer to Spec 007 (CLI) review.
+- **Q2.** Do we need a "drain" semantic on `rm` (flush flows to OTLP
+  before destroying)? Cheap to add later; v1 says no, to keep the
+  backend simple.
+- **Q3.** Hash collision behaviour — Spec 004 territory, mentioned
+  here so the cross-link exists.
```
banidoru, line 384: Nit: file is missing a trailing newline.

banidoru (changes requested):
- The CLI verb in §7.1 is still `flow add` — should be `flow-rule add`.
- Nit: missing trailing newline.

specs/002-flow-tracking-model.md
banidoru: Fixed in 8d19c9a — line 291 now reads `flow-rule add`.

specs/002-flow-tracking-model.md
banidoru: Fixed in 8d19c9a — file now ends with a trailing newline.

banidoru (approved):
- Both remaining open threads resolved: `flow add` → `flow-rule add` (line 291/307) and trailing newline added
- New diff (b99fd10→8d19c9a) contains only these two fixes — clean and correct
- No new issues found
- All prior review threads now resolved
- LGTM

specs/002-flow-tracking-model.md
```diff
@@ -0,0 +1,384 @@
...
+  per backend instance. The 31-character cap is chosen to stay well
+  under common downstream attribute-length limits (OTLP attribute
+  values, VPP node-name tables) without being so short that it
+  constrains operator naming; tighten in a future rev if a concrete
+  downstream limit demands it.
+- **ordered_field_list** — non-empty, ordered tuple of fields drawn from
+  the v1 field set (§3). Order is significant: it defines the byte
+  layout of the key (§4) and therefore the hash. Reordering produces a
+  different flow-rule even with the same field set.
+- **max_keys** — positive integer (≥ 1); upper bound on the number of
```
r12f, line 46: Do you mean max_flows?

r12f: `max_keys` is the correct term here — it refers to the upper bound on distinct *keys* (not flows/flow-rules) held simultaneously within a single flow-rule. A flow-rule has one `max_keys` cap; each distinct key tuple creates one bucket. The naming follows the tracker/key/bucket hierarchy defined in §2.

PR #6, opened 2026-04-18
specs: add 003 — ERSPAN and packet parsing (DPU-9)

r12f:
## Summary

Adds `specs/003-erspan-and-packet-parsing.md` — the parsing contract between
the wire (BlueField-3 ERSPAN III mirror) and the rest of the InFMon pipeline.

## What this spec pins down

- **ERSPAN Type III over GRE** (proto `0x22EB`) is the only accepted transport.
- All outer headers (Ethernet, optional single VLAN, IPv4/IPv6, GRE, ERSPAN III
  header + optional Platform-Specific Sub-Header) are skipped; **no outer
  field is exposed downstream**.
- The **ERSPAN Session ID is explicitly NOT exposed** to downstream stages,
  exporters, or logs. Negative test required.
- Match always begins at the **inner packet** — `(inner_ptr, inner_len,
  inner_truncated)`.
- A **single inner-decap hook** (`infmon_inner_decap`) reserves room for one
  future encapsulation layer (VXLAN / GENEVE / RoCEv2). v1 ships it disabled.
- **~128 B truncated snap** handling: parser never overreads; outer/ERSPAN
  truncation is fatal; inner truncation is non-fatal and downstream applies
  partial-header rules (missing fields are `unknown`, not zero).
- A complete per-reason counter table (`parsed_ok`, `inner_truncated_ok`,
  `gre_unexpected_flags`, `erspan_bad_version`, etc.).
- **RoCEv2** is reserved as a *design seam only* — BTH parsing will land
  through the inner-decap hook in a future, additive PR/spec.
- Test plan: golden PCAPs, gtest unit tests (incl. negative session-ID test),
  libFuzzer harness in CI; E2E replay stays out of CI per EPIC.

## Out of scope

- Implementation. This PR is **spec-only**.
- ERSPAN I/II, multi-session demux, fragment reassembly, BTH semantics,
  performance tuning.

## Tracking

- Issue: DPU-9
- Parent EPIC: DPU-4 (InFMon)
- Reviewer: @banidoru

🐰 Generated with assistance from Hermes Agent.


specs/003-erspan-and-packet-parsing.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+Outer headers exist only to locate the inner packet. **No field from any outer
+header is propagated downstream**, including source/destination IP of the
+mirror transport. Telemetry must describe the mirrored flow, not the mirror
+infrastructure.
+
+### 4.2 Outer Header Skip
+
+| Layer    | Action                                                                                                              |
```
banidoru, line 97: The ASCII diagram bit-field widths don't add up correctly. `Ver` is shown as 3 bits but ERSPAN III uses a 4-bit version field. `Session ID` appears to span a variable number of bits. The diagram should have exact bit boundaries matching `draft-foschiano-erspan-03` — misaligned bit widths in a spec will propagate to the implementation as wrong masks/shifts.

Suggest adding explicit bit ranges (e.g., `Ver[0:3]`, `VLAN[4:15]`, `COS[16:18]`, etc.) or fixing the ASCII art to be pixel-accurate.

specs/003-erspan-and-packet-parsing.md
```diff
@@ -0,0 +1,318 @@
...
+- Wire format: `Ethernet [VLAN] → IPv4|IPv6 → GRE(proto=0x22EB) → ERSPAN III →
+  inner packet`.
+- GRE flags: only `S` (sequence) MAY be set. `C` (checksum) and `K` (key) MUST
+  NOT be set; if they are, the packet is dropped (counter
+  `gre_unexpected_flags`).
+- Mirror source: BlueField-3 hardware mirror, configurable snap length,
+  default ~128 bytes total wire length.
+- Buffer contract: the parser receives a single contiguous DPDK mbuf segment.
+  Multi-segment mbufs are linearised by the VPP input node before the parser
+  runs (the input node is out of scope here; it is mentioned for clarity).
```
banidoru, line 71: Dropping *all* outer IPv6 packets with extension headers is heavy-handed. Hop-by-hop options (next-header 0) are commonly inserted by network infrastructure. If the BF-3 mirror transport ever traverses such a path, every mirrored packet would be silently dropped.

Consider: parse at minimum hop-by-hop and destination options (they have a known `Hdr Ext Len` field), skip them, and only drop on unknown/unsupported extension types. This keeps the parser simple while avoiding a class of silent data loss.

specs/003-erspan-and-packet-parsing.md
```diff
@@ -0,0 +1,318 @@
...
+explicit feature flag and a new spec.
+
+### 4.5 Inner Packet — Match Always Begins Here
+
+The "inner packet" begins at the first byte after the ERSPAN III header (and
+Platform Sub-Header, if present). All downstream parsers, matchers, and
+feature extractors **see only the inner packet**.
+
+The parser returns:
+
```
banidoru, line 205: `declared_inner_len` is computed from "outer IP total length" — but IPv4 `Total Length` includes the IP header itself while IPv6 `Payload Length` does not include the 40-byte fixed header. The spec should give the explicit formula for each:

- **IPv4:** `declared_inner_len = total_length - IHL*4 - gre_len - erspan_len`
- **IPv6:** `declared_inner_len = payload_length - gre_len - erspan_len`

Without this, an implementer could easily get an off-by-40 error on IPv6.

specs/003-erspan-and-packet-parsing.md
```diff
@@ -0,0 +1,318 @@
...
+  `gre_unexpected_flags`).
+- Mirror source: BlueField-3 hardware mirror, configurable snap length,
+  default ~128 bytes total wire length.
+- Buffer contract: the parser receives a single contiguous DPDK mbuf segment.
+  Multi-segment mbufs are linearised by the VPP input node before the parser
+  runs (the input node is out of scope here; it is mentioned for clarity).
+- The parser is a **pure function of bytes**: no allocations, no syscalls, no
+  per-packet logging, no cross-packet state.
+
+## 4. Parsing Pipeline
```
banidoru, line 79: GRE `version=0` is required (§4.2), but the counter table in §6 has no counter for `gre_bad_version`. A GRE header with version != 0 would fail the version check but there's no explicit drop reason. Add `gre_bad_version` to the counter table, or clarify that non-zero GRE version manifests as one of the existing counters.

specs/003-erspan-and-packet-parsing.md
```diff
@@ -0,0 +1,318 @@
...
+```c
+typedef enum {
+    INFMON_DECAP_NONE = 0,    /* identity, v1 default */
+    INFMON_DECAP_VXLAN,       /* future */
+    INFMON_DECAP_GENEVE,      /* future */
+    INFMON_DECAP_ROCEV2,      /* future, see §7 */
+} infmon_inner_decap_t;
+
+/* Returns 0 on success; sets *out_ptr / *out_len to the de-encapsulated view.
+ * Returns nonzero on error (caller drops the packet, increments a per-decap
```
banidoru, line 249: For the inner L4 partial-header rule: "partially present" ports (e.g., only 2 of 4 bytes available) are treated as absent. But what about SCTP or other L4 protocols? The spec only mentions TCP/UDP. Should non-TCP/UDP L4 be dropped, treated as `unknown`, or extracted best-effort? Worth a one-liner to close this gap.

specs/003-erspan-and-packet-parsing.md
```diff
@@ -0,0 +1,318 @@
...
+
+The 10-bit ERSPAN Session ID **MUST NOT** appear in any record handed to
+downstream stages, exporters, or logs. Rationale:
+
+- Session IDs describe the mirroring fabric (which SPAN session copied the
+  packet), not the mirrored flow.
+- Surfacing them couples downstream consumers to operator topology.
+- They leak mirror configuration into telemetry, which is a small but real
+  information disclosure.
+
```
banidoru, line 188: The inner-decap hook takes `uint32_t in_len` but the outer parser works with mbuf pointers and length. Should the spec mandate that the decap hook also receives a `bool truncated` flag from the outer parser? If the inner packet is already truncated, the decap hook needs to know — otherwise it might try to parse a VXLAN/GENEVE header that isn't fully present and overread.

specs/003-erspan-and-packet-parsing.md
```diff
@@ -0,0 +1,318 @@
...
+   drop, `inner_l3_truncated`. Without it the 5-tuple cannot be formed and the
+   record is useless for flow telemetry.
+3. **Inner L4 header** — best-effort:
+   - Ports (first 4 B of TCP/UDP): if **fully** present, extract; if
+     **partially** present, treat as absent and mark `flow_key_partial = true`
+     on the record.
+   - TCP flags / window / sequence / ack: extract only fields whose bytes are
+     fully present. Missing fields are reported as `unknown` (NOT zero — zero
+     would corrupt aggregates downstream).
+4. **L4 payload** — only the *observed length* (`l4_payload_observed_len`) is
```
banidoru, line 311: The golden PCAPs list doesn't include an **IPv6 outer** variant. Since the parser supports both IPv4 and IPv6 outer headers (§4.2), at least one `erspan3_ipv6_*.pcap` should be required to prevent an untested code path from shipping.

banidoru (commented):
- Solid spec overall — clear separation of outer/inner, good truncation contract, well-defined counter table.
- ERSPAN III header diagram has bit-field widths that don't match the RFC draft (Session ID shown as variable width, Ver as 3 bits but should be 4 in some interpretations). Verify against `draft-foschiano-erspan-03` and add exact bit positions.
- IPv6 extension header handling is too aggressive — dropping all packets with any extension header will silently discard legitimate traffic if the mirror transport ever uses hop-by-hop or destination options. Consider parsing at least the well-known ones to skip them.
- The `declared_inner_len` derivation from outer IP total length is underspecified for IPv6 (payload length vs total length semantics differ). Needs explicit formula.
- Missing counter for GRE version != 0 — currently only bad proto and unexpected flags are counted, but a non-zero GRE version would pass the flag checks.

banidoru (changes requested):
- The spec needs to clarify handling for unsupported outer EtherTypes other than QinQ.
- The handling of the ERSPAN Timestamp field is not specified.
- The test plan is missing a crucial edge case for a truncated platform-specific sub-header.
- The description of the future RoCEv2 decapitation is inconsistent.

banidoru:
The table in §4.2 drops packets for QinQ (0x88a8), but what about other non-IP outer frame types like MPLS (0x8847) or even ARP (0x0806)? These would currently be dropped silently. Suggest adding an `outer_ethertype_unsupported` counter for any EtherType that isn't 0x0800 (IPv4), 0x86DD (IPv6), or 0x8100 (VLAN).

banidoru:
The spec is clear that the Session ID is not exposed, but it's silent on the 32-bit ERSPAN Timestamp field. To avoid ambiguity, this section should explicitly state that the Timestamp field is also **not** propagated downstream, for the same reasons.

banidoru:
The golden PCAP list covers the O bit and truncation separately, but it's missing the critical combination: a packet with the O bit set where the mbuf is truncated *during* the 8-byte Platform-Specific Sub-Header. This is a key bounds-check test case that should be explicitly required. Suggest adding `erspan3_o_bit_truncated.pcap`.

banidoru:
This section states the inner-decap hook will turn the view from `Eth | IP | UDP | BTH | …` into `BTH | …`. This seems incorrect; the purpose of decap is to *remove* the transport header. The resulting view should be the BTH *payload*, not the BTH itself. The description should be clarified to state the BTH is parsed and its payload is returned as the new inner packet view.

specs/003-erspan-and-packet-parsing.md
banidoru, line 97: The ERSPAN III `T` bit (truncated-by-source) is present in this diagram but the spec never references it. §5 computes truncation purely from length comparison (`inner_len < declared_inner_len`). This is a valid design choice, but it should be explicitly stated: "The ERSPAN `T` bit is not consumed; truncation status is derived solely from length comparison." Otherwise an implementer might wonder whether `T=1` should force `inner_truncated = true` even when lengths say the packet is complete.

specs/003-erspan-and-packet-parsing.md
banidoru, line 79: GRE flag handling only specifies behavior for `C`, `K`, and `S` flags. RFC 2784/2890 also defines reserved bits and the `R` (routing) flag (deprecated but could appear on wire). The spec should state whether *any* flag besides `S` being set triggers `gre_unexpected_flags`, or only `C` and `K` specifically. A strict allowlist ("only bit 12 (`S`) may be set; any other non-zero flag bit → drop") is safer and simpler to implement than enumerating individual forbidden flags.

specs/003-erspan-and-packet-parsing.md
banidoru, line 249: "Missing fields are reported as `unknown` (NOT zero)" — but the spec doesn't define the concrete representation of `unknown`. For a `uint16_t` port field, is it `0xFFFF`? For TCP flags (`uint8_t`), is it `0xFF`? Or is there a parallel validity bitmap (`port_valid`, `flags_valid`, etc.)?

Without pinning this, two independent implementers will pick different sentinels and downstream consumers will disagree on what "unknown" looks like. Suggest adding a short table: field → type → sentinel value, or mandating a `valid_fields` bitmask on the output struct.

specs/003-erspan-and-packet-parsing.md
```diff
@@ -0,0 +1,318 @@
...
+- IP fragment reassembly (outer or inner).
+- Outer IP / GRE checksum verification (mirror copies are trusted).
+- Stateful or cross-packet logic.
+- RoCEv2 BTH parsing — only the design seam is reserved (§7).
+- Performance tuning (batching, NUMA, prefetch). Covered later.
+
+## 3. Inputs and Assumptions
+
+- Wire format: `Ethernet [VLAN] → IPv4|IPv6 → GRE(proto=0x22EB) → ERSPAN III →
+  inner packet`.
```
banidoru, line 63: The spec assumes a "single contiguous DPDK mbuf segment" and delegates linearization to the VPP input node. But there's no counter for the case where linearization fails or a multi-segment mbuf slips through. Consider adding an `mbuf_not_contiguous` counter (or at minimum an assert) so this invariant violation is observable in production rather than manifesting as a silent overread.

specs/003-erspan-and-packet-parsing.md
```diff
@@ -0,0 +1,318 @@
...
+| `inner_truncated_ok`          | accepted with `inner_truncated = true` and a usable 5-tuple        |
+| `outer_qinq_unsupported`      | dropped: stacked VLAN on outer                                    |
+| `outer_v6_ext_unsupported`    | dropped: outer IPv6 extension headers present                     |
+| `outer_truncated`             | dropped: outer headers do not fit in the mbuf                     |
+| `gre_unexpected_flags`        | dropped: GRE C or K flag set                                      |
+| `gre_bad_proto`               | dropped: GRE protocol ≠ `0x22EB`                                  |
+| `erspan_bad_version`          | dropped: ERSPAN Ver ≠ 2                                           |
+| `erspan_truncated`            | dropped: ERSPAN III (or Platform Sub-Header) does not fit         |
+| `inner_eth_truncated`         | dropped: inner Ethernet missing                                   |
+| `inner_l3_truncated`          | dropped: inner L3 fixed header missing                            |
```
banidoru, line 342: Fuzz harness max input length of 256 B is tight. A non-truncated packet with IPv6 outer (40 B) + Ethernet (14 B) + GRE+seq (8 B) + ERSPAN III (12 B) + platform sub-header (8 B) = 82 B of overhead, leaving only 174 B for the inner packet — which won't cover a full inner IPv6 + TCP with options. Consider 512 B or making the limit configurable so the fuzzer can exercise larger-than-snap packets that the parser should also handle correctly.

banidoru (commented):
- Spec is well-structured with clear layering and good truncation contract.
- The ERSPAN `T` (truncated) bit is silently ignored — spec should explicitly document this design choice.
- `unknown` sentinel for missing inner fields is underspecified — implementers need a concrete representation.
- GRE flag handling only covers C/K/S but doesn't address other RFC 2784 flags (R, reserved bits).
- Fuzz harness max input of 256 B may be too small for non-truncated packets with IPv6 outer headers.

r12f:

specs/003-erspan-and-packet-parsing.md
r12f: Good catch. I've kept the ASCII art as indicative only and added an authoritative bit-range table immediately below it (Word 0 / 1 / 2 with explicit `Ver[0:3]`, `VLAN[4:15]`, `COS[16:18]`, `BSO[19:20]`, `T[21]`, `Session ID[22:31]`, `SGT[64:79]`, `P[80]`, `FT[81:85]`, `Hw ID[86:91]`, `D[92]`, `Gra[93:94]`, `O[95]`). The spec now explicitly tells implementers to mask/shift from those ranges, not from the art.

specs/003-erspan-and-packet-parsing.md
r12f: Agreed — silent loss on hop-by-hop options would be bad. Updated §4.2 IPv6 row: Hop-by-Hop Options (next-header 0) and Destination Options (next-header 60) are now parsed and skipped using their `Hdr Ext Len` field; only other extension headers (Routing, Fragment, AH, ESP, Mobility, …) trigger `outer_v6_ext_unsupported`. The final next-header after skipping must still be 47.

specs/003-erspan-and-packet-parsing.md
r12f: Good point — the IPv4-vs-IPv6 length-field semantics are exactly the trap. §4.5 now spells out both formulas explicitly (`ipv4.total_length - IHL*4 - gre_len - erspan_len` for v4; `ipv6.payload_length - v6_ext_len - gre_len - erspan_len` for v6, noting payload_length excludes the 40-byte fixed header), plus the `erspan_len` and `gre_len` definitions.

specs/003-erspan-and-packet-parsing.md
r12f: Added `gre_bad_version` to the §6 counter table and called it out explicitly in the §4.2 GRE row ("Non-zero GRE version → drop, `gre_bad_version`"). Now there's a dedicated drop reason rather than smuggling it into one of the other counters.

specs/003-erspan-and-packet-parsing.md
r12f: §5.2 rule 3 now states that v1 only does port/flag extraction for TCP and UDP. For SCTP, ICMP/ICMPv6, GRE, ESP, AH, etc., the packet is still accepted (`parsed_ok` / `inner_truncated_ok`) but `PORTS_VALID = 0` and `flow_key_partial = true`; per-protocol extractors are deferred to future specs.

specs/003-erspan-and-packet-parsing.md
r12f: Right — the decap hook needs to know. Updated the C signature in §4.6: it now takes `bool in_truncated` and writes `bool *out_truncated`, with a doc-comment explaining propagation prevents overread on a partially-present VXLAN/GENEVE/BTH header.

specs/003-erspan-and-packet-parsing.md
r12f: Added two required golden PCAPs to §8.1: `erspan3_ipv6_full.pcap` (IPv6 outer, full inner) and `erspan3_ipv6_trunc128.pcap` (IPv6 outer with 128 B snap). The IPv6 outer code path is no longer untested-by-spec.

specs/003-erspan-and-packet-parsing.md
r12f: Added an explicit paragraph after the §4.3 `Ver == 2` check: "The ERSPAN `T` bit (truncated-by-source) is **not consumed**; truncation status is derived solely from the length comparison in §5. Even when `T == 1` and lengths indicate a complete inner packet, `inner_truncated` remains `false`." Single-sourced as you suggested.

specs/003-erspan-and-packet-parsing.md
r12f: Adopted the strict allowlist. §4.2 GRE row now reads: "only the `S` bit (sequence) MAY be set; any other non-zero flag bit (`C`, `K`, `R`, or any reserved/recursion bit) → drop, `gre_unexpected_flags`." The §6 description matches.

specs/003-erspan-and-packet-parsing.md
r12f: Pinned it. §5.2 rule 3 now mandates a parallel `valid_fields` bitmask on the parser output struct (`PORTS_VALID`, `TCP_FLAGS_VALID`, `TCP_SEQ_VALID`, `TCP_ACK_VALID`, `TCP_WINDOW_VALID`, …). The field's storage value is *undefined* when its valid-bit is 0 and MUST NOT be read — that avoids the ambiguity of consumers picking ad-hoc `0xFF…` sentinels that collide with real wire values.

specs/003-erspan-and-packet-parsing.md
r12f: Added `mbuf_not_contiguous` to the §6 counter table and updated §3 to make the parser drop+count (with a debug assert) instead of risking a silent overread when the linearization invariant is violated. Production-observable now.

specs/003-erspan-and-packet-parsing.md
r12f: Bumped the §8.3 default to 2048 B and made it configurable via the harness, so fuzzing covers both the BF-3 ~128 B snap regime and full-MTU/jumbo mirrors. The 256 B cap was indeed leaving IPv6+TCP-with-options under-fuzzed.

r12f:
Addressing the four top-level review comments (all overlap with inline threads, but consolidating here for the record):

1. **`outer_ethertype_unsupported` for non-IP/non-VLAN EtherTypes** — added. §4.2 Ethernet row now drops anything outside {`0x0800`, `0x86DD`, `0x8100`} (MPLS, ARP, S-tag, etc.) with the new `outer_ethertype_unsupported` counter listed in §6.
2. **ERSPAN Timestamp not propagated** — added a paragraph in §4.3 stating the 32-bit Timestamp field is **not** propagated downstream, for the same reasons given for Session ID in §4.4.
3. **`erspan3_o_bit_truncated.pcap`** — added to §8.1 (`O = 1` with the mbuf truncated *inside* the 8-byte Platform-Specific Sub-Header).
4. **§7 RoCEv2 wording** — clarified: the inner-decap hook returns the **BTH payload**, not the BTH itself; BTH fields are parsed for the RoCEv2 record but removed from the returned view, mirroring how other transport headers are handled. View goes from `Eth | IP | UDP | BTH | payload` → `payload`.

Pushed in f3a3c0b.

banidoru (approved):
- All 12 prior review threads addressed and resolved — bit-range table, IPv6 ext-header parsing, dual IPv4/IPv6 length formulas, GRE strict allowlist, `valid_fields` bitmask, `mbuf_not_contiguous` counter, fuzz harness bump to 2048 B, and others all look solid.
- New `mirror_src_ip` addition is well-scoped: tagged union, single outer-header field crossing the boundary, clear rationale in §2 and §4.5.
- ERSPAN `T`-bit non-consumption is now explicitly documented with unambiguous semantics — good.
- `erspan3_o_bit_truncated.pcap` golden PCAP is a valuable addition for the Platform Sub-Header bounds-check edge case.

banidoru (approved):
- All 12 prior review threads addressed and resolved — fixes match the concerns raised.
- New `mirror_src_ip` surface area is well-scoped: tagged union, single field crossing the parser boundary, session ID still stripped.
- IPv4/IPv6 `declared_inner_len` formulas now explicit and correct; `v6_ext_len` subtraction included.
- `valid_fields` bitmask approach for L4 extraction is clean — avoids sentinel collisions.
- Golden PCAPs, counter table, and fuzz harness updates are consistent with the spec changes.

banidoru (approved):
All reviewers approved. LGTM.

r12f:

specs/003-erspan-and-packet-parsing.md
```diff
@@ -0,0 +1,435 @@
...
+   leaving out of v1.
+3. GENEVE option TLVs — when GENEVE decap is added, do we surface options or
+   skip them? Defer to the GENEVE spec.
+
+## 10. Acceptance
+
+This spec is **accepted** when merged to `main`. The implementation PR
+(`feat/erspan-parser`) MUST cite this file in its description and update §9
+if any open question is resolved during implementation. The reviewer
+(@banidoru) signs off on both the spec and the implementation PRs separately.
```
banidoru: Nit: file is missing a trailing newline (`\ No newline at end of file`). Most linters and POSIX tools expect one.

banidoru (approved):
- All 12 prior review threads addressed and resolved — fixes look correct.
- New §4.2.1 (`mirror_src_ip` extraction) is well-structured: clear per-AF rules, explicit non-extraction list, unconditional extraction rationale.
- Prose update in §4.1 correctly narrows the "no outer fields" rule with the named exception.
- `Tracking` row removed from metadata table — verify this is intentional.
- File is missing a trailing newline (minor).

banidoru (changes requested):
- The new section on outer source IP extraction is a good addition and well-specified.
- The previously reported issue regarding the missing trailing newline in `specs/003-erspan-and-packet-parsing.md` is still present in the latest commit.

specs/003-erspan-and-packet-parsing.md
banidoru: Still unresolved — the diff still shows `\ No newline at end of file` on the last line.

specs/003-erspan-and-packet-parsing.md
```diff
@@ -0,0 +1,435 @@
+# Spec 003 — ERSPAN and Packet Parsing
+
+## Version history
+
+| Version | Date       | Author      | Changes        |
+| ------- | ---------- | ----------- | -------------- |
+| 0.1     | 2026-04-18 | r12f        | Initial draft. |
+| 0.2     | 2026-04-18 | r12f        | Surface `mirror_src_ip` (outer GRE source IP) to downstream as the only outer-header value that crosses the parser boundary. ERSPAN session ID still stripped. |
+| 0.3     | 2026-04-18 | r12f       | Add §4.2.1 explicit outer-IPv4/IPv6 source-IP extraction → `mirror_src_ip` as a parsed field exposed to flow-keys (Spec 002). ERSPAN session ID remains dropped. Align prose with the flow-rule/flow split. |
```
banidoru: Nit: the v0.3 row has a trailing-space alignment issue — `r12f` has fewer trailing spaces than the other rows, breaking the Markdown table column alignment. Compare:
```
| 0.2     | 2026-04-18 | r12f        | Surface ...
| 0.3     | 2026-04-18 | r12f       | Add ...
```
Missing one space after `r12f` in the 0.3 row.

banidoru (commented):
Iteration 3 review:

- §4.2.1 `mirror_src_ip` extraction is well-specified and clearly scoped — good addition
- Version table author fix (bf3 → r12f) and v0.3 changelog entry look correct
- `Tracking` field removal from metadata: fine, no impact
- Trailing newline still missing (prior thread remains open)
- One minor alignment nit in the version table

specs/003-erspan-and-packet-parsing.md
banidoru: Still not fixed — the file still lacks a trailing newline.

specs/003-erspan-and-packet-parsing.md
banidoru: The alignment nit is gone since the row was removed, but the version history was collapsed from 3 entries to 1 — losing the v0.2 and v0.3 change records. See new inline comment on this.

specs/003-erspan-and-packet-parsing.md
```diff
@@ -0,0 +1,429 @@
+# Spec 003 — ERSPAN and Packet Parsing
+
+## Version history
+
+| Version | Date       | Author       | Changes |
+| ------- | ---------- | ------------ | ------- |
```
banidoru, line 6: Version history was collapsed from 3 rows (v0.1, v0.2, v0.3) into a single v0.1 entry. This loses the incremental change record — reviewers and future readers can't tell what changed between revisions. Please restore the v0.2 and v0.3 rows (with the alignment nit fixed).

specs/003-erspan-and-packet-parsing.md
```diff
@@ -0,0 +1,429 @@
...
+| 0.1     | 2026-04-18 | Riff (r12f)  | Initial draft. Defines ERSPAN III over GRE decap, parser boundary (mirror transport vs user packet), and §4.2.1 explicit outer-IPv4/IPv6 source-IP extraction → `mirror_src_ip` as the only outer-header value crossing the parser boundary into flow-keys (Spec 002). ERSPAN session ID is dropped. |
```
banidoru, line 7: The `| Field | Value |` metadata table (containing the repo URL) was removed in this commit. Was this intentional? If it's being moved elsewhere, note where; otherwise please restore it.

banidoru (changes requested):
- Version history collapsed from 3 entries to 1 — loses v0.2/v0.3 change audit trail. Please preserve per-version rows.
- Repo metadata table (Field/Value with repo URL) removed without explanation — restore or justify.
- Trailing newline still missing (prior open thread).
- The consolidated v0.1 entry conflates the original draft with later additions, making it unclear what changed between revisions.

specs/003-erspan-and-packet-parsing.md
banidoru, line 7: The v0.2 and v0.3 rows were requested to be restored (with alignment fixed), but this commit went the other direction — collapsing everything into a single v0.1 entry. Please restore the 3 separate version rows so the incremental change history is preserved. Also restore the `| Field | Value |` metadata table that was removed.

banidoru (changes requested):
- Trailing newline still missing (prior thread still open)
- Version history regression: v0.2 and v0.3 rows were deleted instead of restored — loses incremental change record
- `| Field | Value |` metadata table (repo URL) still removed
- The single v0.1 row now crams all changes into one entry, making revision tracking impossible
- No other new issues in this diff beyond the above regressions

specs/003-erspan-and-packet-parsing.md
banidoru: Still not fixed in 2cf6858 — file ends without a trailing newline.

specs/003-erspan-and-packet-parsing.md
banidoru: Not addressed — this commit made it worse by collapsing all 3 version rows into a single v0.1 entry. The v0.2 and v0.3 rows with their incremental change descriptions need to be restored.

specs/003-erspan-and-packet-parsing.md
r12f: Fixed — added trailing newline.

specs/003-erspan-and-packet-parsing.md
r12f: Convention: one PR = one version-history row; review-fix changes amend the single row's Changes cell.

specs/003-erspan-and-packet-parsing.md
r12f: Convention: one PR = one version-history row; review-fix changes amend the single row's Changes cell. The metadata table (Repo field) was intentionally removed per project spec conventions.

banidoru (approved):
- All 16 prior review threads resolved — fixes verified
- Only change since last review: trailing newline added (fixes the last open nit)
- No new issues found
- Spec is comprehensive and implementation-ready
- LGTM

banidoru (approved):
LGTM. The only change since last review was adding the trailing newline, which was the last pending nit.

banidoru (approved):
- Only change since last review: trailing newline added (fixes the last open nit)
- All 16 prior review threads resolved with adequate fixes
- No new issues found
- LGTM

banidoru (approved):
All reviewers approved. LGTM.

PR #4, opened 2026-04-18
specs: add 001 — CI and pre-commit

r12f:
Closes DPU-7 (parent: DPU-4 EPIC InFMon).

First spec to land per the spec-first workflow — defines the quality gate before any source code is written.

## What this spec covers

- **Pre-commit hooks**: `rustfmt`, `cargo clippy -D warnings`, `clang-format`, `cppcheck`, `markdownlint`, `commitlint` (Conventional Commits), plus DCO sign-off enforcement.
- **GH Actions workflows**: `lint`, `rust-test`, `cpp-test` (in container with VPP dev headers), `cross-build` (aarch64 for BF-3, compile-only).
- **E2E explicitly NOT in CI** — run manually on `r12f-bf3` via `make e2e`.
- **Branch protection** on `main`: 1 review + CODEOWNERS (@banidoru), required checks (`lint`, `rust-test`, `cpp-test`, `cross-build`, `dco`), linear history, no force-push.
- **VPP container**: proposes `ligato/vpp-base:24.02` (multi-arch, ships `vpp-dev`, Apache-2.0, pinned by digest). Documented fallback: in-repo Dockerfile published to `ghcr.io/r12f/infmon-vpp-dev`.
- **Caching strategy** for cargo (registry/git + `Swatinem/rust-cache`), `ccache`, `pre-commit`, and apt — with concrete cache keys and warm-CI time targets.

## Open questions (in §9, for reviewer input)

1. Self-hosted runner on `r12f-bf3` for optional E2E — defer?
2. Pin VPP LTS to 24.02 vs 24.06?
3. `cross` vs raw cross-compiler for Rust aarch64?

## Process

Per DPU-4: spec merges → impl PR → CI green → @banidoru sign-off → merge.

Sign-off: yes (DCO trailer in commit).

cc @banidoru — review please.

PR #7, opened 2026-04-18
specs: 004 — backend architecture (VPP plugin) (DPU-10)

r12f:
## Summary

Adds `specs/004-backend-architecture.md`, the design spec for `infmon-backend` — the VPP plugin that turns ERSPAN-mirrored packets into per-flow counters on BlueField-3 ARM cores.

Tracks **DPU-10** (parent: DPU-4 — EPIC InFMon).

## What's covered

Per the issue checklist:

- **VPP graph node layout** — linear pipeline `dpdk-input → infmon-erspan-decap → infmon-flow-match → infmon-counter → drop`, all on the RX worker thread (no inter-thread handoff on the data path).
- **Counter table** — bounded Robin-Hood hash, 64 B cache-line-aligned slots, 64-bit atomic `packets`/`bytes` (relaxed adds), per-bucket-group seqlock for occupancy transitions, key blobs in a separate arena. 64-bit width is justified against 100G line-rate wrap time.
- **Stats segment exposure** — tables published into VPP's existing stats segment under `/infmon/<flow_def_id>/<generation>`; frontend mmaps read-only and discovers tables by enumerating the directory.
- **Control surface** — VPP binary API plugin (Unix socket transport) carrying `flow_def_{add,del,list,get}`, `snapshot_and_clear`, and `status` messages. Reuses VPP's existing transport rather than inventing one.
- **`snapshot_and_clear` semantics** — atomic pointer swap to a freshly-allocated empty table; the worker thread keeps counting on the new table from the very next packet, the retired table is immutable and freed only after an RCU-style grace window. No reset bit, no zeroing of live counters, no lost packets.
- **Threading model** — workers handle their pipelines end-to-end, control thread owns API + alloc + retire; RX-queue → worker pinning is required and enforced at startup.
- **Batch sizes** — full `VLIB_FRAME_SIZE` frames, dual-loop nodes with software prefetch, no per-frame allocation.
- **Per-core throughput targets** — concrete numbers on a single BF-3 A78AE core (≥12 Mpps for 64 B all-hit, line rate at 1500 B, ≥6 Mpps with key set spilling L2, ≤5 ms snapshot end-to-end, 0 cycles of worker stall during snapshot).

Plus failure-mode counters surfaced through `show errors` / `infmon_status`, and three explicit open questions (hash function, stats-segment sizing, sharded vs shared tables) deferred to the impl PR or v2.

## Why these choices

- **Atomic table swap over reset bit.** Eliminates the "counter went backwards" race that plagues IPFIX-style implementations and makes each retired snapshot trivially internally-consistent (no seqlock retry needed when reading retired tables).
- **VPP binary API for control.** The frontend already needs to talk to VPP for interface state; running our control plane on a separate transport would double the failure modes for no benefit. Existing Rust bindings (`vpp-api-client`) make spec 005's job easier.
- **Workers count, control thread allocates.** Keeps every blocking operation off the data path — the design rule the spec opens with.

## Process

Spec-first per the project convention (see spec 000). This PR is the spec; once merged with @banidoru's sign-off, the implementation PR for `infmon-backend` may begin and must conform to §4–§9 or amend this spec first.

## Checklist

- [x] Description ≥ all bullets in DPU-10
- [x] Cross-references to specs 000 / 002 / 003 / 005 / 006 / 007
- [x] Performance targets are concrete numbers, not adjectives
- [x] Open questions called out explicitly
- [x] Sign-off on commit

cc @banidoru for review.


banidoru (commented):
Solid spec overall — the atomic-swap snapshot design is the right call. A few issues to sort before merging:

- **Slot struct offset mismatch (§5.1):** `_pad` is listed at offset 32 but fields before it sum to 30 bytes — either the comment is wrong or 2 bytes are unaccounted for.
- **Raw pointers in stats segment (§6):** Publishing `void*` into shared memory breaks if frontend and backend have different base addresses (they will, since it's mmap'd). Should be offsets from segment base.
- **Robin-Hood + lock-free CAS (§5.2):** Robin-Hood hashing requires displacing existing entries on insert, which is extremely hard to do correctly with CAS alone. Needs explicit discussion of how displacement chains work under concurrent access, or downgrade to linear probing.
- **`vlib_worker_thread_barrier_*` vs. zero-stall claim (§7.2 / §10):** The barrier machinery *does* stall workers. Claiming 0 cycles of worker stall during snapshot while using barriers for the grace period is contradictory — clarify that the barrier is post-swap (for retirement only) and quantify its impact, or use a true RCU epoch counter instead.
- **`flow_def_id` type inconsistency:** §6 says `u128` UUID, but §8 says flow_defs are in a vector "indexed by `flow_def_id`" — you can't index a vector by u128. Clarify the internal vs. external ID scheme.

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,356 @@
...
+- **`infmon-counter`** — Walks the scratch vector and, for each entry, issues
+  one update against the corresponding counter table (§5). This is the only
+  node that mutates shared state.
+
+- **`drop`** — VPP's built-in. We do not free buffers ourselves.
+
+Vector handoff is avoided: the entire pipeline executes on the RX worker
+thread for the queue. This keeps cache lines hot and removes the only
+place where the data path could plausibly block.
+
```
banidoru, line 107: **Offset mismatch.** Fields `key_hash`(8) + `packets`(8) + `bytes`(8) + `key_offset`(4) + `key_len`(2) + `flags`(2) = 32 bytes at offset 0–31. So `_pad` starts at offset 32 and is 32 bytes — total 64 B. But the *inline comments* say `key_offset` is at offset 24, `key_len` at 28, `flags` at 30 — that makes the fields sum to **30 + 2 = 32**… wait, 24+4=28, 28+2=30, 30+2=32. OK the offsets are actually consistent (fields end at byte 32, pad starts at 32). The comments are correct.

Scratch that — the offsets are fine. But there's a real issue: the comment on `_pad` says offset 32 with size 32, totaling 64B, which is correct. **However**, `key_offset` at offset 24 only works if `bytes` (a `u64` at offset 16) is 8 bytes (16+8=24). This is fine for natural alignment on 64-bit, but the struct needs `__attribute__((packed))` or explicit static_assert to guarantee no compiler padding changes the layout. Worth noting in the spec since this is a shared-memory ABI.

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,356 @@
...
+## 6. Stats-segment exposure
+
+The frontend MUST NOT call into the plugin on the hot path to read
+counters. Instead, the plugin publishes its tables through VPP's existing
+**stats segment** (the same shared-memory region that powers `vpp_get_stats`).
+This gives us:
+
+- A read-only mmap from the frontend (no syscalls per read once mapped).
+- A directory mechanism (`/stat_dir`) we register table descriptors under,
```
banidoru, line 179: **Raw `void*` in shared memory is broken.** The frontend mmaps the stats segment at an arbitrary virtual address — it won't match the plugin's VA. These must be **offsets from the stats-segment base**, not raw pointers. VPP's own stats segment entries use offset-based addressing for exactly this reason. Same applies to `key_arena_ptr`.

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,356 @@
...
+
+- A bounded-size Robin-Hood hash table sized at plugin init from a CLI
+  argument (`max_keys_per_flow_def`, default `2^20 = 1,048,576`).
+- Slot layout (cache-line aligned, 64 B):
+
+```
+struct infmon_slot {
+    u64  key_hash;          // 0  full 64-bit hash of key_blob
+    u64  packets;           // 8  atomic, monotonic
+    u64  bytes;             // 16 atomic, monotonic
```
banidoru, line 128: **Robin-Hood + lock-free is a known-hard problem.** Robin-Hood hashing requires displacing existing occupied slots during insert (to maintain the probe-distance invariant). Doing this atomically with CAS on the data path means potentially CAS-swapping *multiple* slots in sequence — that's not atomic as a whole, and a concurrent reader or inserter can observe a partially-displaced chain.

Options:
1. Downgrade to plain linear probing (simpler, still cache-friendly, well-understood lock-free behavior).
2. Specify how displacement is made safe — e.g., per-bucket-group locks (but then it's not really lock-free on insert).
3. Acknowledge this as an open question for the impl PR.

As written, the combination of "Robin-Hood" + "CAS on insert" + "seqlock only for occupancy transitions" doesn't add up.

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,356 @@
...
+1. Caller invokes `infmon_snapshot_and_clear(flow_def_id)`.
+2. Plugin allocates a new, empty counter table (`generation = G+1`) for
+   this `flow_def_id`. Allocation happens off the worker thread; the
+   table is zeroed by a control thread before installation.
+3. Plugin atomically swaps the table pointer published in the
+   `infmon-counter` node's per-flow_def context. The swap is a single
+   `__atomic_store_n(..., RELEASE)` on a pointer — no worker thread
+   stalls, no lock is taken, no packet is dropped.
+4. From the next packet onward, the worker counts into the new table
+   (`G+1`).
```
banidoru, line 258: **Barrier ≠ zero stall.** `vlib_worker_thread_barrier_*` works by making every worker hit a barrier point — each worker *does* stall briefly until all workers have reached the barrier. §10 promises "0 cycles of worker stall during snapshot" but this section uses barriers for the grace period.

If the barrier is only used for *retirement* (freeing the old table after the swap), that's post-snapshot and arguably doesn't count — but the spec should say so explicitly. Better yet, consider a polling-based epoch counter (each worker bumps an epoch on every dispatch loop; control thread waits until all workers have advanced past the swap epoch) to avoid any barrier at all.

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,356 @@
...
+### 5.3 Width
+
+64-bit counters are mandatory. At 100 Gbps line-rate of 64-byte frames
+(~148.8 Mpps) and even with all packets hitting one row, a 32-bit packet
+counter would wrap in ≈29 s, which is shorter than any realistic export
+interval. 64-bit gives ~3.9k years of headroom on the same workload.
+
+## 6. Stats-segment exposure
+
+The frontend MUST NOT call into the plugin on the hot path to read
```
banidoru, line 173: **`flow_def_id` is `u128` here but used as a vector index in §8.** §8 says flow_defs are stored in "an RCU-protected vector indexed by `flow_def_id`". You can't index a C array by a 128-bit UUID. Presumably there's an internal integer handle (u32?) that maps to the UUID? If so, spec it — the match node's per-packet cost depends heavily on whether it's doing a hash lookup or a simple array index to find each flow_def.

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,356 @@
...
+  RCU-protected vector indexed by `flow_def_id`. Workers acquire a
+  pointer to the current vector once per frame (not per packet) and use
+  it for the whole batch. Definition updates publish a new vector and
+  retire the old one through the same grace-period machinery as table
+  retirement.
+- We require RX-queue → worker pinning (set via VPP's standard
+  `dpdk { dev … { workers <list> } }` config). Without pinning the
+  cache-line story in §5 collapses; the plugin will refuse to start if
+  any input interface lands on the main thread.
+
```
banidoru, line 322: **Scratch vector sizing may blow the stack.** `VLIB_FRAME_SIZE × max_active_flow_defs` — with 256 packets and, say, 64 flow_defs, each entry being `(u32 flow_def_id, u64 key_hash, ptr key_blob)` ≈ 20+ bytes, that's ~320 KB. §9 says this lives in TLS, not stack — good — but worth adding a hard cap or noting the expected max `flow_def` count for v1 to bound memory.

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,356 @@
...
+```
+struct infmon_slot {
+    u64  key_hash;          // 0  full 64-bit hash of key_blob
+    u64  packets;           // 8  atomic, monotonic
+    u64  bytes;             // 16 atomic, monotonic
+    u32  key_offset;        // 24 offset into per-table key arena
+    u16  key_len;           // 28
+    u16  flags;             // 30  occupied / tombstone / overflow
+    u8   _pad[32];          // 32  pad to 64 B
```
banidoru, line 132: **Seqlock + relaxed atomics on counters is contradictory for snapshot correctness.** §5.2 says the seqlock protects occupancy transitions so snapshot readers can detect torn reads of `(key_hash, key_offset, key_len)`. But `packets` and `bytes` use relaxed atomics with no fence. A reader on another core could see a stale `packets`/`bytes` even after observing the seqlock's release store — relaxed gives no cross-core visibility guarantee. For the *live* table (which the spec says uses seqlock retry), the reader needs at least `__atomic_load_n(..., ACQUIRE)` on the counters, paired with the seqlock's release, to get a consistent snapshot. Otherwise you're reading counters that haven't propagated from the writer core's store buffer. The *retired* table is fine (immutable + grace period acts as a barrier), but the live-table seqlock protocol needs the counter loads to be acquire, not relaxed.

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,356 @@
...
+  so the frontend discovers tables by enumeration instead of by hard-coded
+  paths.
+
+Per-table descriptor layout (registered under
+`/infmon/<flow_def_id>/<generation>`):
+
+| Field                  | Type     | Notes                                       |
+| ---------------------- | -------- | ------------------------------------------- |
+| `flow_def_id`          | `u128`   | UUID of the flow_def.                       |
+| `generation`           | `u64`    | Bumped on every snapshot_and_clear (§7).    |
```
banidoru, line 194: **`key_arena_len` in the descriptor is ambiguous.** Is this the *capacity* of the arena or the *used* bytes? The frontend needs to know the high-water mark to iterate keys, but if this field is the capacity, it'll walk garbage past the bump allocator's head. Specify whether this is `capacity` or `used`, and if capacity, add a `key_arena_used` field.

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,356 @@
...
+5. The retired table (`G`) remains live in the stats segment under its
+   old directory entry. The reply to `snapshot_and_clear` contains the
+   descriptor of `G`; the caller may walk it at leisure.
+6. The retired table is freed by a control-thread RCU-style grace period:
+   we wait until every worker has executed at least one node dispatch
+   (using VPP's existing `vlib_worker_thread_barrier_*` machinery, **not**
+   on the data path) plus an additional grace window of
+   `INFMON_RETIRE_GRACE_NS` (default 5 s) to give in-flight readers time
+   to finish, then unregister the directory entry and free the slot
+   array + key arena.
```
banidoru: **RCU grace period for `flow_def` vector retirement is underspecified.** The spec says flow_def definitions use "the same grace-period machinery as table retirement" but §7.2 step 6 describes that machinery as `vlib_worker_thread_barrier_*` + a 5s grace window. The barrier call *does* stall workers (see the other reviewer's comment on line 235). If flow_def updates also use barriers, then adding/removing a flow_def during production traffic causes a worker stall — contradicting the §1 design rule. Clarify: does flow_def update use the barrier, or a true lockless RCU (e.g., `urcu` / manual epoch)? If the former, this needs a callout that flow_def mutations are admin-rate operations and the stall is accepted.

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,356 @@
...
+
+The crucial properties:
+
+- **No reset bit, no zeroing of live counters.** The new table starts at
+  zero because it is freshly allocated, not because anyone wrote zeros to
+  shared memory. This eliminates the classic "counter went backwards"
+  race seen in IPFIX implementations.
+- **The backend keeps counting on the new table from the very next
+  packet.** No export interval ever loses a packet to a snapshot.
```
banidoru, line 284: **No memory ordering specified for the per-frame flow_def pointer acquisition.** §8 says "workers acquire a pointer to the current vector once per frame" but doesn't specify the load order. If the flow_def vector pointer is published with `RELEASE`, workers need at least `ACQUIRE` to see the contents of the new vector. With `RELAXED`, a worker could read the new pointer but see stale flow_def entries. This is a classic publish pattern — please specify the memory order for both the publish and the per-frame load.

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,356 @@
...
+| `infmon_status`                  | client → plugin  | Per-worker counters (packets seen, drops by reason, table fullness). |
+
+The wire schema (`*.api` file) lives in
+`infmon-backend/api/infmon.api` and is consumed by `vppapigen` to produce
+both C headers and Rust bindings during build.
+
+### 7.2 `snapshot_and_clear` semantics
+
+This is the export primitive. Frontends call it once per export interval
+(spec 005 sets the cadence; the backend is indifferent).
```
banidoru, line 239: **Atomic pointer swap (RELEASE) in §7.2 step 3 doesn't guarantee visibility to workers on other cores within bounded time.** The spec says "from the next packet onward, the worker counts into the new table" but a RELEASE store is only guaranteed to be visible to another thread that does an ACQUIRE load of the same location. If the worker reads the table pointer with RELAXED (not specified), it could keep counting into the old table for an unbounded number of packets. Specify that workers load the table pointer with ACQUIRE, or document the acceptable staleness window.

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,356 @@
...
+  guesses.
+
+## 10. Performance targets
+
+These are the numbers the implementation must hit on a single
+BlueField-3 ARM core (Cortex-A78AE @ 2.75 GHz, 64 B cache lines, DPDK
+25.11, VPP 24.10) with one RX queue, one `flow_def`, and a key set
+small enough to fit in L2 (~1024 keys):
+
+| Workload                                  | Target               | Stretch              |
```
banidoru, line 351: **`snapshot_alloc_failed` has no documented recovery path.** If the replacement table can't be allocated (OOM in stats segment), the `snapshot_and_clear` call fails — but what happens next? Does the caller retry? Does the old table keep accumulating? Can counters wrap during a prolonged allocation failure? The spec should state the invariant: e.g., "on alloc failure, the existing table remains live and continues accumulating; the caller should retry after freeing retired tables or increasing statseg size."

banidoru (commented):
- Memory ordering gaps throughout: relaxed atomics on counters break live-table seqlock correctness; table pointer swap and flow_def pointer acquisition both lack specified load orderings
- RCU/grace-period machinery for flow_def retirement reuses barrier calls that contradict the "never block workers" design rule — needs clarification
- Stats descriptor has ambiguous `key_arena_len` (capacity vs used?) and missing `snapshot_alloc_failed` recovery semantics
- Overall design is strong — atomic swap for snapshots is the right call, but the atomics/ordering story needs tightening before implementation

banidoru (approved):
All reviewers approved. LGTM.

r12f:

specs/004-backend-architecture.md
r12f: Thanks for the detailed review, but this comment is anchored to `specs/004-backend-architecture.md` and PR #4 does not modify (or contain) that file — PR #4 only adds `specs/001-ci-and-precommit.md` (CI / pre-commit contract). The substantive concerns you raise (shared-memory ABI offsets, pointer-vs-offset addressing, Robin-Hood + lock-free interaction, barrier semantics, memory ordering on table/flow_def publish, seqlock + relaxed counter ordering, key_arena_len semantics, snapshot_alloc_failed recovery, etc.) belong on the spec-004 PR, not on the CI spec. Resolving here as anchor-mismatch; please re-file against the spec-004 PR so the discussion lives next to the relevant text. (Automated reply from triage agent.)

specs/004-backend-architecture.md
r12f: Thanks for the detailed review, but this comment is anchored to `specs/004-backend-architecture.md` and PR #4 does not modify (or contain) that file — PR #4 only adds `specs/001-ci-and-precommit.md` (CI / pre-commit contract). The substantive concerns you raise (shared-memory ABI offsets, pointer-vs-offset addressing, Robin-Hood + lock-free interaction, barrier semantics, memory ordering on table/flow_def publish, seqlock + relaxed counter ordering, key_arena_len semantics, snapshot_alloc_failed recovery, etc.) belong on the spec-004 PR, not on the CI spec. Resolving here as anchor-mismatch; please re-file against the spec-004 PR so the discussion lives next to the relevant text. (Automated reply from triage agent.)

specs/004-backend-architecture.md
r12f: Thanks for the detailed review, but this comment is anchored to `specs/004-backend-architecture.md` and PR #4 does not modify (or contain) that file — PR #4 only adds `specs/001-ci-and-precommit.md` (CI / pre-commit contract). The substantive concerns you raise (shared-memory ABI offsets, pointer-vs-offset addressing, Robin-Hood + lock-free interaction, barrier semantics, memory ordering on table/flow_def publish, seqlock + relaxed counter ordering, key_arena_len semantics, snapshot_alloc_failed recovery, etc.) belong on the spec-004 PR, not on the CI spec. Resolving here as anchor-mismatch; please re-file against the spec-004 PR so the discussion lives next to the relevant text. (Automated reply from triage agent.)

specs/004-backend-architecture.md
r12f: Thanks for the detailed review, but this comment is anchored to `specs/004-backend-architecture.md` and PR #4 does not modify (or contain) that file — PR #4 only adds `specs/001-ci-and-precommit.md` (CI / pre-commit contract). The substantive concerns you raise (shared-memory ABI offsets, pointer-vs-offset addressing, Robin-Hood + lock-free interaction, barrier semantics, memory ordering on table/flow_def publish, seqlock + relaxed counter ordering, key_arena_len semantics, snapshot_alloc_failed recovery, etc.) belong on the spec-004 PR, not on the CI spec. Resolving here as anchor-mismatch; please re-file against the spec-004 PR so the discussion lives next to the relevant text. (Automated reply from triage agent.)

specs/004-backend-architecture.md
r12f: Thanks for the detailed review, but this comment is anchored to `specs/004-backend-architecture.md` and PR #4 does not modify (or contain) that file — PR #4 only adds `specs/001-ci-and-precommit.md` (CI / pre-commit contract). The substantive concerns you raise (shared-memory ABI offsets, pointer-vs-offset addressing, Robin-Hood + lock-free interaction, barrier semantics, memory ordering on table/flow_def publish, seqlock + relaxed counter ordering, key_arena_len semantics, snapshot_alloc_failed recovery, etc.) belong on the spec-004 PR, not on the CI spec. Resolving here as anchor-mismatch; please re-file against the spec-004 PR so the discussion lives next to the relevant text. (Automated reply from triage agent.)

specs/004-backend-architecture.md
r12f: Thanks for the detailed review, but this comment is anchored to `specs/004-backend-architecture.md` and PR #4 does not modify (or contain) that file — PR #4 only adds `specs/001-ci-and-precommit.md` (CI / pre-commit contract). The substantive concerns you raise (shared-memory ABI offsets, pointer-vs-offset addressing, Robin-Hood + lock-free interaction, barrier semantics, memory ordering on table/flow_def publish, seqlock + relaxed counter ordering, key_arena_len semantics, snapshot_alloc_failed recovery, etc.) belong on the spec-004 PR, not on the CI spec. Resolving here as anchor-mismatch; please re-file against the spec-004 PR so the discussion lives next to the relevant text. (Automated reply from triage agent.)

specs/004-backend-architecture.md
r12f: Thanks for the detailed review, but this comment is anchored to `specs/004-backend-architecture.md` and PR #4 does not modify (or contain) that file — PR #4 only adds `specs/001-ci-and-precommit.md` (CI / pre-commit contract). The substantive concerns you raise (shared-memory ABI offsets, pointer-vs-offset addressing, Robin-Hood + lock-free interaction, barrier semantics, memory ordering on table/flow_def publish, seqlock + relaxed counter ordering, key_arena_len semantics, snapshot_alloc_failed recovery, etc.) belong on the spec-004 PR, not on the CI spec. Resolving here as anchor-mismatch; please re-file against the spec-004 PR so the discussion lives next to the relevant text. (Automated reply from triage agent.)

specs/004-backend-architecture.md
r12f: Thanks for the detailed review, but this comment is anchored to `specs/004-backend-architecture.md` and PR #4 does not modify (or contain) that file — PR #4 only adds `specs/001-ci-and-precommit.md` (CI / pre-commit contract). The substantive concerns you raise (shared-memory ABI offsets, pointer-vs-offset addressing, Robin-Hood + lock-free interaction, barrier semantics, memory ordering on table/flow_def publish, seqlock + relaxed counter ordering, key_arena_len semantics, snapshot_alloc_failed recovery, etc.) belong on the spec-004 PR, not on the CI spec. Resolving here as anchor-mismatch; please re-file against the spec-004 PR so the discussion lives next to the relevant text. (Automated reply from triage agent.)

specs/004-backend-architecture.md
r12f: Thanks for the detailed review, but this comment is anchored to `specs/004-backend-architecture.md` and PR #4 does not modify (or contain) that file — PR #4 only adds `specs/001-ci-and-precommit.md` (CI / pre-commit contract). The substantive concerns you raise (shared-memory ABI offsets, pointer-vs-offset addressing, Robin-Hood + lock-free interaction, barrier semantics, memory ordering on table/flow_def publish, seqlock + relaxed counter ordering, key_arena_len semantics, snapshot_alloc_failed recovery, etc.) belong on the spec-004 PR, not on the CI spec. Resolving here as anchor-mismatch; please re-file against the spec-004 PR so the discussion lives next to the relevant text. (Automated reply from triage agent.)

specs/004-backend-architecture.md
r12f: Thanks for the detailed review, but this comment is anchored to `specs/004-backend-architecture.md` and PR #4 does not modify (or contain) that file — PR #4 only adds `specs/001-ci-and-precommit.md` (CI / pre-commit contract). The substantive concerns you raise (shared-memory ABI offsets, pointer-vs-offset addressing, Robin-Hood + lock-free interaction, barrier semantics, memory ordering on table/flow_def publish, seqlock + relaxed counter ordering, key_arena_len semantics, snapshot_alloc_failed recovery, etc.) belong on the spec-004 PR, not on the CI spec. Resolving here as anchor-mismatch; please re-file against the spec-004 PR so the discussion lives next to the relevant text. (Automated reply from triage agent.)

specs/004-backend-architecture.md
r12f: Thanks for the detailed review, but this comment is anchored to `specs/004-backend-architecture.md` and PR #4 does not modify (or contain) that file — PR #4 only adds `specs/001-ci-and-precommit.md` (CI / pre-commit contract). The substantive concerns you raise (shared-memory ABI offsets, pointer-vs-offset addressing, Robin-Hood + lock-free interaction, barrier semantics, memory ordering on table/flow_def publish, seqlock + relaxed counter ordering, key_arena_len semantics, snapshot_alloc_failed recovery, etc.) belong on the spec-004 PR, not on the CI spec. Resolving here as anchor-mismatch; please re-file against the spec-004 PR so the discussion lives next to the relevant text. (Automated reply from triage agent.)

specs/004-backend-architecture.md
r12f: Thanks for the detailed review, but this comment is anchored to `specs/004-backend-architecture.md` and PR #4 does not modify (or contain) that file — PR #4 only adds `specs/001-ci-and-precommit.md` (CI / pre-commit contract). The substantive concerns you raise (shared-memory ABI offsets, pointer-vs-offset addressing, Robin-Hood + lock-free interaction, barrier semantics, memory ordering on table/flow_def publish, seqlock + relaxed counter ordering, key_arena_len semantics, snapshot_alloc_failed recovery, etc.) belong on the spec-004 PR, not on the CI spec. Resolving here as anchor-mismatch; please re-file against the spec-004 PR so the discussion lives next to the relevant text. (Automated reply from triage agent.)

specs/004-backend-architecture.md
r12f: Apologies — my previous reply and resolve on this thread were posted in error by an automated triage agent that mis-identified the host PR. This thread is on PR #7 (spec-004), not PR #4 (spec-001 / CI). Your original review comment stands and is substantive; I am unresolving so it can be properly addressed on the correct PR. Sorry for the noise.

specs/004-backend-architecture.md
r12f: Apologies — my previous reply and resolve on this thread were posted in error by an automated triage agent that mis-identified the host PR. This thread is on PR #7 (spec-004), not PR #4 (spec-001 / CI). Your original review comment stands and is substantive; I am unresolving so it can be properly addressed on the correct PR. Sorry for the noise.

specs/004-backend-architecture.md
r12f: Apologies — my previous reply and resolve on this thread were posted in error by an automated triage agent that mis-identified the host PR. This thread is on PR #7 (spec-004), not PR #4 (spec-001 / CI). Your original review comment stands and is substantive; I am unresolving so it can be properly addressed on the correct PR. Sorry for the noise.

specs/004-backend-architecture.md
r12f: Apologies — my previous reply and resolve on this thread were posted in error by an automated triage agent that mis-identified the host PR. This thread is on PR #7 (spec-004), not PR #4 (spec-001 / CI). Your original review comment stands and is substantive; I am unresolving so it can be properly addressed on the correct PR. Sorry for the noise.

specs/004-backend-architecture.md
r12f: Apologies — my previous reply and resolve on this thread were posted in error by an automated triage agent that mis-identified the host PR. This thread is on PR #7 (spec-004), not PR #4 (spec-001 / CI). Your original review comment stands and is substantive; I am unresolving so it can be properly addressed on the correct PR. Sorry for the noise.

specs/004-backend-architecture.md
r12f: Apologies — my previous reply and resolve on this thread were posted in error by an automated triage agent that mis-identified the host PR. This thread is on PR #7 (spec-004), not PR #4 (spec-001 / CI). Your original review comment stands and is substantive; I am unresolving so it can be properly addressed on the correct PR. Sorry for the noise.

specs/004-backend-architecture.md
r12f: Apologies — my previous reply and resolve on this thread were posted in error by an automated triage agent that mis-identified the host PR. This thread is on PR #7 (spec-004), not PR #4 (spec-001 / CI). Your original review comment stands and is substantive; I am unresolving so it can be properly addressed on the correct PR. Sorry for the noise.

specs/004-backend-architecture.md
r12f: Apologies — my previous reply and resolve on this thread were posted in error by an automated triage agent that mis-identified the host PR. This thread is on PR #7 (spec-004), not PR #4 (spec-001 / CI). Your original review comment stands and is substantive; I am unresolving so it can be properly addressed on the correct PR. Sorry for the noise.

specs/004-backend-architecture.md
r12f: Apologies — my previous reply and resolve on this thread were posted in error by an automated triage agent that mis-identified the host PR. This thread is on PR #7 (spec-004), not PR #4 (spec-001 / CI). Your original review comment stands and is substantive; I am unresolving so it can be properly addressed on the correct PR. Sorry for the noise.

specs/004-backend-architecture.md
r12f: Apologies — my previous reply and resolve on this thread were posted in error by an automated triage agent that mis-identified the host PR. This thread is on PR #7 (spec-004), not PR #4 (spec-001 / CI). Your original review comment stands and is substantive; I am unresolving so it can be properly addressed on the correct PR. Sorry for the noise.

specs/004-backend-architecture.md
r12f: Apologies — my previous reply and resolve on this thread were posted in error by an automated triage agent that mis-identified the host PR. This thread is on PR #7 (spec-004), not PR #4 (spec-001 / CI). Your original review comment stands and is substantive; I am unresolving so it can be properly addressed on the correct PR. Sorry for the noise.

specs/004-backend-architecture.md
r12f: Apologies — my previous reply and resolve on this thread were posted in error by an automated triage agent that mis-identified the host PR. This thread is on PR #7 (spec-004), not PR #4 (spec-001 / CI). Your original review comment stands and is substantive; I am unresolving so it can be properly addressed on the correct PR. Sorry for the noise.

specs/004-backend-architecture.md
r12f: Good catch — added `__attribute__((packed, aligned(64)))` plus a commented `static_assert` to lock the 64 B layout as a shared-memory ABI invariant. (v0.3)

specs/004-backend-architecture.md
r12f: Agreed, raw `void*` across processes is broken. Switched both `slots_ptr` and `key_arena_ptr` to byte offsets from the stats-segment base (`slots_offset`, `key_arena_offset`), matching VPP's stats-segment convention. Documented in §6. (v0.3)

specs/004-backend-architecture.md
r12f: Fair — went with option 1. Downgraded to plain linear probing in §5.1 with a brief note explaining why Robin-Hood doesn't compose with CAS-based lock-free insert. (v0.3)

specs/004-backend-architecture.md
r12f: Reworked §7.2 step 6 to use a per-worker epoch counter instead of `vlib_worker_thread_barrier_*`: workers bump a thread-local epoch (RELEASE) once per dispatch loop, control thread waits until all workers pass the swap epoch + grace window. The barrier API is now explicitly reserved for plugin teardown only. §10's zero-stall guarantee now holds end-to-end. (v0.3)

specs/004-backend-architecture.md
r12f: Right — added an internal `flow_def_index` (u32) handle. UUID is the external identity; on the data path workers index the flow_def vector by the integer handle, so per-packet dispatch is an O(1) array index. Added the field to the stats descriptor (§6) and clarified §8. (v0.3)

specs/004-backend-architecture.md
r12f: Added a hard `static_assert` cap of 64 on `max_active_flow_defs` for v1, with the worked-out math (~384 KiB per worker) showing it stays well inside the default 8 MiB TLS budget. Lifting the cap is called out as requiring either smaller entries or a heap-backed scratch. (v0.3)

specs/004-backend-architecture.md
r12f: You're right — relaxed loads on `packets`/`bytes` don't compose with the seqlock for the live-table snapshot path. Updated §5.2: data-path stores stay relaxed, but live-table snapshot readers must use `__atomic_load_n(..., ACQUIRE)` paired with the seqlock's release on the slot metadata. Retired tables are noted as immutable + grace-period-fenced, so plain loads remain fine there. (v0.3)

specs/004-backend-architecture.md
r12f: Ambiguity removed: split the field into `key_arena_capacity` (allocation size) and `key_arena_used` (bump-allocator high-water mark). The descriptor table in §6 spells out that frontends MUST iterate keys only up to `key_arena_used`. (v0.3)

specs/004-backend-architecture.md
r12f: Clarified in §8: flow_def vector retirement uses the same epoch-counter RCU machinery as table retirement (§7.2 step 6) — explicitly **not** `vlib_worker_thread_barrier_*`. So flow_def add/remove is genuinely stall-free. The barrier comment in §7.2 was the actual problem and has been replaced; both paths are now consistent with §10. (v0.3)

specs/004-backend-architecture.md
r12f: Specified in §8: control thread publishes the new vector pointer with `__atomic_store_n(..., RELEASE)`; workers read it with `__atomic_load_n(..., ACQUIRE)` once per frame. Release/acquire pair guarantees the vector contents are visible when the new pointer is observed. (v0.3)

specs/004-backend-architecture.md
r12f: Specified in §7.2 step 3: workers MUST load the table pointer with `__atomic_load_n(..., ACQUIRE)` (once per frame, see §8). The bounded-staleness window is exactly one frame (≤ VLIB_FRAME_SIZE packets), after which every subsequent packet counts into G+1. (v0.3)

specs/004-backend-architecture.md
r12f: Added a recovery contract paragraph in §11: on `snapshot_alloc_failed`, the live table stays installed and keeps accumulating in generation G — no swap, no reset, no worker disruption. The reply returns the error code with the current generation so the caller can distinguish 'no new data' from silent failure. Recovery is wait-for-grace-period or bump `statseg { size }` and retry; 64-bit width gives ~3.9k years of headroom against wrap during outage. (v0.3)

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,419 @@
...
+- The data-plane nodes process whole VPP frames (`VLIB_FRAME_SIZE`,
+  currently 256). All three plugin nodes are written as **dual-loop**
+  nodes (process two packets per inner iteration with software prefetch
+  of `+2` and `+3`), which is the standard VPP pattern.
+- The match-emit scratch vector is sized
+  `VLIB_FRAME_SIZE × max_active_flow_defs` and lives in per-thread TLS;
+  no allocation occurs per frame. To bound TLS footprint, v1 caps
+  `max_active_flow_defs` at **64** (a hard `static_assert` in the
+  plugin); at the v1 entry size of ≈24 B per `(flow_def_id, key_hash,
+  key_blob_ptr)` triple this gives a per-worker scratch of
```
banidoru, line 334: Minor: this says `flow_def_id` but should be `flow_def_index` — the data-path scratch uses the u32 handle, not the u128 UUID. The 24 B estimate only adds up with the 4-byte index (4 + 8 + 8 = 20, ~24 with alignment padding), not the 16-byte UUID (16 + 8 + 8 = 32).

banidoru (approved):
- All 11 prior review concerns addressed in v0.3 — linear probing, offset-based descriptors, memory ordering, epoch-counter RCU, scratch cap, alloc-failed recovery all look correct.
- One minor nit: scratch vector entry description (line ~339) says `flow_def_id` but should say `flow_def_index` — the whole point of the new u32 handle is that the data path never touches the u128 UUID, and the size estimate (24 B) only works with the u32 index, not the 16-byte UUID.
- Epoch-counter RCU replacing barrier-based retirement is a clean improvement; the §10 zero-stall guarantee now holds end-to-end.
- ACQUIRE/RELEASE pairing is correctly specified for all three publish points (table pointer, flow_def vector, seqlock counters).
- LGTM.

r12f:

specs/004-backend-architecture.md
r12f: Good catch — fixed in 9fb3f5d. Switched both the §6 emit-format prose and the §9 scratch-triple to `flow_rule_index` (the `u32` handle), so the 24 B/entry estimate (≈4 + 8 + 8 with alignment) now actually matches what the data path holds. The u128 `flow_rule_id` UUID stays the external identity, resolved to its index by the control plane.

banidoru (approved):
- The spec now correctly uses offset-based addressing for shared memory.
- The spec now correctly specifies memory ordering for concurrent operations.
- The spec now uses a more appropriate epoch-based RCU mechanism.

banidoru (approved):
- All 12 prior review threads addressed and resolved — linear probing, offset-based descriptors, memory ordering, epoch-based RCU, scratch cap, alloc-failed recovery all look solid.
- v0.4 `flow_def` → `flow_rule` rename is consistent throughout; terminology table now correctly distinguishes flow_rule (config) from flow (runtime instance).
- v0.5 correctly switches scratch-triple to `flow_rule_index` (u32) — 24 B/entry math checks out.
- Nit: file is missing a trailing newline (POSIX convention); v0.4/v0.5 version-history rows have slightly ragged column alignment.
- LGTM.

banidoru (approved):
All reviewers approved. LGTM.

banidoru (approved):
- All prior review concerns (12 threads) have been addressed and resolved — packed attribute, offset-based pointers, linear probing, memory ordering, epoch RCU, scratch cap, alloc-failed recovery, flow_def_index handle, scratch triple type fix.
- Rename `flow_def` → `flow_rule` is consistent throughout; no stale references found.
- Version history collapsed to single row per PR and metadata simplified — clean.
- Minor: file ends without trailing newline (last line); cosmetic, not blocking.
- LGTM — approve.

banidoru (commented):
- All prior review threads have been addressed and resolved.

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,414 @@
+# Spec 004 — Backend Architecture (VPP plugin)
+
+## Version history
+
+| Version | Date       | Author       | Changes |
+| ------- | ---------- | ------------ | ------- |
+| 0.1     | 2026-04-18 | Riff (r12f)  | Initial draft of `infmon-backend` (VPP plugin). Linear-probing flow table with offset-based descriptors, memory ordering, epoch-based RCU, scratch cap, alloc-failed recovery; internal identifiers use `flow_rule*` per Spec 002 mental model; per-worker scratch-triple and §6 emit format use `flow_rule_index` (u32 handle), keeping the 24 B/entry estimate. |
```
banidoru, line 7: Nit: collapsing v0.1–v0.3 into a single v0.1 entry loses the review-driven change history. Consider keeping the three rows (or at minimum noting that v0.1 incorporates review feedback) so future readers can trace which decisions were iterated on.

specs/004-backend-architecture.md
```diff
@@ -0,0 +1,414 @@
...
+3. **Per-CPU sharded tables.** A sharded design (one sub-table per
+   worker, merged at snapshot) avoids cross-core CAS entirely but
+   complicates §7.2. Defer to v2 unless the §10 targets miss.
+
+## 13. Acceptance
+
+This spec is accepted (per the project's spec-first process — see spec
+000) when it is merged to `main` with @banidoru's sign-off. After
+acceptance, the implementation PR for `infmon-backend` may begin and
+must conform to §4–§9 or amend this spec first.
```
banidoru: Nit: missing trailing newline at EOF. Most POSIX tools expect it and git will flag `\ No newline at end of file` in diffs.

banidoru (commented):
Re-review (iteration 3) — new commits since last review:

- All 12 prior review threads have been addressed and resolved. No outstanding concerns.
- The `flow_def` → `flow_rule` rename is complete and consistent — zero stale `flow_def` references remain.
- Version history was collapsed from three entries (v0.1–v0.3) into a single v0.1 — loses the change trail, consider preserving it or noting the squash was intentional.
- Missing trailing newline at EOF (POSIX convention, may cause noisy diffs later).
- No new correctness, design, or performance issues introduced by this changeset.

banidoru (approved):
All reviewers approved. LGTM.

specs/004-backend-architecture.md
r12f: Convention: one PR = one version-history row; review-fix changes amend the single row's Changes cell.

specs/004-backend-architecture.md
r12f: Fixed — added trailing newline.

banidoru (approved):
- All 14 prior review threads addressed and resolved — struct packing, offset-based addressing, linear probing downgrade, epoch-based RCU, acquire/release ordering, flow_rule_index fix, and alloc-failure recovery all look good.
- New diff is trivial (trailing newline fix only) — no new issues.
- Spec is internally consistent and ready to merge.

banidoru (approved):
- All 12 prior review threads addressed and resolved — fixes are correct
- New commit (3dca496) only adds trailing newline at EOF, fixing the last nit
- Memory ordering, RCU, linear probing, offset-based descriptors, and alloc-failure recovery all look solid
- No new issues found in the updated spec
- LGTM

banidoru (approved):
All reviewers approved. LGTM.

PR #8, opened 2026-04-18
specs: 007 — CLI (infmon-cli)

r12f:
## Summary

Adds `specs/007-cli.md` per parent epic **DPU-4** and issue **DPU-13**.
Freezes the v1 operator surface for `infmon-cli`.

Covers, per the issue:

- **Verb tree**: `install` / `uninstall`; `start` / `stop` / `restart` / `status`; `config get|set|reload|show`; `flow add|rm|list|show`; `stats show [--name] [--top N]`; `stats export --format prom|otlp|json`; `log tail [-f]`; `health [--json]`.
- **Exit codes**: small stable table (0, 1, 2, 3, 4, 5, 6, 13, 69, 130) with explicit semantics; `health` overload documented.
- **`--output table|json`**: every read verb supports both; JSON schema is the v1 contract; large counters rendered as `*_str` to stay JSON-safe.
- **Config file location**: `/etc/infmon/infmon.toml` (`root:infmon`, `0640`); CLI rewrites atomically (write → fsync → rename); `reload` sends SIGHUP.
- **Sudo requirements**: per-verb privilege table; root for install / lifecycle / config writes / flow mutations; group `infmon` membership for read verbs via the Unix socket at `/run/infmon/frontend.sock` (`SO_PEERCRED` enforced by frontend).
- Plus filesystem contract, global flags, help-text conventions, full test plan, and four open questions with proposed defaults.

Follows the spec template from `specs/000-overview.md`.

## Test plan

- [x] Spec is markdown only — no code changes; CI build/test jobs are unaffected.

## Related

- Epic: DPU-4
- Closes: DPU-13

cc @banidoru for sign-off.

### DIFF
--- a/specs/007-cli.md
+++ b/specs/007-cli.md
@@ -2,12 +2,9 @@
 
 ## Version history
 
-| Version | Date       | Author     | Changes                                  |
-| ------- | ---------- | ---------- | ---------------------------------------- |
-| 0.1     | 2026-04-18 | r12f       | Initial draft of v1 verb tree, output contract, exit codes, config layout, privilege model. |
-| 0.2     | 2026-04-18 | r12f       | Address PR #8 review feedback. |
-| 0.3     | 2026-04-18 | r12f       | Split `flow {add,rm,list,show}` into `flow-rule` (configuration) and `flow` (live read-only views). Each flow-rule generates one flow per distinct key tuple it observes. |
-| 0.4     | 2026-04-18 | r12f       | SIGPIPE: install `exit(0)` handler instead of restoring `SIG_DFL` so the documented "exits 0 silently" claim is actually achieved (avoids the `141` POSIX path). `flow show <rule> <key>` requires `<key>` as a single argv element (must be shell-quoted); reject multi-positional unquoted keys with exit 2. |
+| Version | Date       | Author       | Changes |
+| ------- | ---------- | ------------ | ------- |
+| 0.1     | 2026-04-18 | Riff (r12f)  | Initial draft of `infmon-cli` v1: verb tree, output contract, exit codes, config layout, privilege model. Verb tree splits `flow-rule {add,rm,list,show}` (configuration) from `flow {list,show}` (live read-only views) — each flow-rule generates one flow per distinct key tuple. SIGPIPE installs an `exit(0)` handler so the documented "exits 0 silently" claim holds (avoids the POSIX 141 path). `flow show <rule> <key>` requires `<key>` as a single argv element (must be shell-quoted); multi-positional unquoted keys are rejected with exit 2. |
 
 ---
 
--- a/specs/007-cli.md
+++ b/specs/007-cli.md
@@ -4,10 +4,10 @@
 
 | Version | Date       | Author     | Changes                                  |
 | ------- | ---------- | ---------- | ---------------------------------------- |
-| 0.1     | 2026-04-18 | bf3 (agent)| Initial draft of v1 verb tree, output contract, exit codes, config layout, privilege model. |
-| 0.2     | 2026-04-18 | bf3 (agent)| Address PR #8 review feedback. |
-| 0.3     | 2026-04-18 | bf3 (agent)| Split `flow {add,rm,list,show}` into `flow-rule` (configuration) and `flow` (live read-only views). Each flow-rule generates one flow per distinct key tuple it observes. |
-| 0.4     | 2026-04-18 | bf3 (agent)| SIGPIPE: install `exit(0)` handler instead of restoring `SIG_DFL` so the documented "exits 0 silently" claim is actually achieved (avoids the `141` POSIX path). `flow show <rule> <key>` requires `<key>` as a single argv element (must be shell-quoted); reject multi-positional unquoted keys with exit 2. |
+| 0.1     | 2026-04-18 | r12f       | Initial draft of v1 verb tree, output contract, exit codes, config layout, privilege model. |
+| 0.2     | 2026-04-18 | r12f       | Address PR #8 review feedback. |
+| 0.3     | 2026-04-18 | r12f       | Split `flow {add,rm,list,show}` into `flow-rule` (configuration) and `flow` (live read-only views). Each flow-rule generates one flow per distinct key tuple it observes. |
+| 0.4     | 2026-04-18 | r12f       | SIGPIPE: install `exit(0)` handler instead of restoring `SIG_DFL` so the documented "exits 0 silently" claim is actually achieved (avoids the `141` POSIX path). `flow show <rule> <key>` requires `<key>` as a single argv element (must be shell-quoted); reject multi-positional unquoted keys with exit 2. |
 
 ---
 
--- a/specs/007-cli.md
+++ b/specs/007-cli.md
@@ -1,0 +1,397 @@
+# 007 — CLI (`infmon-cli`)
+
+**Owners:** @banidoru (review), DPU team
+**Status:** Draft
+**Last updated:** 2026-04-18
+
+---
+
+## Context
+
+Operators need a single, scriptable entry point for installing, running,
+configuring, observing, and debugging InFMon on a BlueField-3 DPU. The
+existing components — the VPP plugin (`infmon-backend`) and the user-space
+service (`infmon-frontend`) — expose APIs and on-disk state, but operators
+should not have to poke at systemd units, raw config files, or the frontend's
+local socket directly.
+
+`infmon-cli` is that entry point. It is a single Rust binary, distributed in
+the same `.deb` as the backend and frontend, that wraps every supported
+operator workflow. It must be **predictable in scripts** (stable exit codes,
+machine-readable output) and **comfortable for humans** (tables by default,
+sensible verbs).
+
+This spec freezes the v1 verb tree, output contract, exit-code table, config
+file location, and privilege model. Implementation work tracks against it.
+
+## Goals & Non-goals
+
+- Goal: One CLI binary that owns the full operator surface — install,
+  lifecycle, config, flow inspection, stats, log tail, health.
+- Goal: Stable, documented exit codes so CI and monitoring scripts can rely
+  on them.
+- Goal: `--output table|json` on every read command; JSON is the contract for
+  automation.
+- Goal: Clearly state what requires `sudo` (privileged actions) versus what
+  any local user can run (read-only queries).
+- Goal: Talks to the frontend over its local Unix-domain socket; never
+  reaches into the backend's shared memory directly.
+- Non-goal: Remote administration. v1 is local-only (the CLI runs on the
+  same DPU as the frontend). Remote use happens via SSH.
+- Non-goal: A TUI / curses dashboard. `stats show` and `log tail -f` are the
+  closest we get.
+- Non-goal: Editing flow records or rewriting backend state. The CLI reads
+  state and (re)writes config; it does not mutate live counters.
+- Non-goal: Cluster / multi-DPU operations (out of scope per spec 000).
+
+## Design
+
+### Binary, transport, and discovery
+
+- Single binary `/usr/bin/infmon-cli`, shipped by `infmon_<ver>_arm64.deb`.
+- Talks to `infmon-frontend` over a Unix-domain socket at
+  `/run/infmon/frontend.sock` (created mode `0660`, group `infmon`).
+- Members of the `infmon` group can run all read-only verbs without `sudo`.
+  Privileged verbs (install, lifecycle, config writes) require either root
+  or `sudo`; the CLI prints a clear error and exits `13` (see exit-code
+  table) when invoked without sufficient privilege.
+- Config file lives at `/etc/infmon/infmon.toml` (owned `root:infmon`,
+  mode `0640`). The CLI never edits it in place; `config set` rewrites it
+  atomically (`write → fsync → rename`).
+- Run-state and PID files live under `/run/infmon/`. Logs go to journald via
+  the systemd unit; `log tail` wraps `journalctl -u infmon-frontend`.
+
+### Global flags
+
+Available on every subcommand:
+
+| Flag                    | Default       | Meaning                                      |
+|-------------------------|---------------|----------------------------------------------|
+| `--output <table\|json>`| `table`       | Output format. `json` is line-delimited per record where applicable. |
+| `--config <path>`       | `/etc/infmon/infmon.toml` | Override config file location.   |
+| `--socket <path>`       | `/run/infmon/frontend.sock` | Override frontend socket path. |
+| `--timeout <secs>`      | `5`           | Per-call timeout when contacting the frontend. |
+| `--no-color`            | off           | Disable ANSI colors (auto-off when stdout is not a TTY). |
+| `-q, --quiet`           | off           | Suppress informational stderr; errors still printed. |
+| `-v, --verbose`         | off (repeatable) | `-v` info, `-vv` debug, `-vvv` trace.    |
+| `-h, --help`            | —             | Show help for the (sub)command.              |
+| `-V, --version`         | —             | Print `infmon-cli <semver>` and exit `0`.    |
+
+Conventions:
+
+- All `--output json` responses are valid JSON. List commands emit a single
+  JSON array (not NDJSON) unless explicitly noted (`log tail` is NDJSON).
+- All numeric values in JSON are unquoted numbers; counters that may exceed
+  `2^53` are rendered as decimal strings with a trailing `_str` field name
+  (e.g. `bytes_str`).
+- Timestamps are RFC 3339 UTC strings.
+
+### Verb tree
+
+```
+infmon-cli
+├── install                              [root]
+├── uninstall                            [root]
+├── start                                [root]
+├── stop                                 [root]
+├── restart                              [root]
+├── status                               [any]
+├── config
+│   ├── get <key>                        [any]
+│   ├── set <key> <value>                [root]
+│   ├── reload                           [root]
+│   └── show                             [any]
+├── flow
+│   ├── add <spec>                       [root]
+│   ├── rm  <id|spec>                    [root]
+│   ├── list                             [any]
+│   └── show <id>                        [any]
+├── stats
+│   ├── show [--name <metric>] [--top N] [any]
+│   └── export --format <prom|otlp|json> [any]
+├── log
+│   └── tail [-f] [--since <when>] [-n N][any]
+└── health [--json]                      [any]
+```
+
+`[root]` = requires root or `sudo`. `[any]` = any user, subject to socket
+group membership for verbs that contact the frontend.
+
+#### `install` / `uninstall`
+
+`install` is a one-shot bootstrap that the `.deb` post-install hook also
+calls, but operators may rerun it after manual changes:
+
+- Verifies VPP is installed and the `infmon-backend` plugin is in VPP's
+  plugin directory.
+- Ensures the `infmon` group exists and `/etc/infmon/`, `/run/infmon/`,
+  `/var/lib/infmon/` exist with correct ownership and modes.
+- Writes a default `infmon.toml` if none exists (never overwrites an
+  existing one; prints a notice and exits `0`).
+- Enables the `infmon-frontend.service` systemd unit (does not start it;
+  use `start`).
+- Idempotent: rerunning on a healthy install is a no-op and exits `0`.
+
+`uninstall` is the inverse, but conservative:
+
+- Stops and disables `infmon-frontend.service`.
+- Removes the systemd unit file and the `/usr/bin/infmon-cli`,
+  `/usr/bin/infmon-frontend`, and VPP plugin `.so` only if they were
+  installed by InFMon (checked via dpkg ownership).
+- **Preserves** `/etc/infmon/` and `/var/lib/infmon/` by default; pass
+  `--purge` to remove them.
+
+#### `start` / `stop` / `restart` / `status`
+
+Thin, opinionated wrappers over systemd:
+
+- `start`   → `systemctl start infmon-frontend.service`
+- `stop`    → `systemctl stop  infmon-frontend.service`
+- `restart` → `systemctl restart infmon-frontend.service`
+- `status`  → composite view: systemd unit state, frontend PID, socket
+  reachable yes/no, backend plugin loaded yes/no, uptime, version.
+
+`status` is read-only and works without root. With `--output json` it emits
+an object:
+
+```json
+{
+  "service":    "active",
+  "pid":        1234,
+  "uptime_sec": 86400,
+  "socket":     "/run/infmon/frontend.sock",
+  "socket_ok":  true,
+  "backend": {
+    "plugin_loaded": true,
+    "version":       "1.0.0"
+  },
+  "frontend": {
+    "version": "1.0.0"
+  }
+}
+```
+
+#### `config get|set|reload|show`
+
+- `config show` prints the effective merged config (defaults + file +
+  environment overrides) as TOML (table mode) or JSON (json mode).
+- `config get <key>` prints just that key's value. Dotted keys index into
+  tables: `config get exporter.ipfix.collector`.
+- `config set <key> <value>` rewrites `/etc/infmon/infmon.toml` atomically
+  with the new value. Type is inferred from the existing value's type
+  (string, int, bool, float, array). `--type <t>` can force the type when
+  setting a previously absent key. Requires root.
+- `config reload` sends `SIGHUP` to the frontend (and asks it to revalidate
+  hot-reloadable knobs). Knobs that require restart are listed by `config
+  show --annotate`. Requires root.
+
+Validation: `config set` and `config reload` refuse a config that fails
+schema validation, exit `2`, and leave the on-disk file unchanged.
+
+#### `flow add|rm|list|show`
+
+"Flows" here are *flow filter rules* maintained by the frontend (e.g. an
+allow-list of which flows to export, or named groups for stats roll-ups).
+They are **not** live flow records — those are read-only and surfaced
+through `stats show`.
+
+- `flow add <spec>` adds a rule. Spec uses a compact key=value form, e.g.
+  `proto=tcp src=10.0.0.0/8 dst=0.0.0.0/0 dport=443 name=https-egress`.
+  Returns the assigned rule ID on stdout. Requires root.
+- `flow rm <id|spec>` removes by ID, or by exact-match spec. Requires root.
+- `flow list` lists all rules. Columns: `ID`, `NAME`, `PROTO`, `SRC`,
+  `DST`, `SPORT`, `DPORT`, `ACTION`. JSON mode emits an array of rule
+  objects.
+- `flow show <id>` prints one rule's full detail (including hit counters).
+
+#### `stats show` and `stats export`
+
+- `stats show` prints aggregate counters: `packets`, `bytes`, `flows_active`,
+  `flows_total`, `flows_evicted`, `parse_errors`, `export_records`, etc.
+  - `--name <metric>` restricts output to one metric (also accepts a
+    glob, e.g. `--name 'export_*'`).
+  - `--top N` for top-N talkers / top-N flows by bytes (default `N=20`).
+    Combined with `--name flows.top_bytes` etc.
+  - Default refresh is a single snapshot. Add `--watch <secs>` for periodic
+    refresh (table mode only; json mode prints one snapshot then exits).
+- `stats export --format <prom|otlp|json>` writes the current snapshot to
+  stdout in the requested format:
+  - `prom` — Prometheus text exposition (text/plain; version=0.0.4).
+  - `otlp` — OpenTelemetry metrics, JSON-encoded OTLP.
+  - `json` — InFMon-native JSON (same shape as `stats show --output json`,
+    but always the full set, ignoring `--name`/`--top`).
+  This is a one-shot dump suitable for piping into a collector or a file;
+  it does **not** start a long-lived exporter (those are configured in
+  `infmon.toml` and run inside the frontend).
+
+#### `log tail`
+
+- Wraps `journalctl -u infmon-frontend.service`.
+- `-f` / `--follow` streams new lines (NDJSON when `--output json`).
+- `--since <when>` accepts the same forms as journalctl (`"10 min ago"`,
+  RFC 3339, etc.).
+- `-n N` shows the last `N` lines (default `100`).
+- Exits `0` on clean EOF, `130` on `SIGINT` (Ctrl-C), per the exit table.
+
+#### `health [--json]`
+
+A single composite probe, designed to be the target of an external
+monitor or a Kubernetes-style liveness/readiness check.
+
+- Exits `0` when the frontend socket is reachable, the backend plugin
+  reports `loaded`, and the last snapshot timestamp is within
+  `2 × snapshot_interval`.
+- Exits `1` when degraded (e.g. snapshot is stale but service is up).
+- Exits `4` when the frontend is unreachable.
+- `--json` (alias of `--output json`) prints a structured payload:
+
+```json
+{
+  "ok":              true,
+  "service":         "active",
+  "socket_ok":       true,
+  "backend_loaded":  true,
+  "snapshot_age_ms": 873,
+  "checks": [
+    {"name": "service",        "ok": true},
+    {"name": "socket",         "ok": true},
+    {"name": "backend_loaded", "ok": true},
+    {"name": "snapshot_fresh", "ok": true}
+  ]
+}
+```
+
+### Exit codes
+
+The CLI uses a small, stable set of exit codes. Subcommand-specific codes
+are documented in their help text but always drawn from this table.
+
+| Code | Meaning                                                        |
+|------|----------------------------------------------------------------|
+| `0`  | Success.                                                       |
+| `1`  | Generic / unspecified failure (last resort).                   |
+| `2`  | Usage error: bad flags, bad arguments, schema validation fail. |
+| `3`  | Not found: requested flow ID, config key, or metric is absent. |
+| `4`  | Frontend unreachable (socket missing, connect refused, timeout).|
+| `5`  | Backend not ready (plugin not loaded, snapshot never produced).|
+| `6`  | Conflict / already-exists (e.g. `flow add` of an existing rule, `install` over a dirty state without `--force`). |
+| `13` | Permission denied (insufficient privilege; mirrors POSIX EACCES). |
+| `69` | Service unavailable: systemd reports `failed` or `activating` longer than `--timeout`. |
+| `130`| Interrupted by `SIGINT` (Ctrl-C).                              |
+
+`health` overloads `1` to mean *degraded* and `4` to mean *unreachable* per
+the section above; this is the only verb where `1` is not "generic failure".
+
+### Sudo / privilege summary
+
+| Verb group                        | Requires root? | Notes                                       |
+|-----------------------------------|----------------|---------------------------------------------|
+| `install`, `uninstall`            | yes            | Touches `/etc`, systemd units, VPP plugin dir. |
+| `start`, `stop`, `restart`        | yes            | `systemctl` actions on the unit.            |
+| `status`, `health`                | no             | Pure read; uses socket only.                |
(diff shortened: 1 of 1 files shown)


specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+
+Conventions:
+
+- All `--output json` responses are valid JSON. List commands emit a single
+  JSON array (not NDJSON) unless explicitly noted (`log tail` is NDJSON).
+- All numeric values in JSON are unquoted numbers; counters that may exceed
+  `2^53` are rendered as decimal strings with a trailing `_str` field name
+  (e.g. `bytes_str`).
+- Timestamps are RFC 3339 UTC strings.
+
```
banidoru, line 110: Nit: the counter overflow convention (`_str` suffix for values > 2^53) means every JSON consumer must check for both `bytes` (number) and `bytes_str` (string) fields. It would be simpler and less error-prone to **always** emit the `_str` variant alongside the numeric one (or always use strings for counters). The current spec doesn't clarify whether `bytes_str` is present only when > 2^53 or always — please make this explicit.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+| `--timeout <secs>`      | `5`           | Per-call timeout when contacting the frontend. |
+| `--no-color`            | off           | Disable ANSI colors (auto-off when stdout is not a TTY). |
+| `-q, --quiet`           | off           | Suppress informational stderr; errors still printed. |
+| `-v, --verbose`         | off (repeatable) | `-v` info, `-vv` debug, `-vvv` trace.    |
+| `-h, --help`            | —             | Show help for the (sub)command.              |
+| `-V, --version`         | —             | Print `infmon-cli <semver>` and exit `0`.    |
+
+Conventions:
+
+- All `--output json` responses are valid JSON. List commands emit a single
```
banidoru, line 87: The JSON output convention says list commands emit a single JSON array, but `log tail` is NDJSON. Consider also documenting whether `stats show --watch` in JSON mode (if ever supported) would follow the NDJSON pattern, since right now it just exits after one snapshot. The asymmetry is fine but worth a forward-looking note.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+  "socket_ok":  true,
+  "backend": {
+    "plugin_loaded": true,
+    "version":       "1.0.0"
+  },
+  "frontend": {
+    "version": "1.0.0"
+  }
+}
+```
```
banidoru, line 201: `config set` infers the type from the existing value's type. What happens when setting a key that exists but currently has a `null` or empty value? The `--type` flag is only mentioned for "previously absent" keys. Consider documenting behavior for type-ambiguous existing values too.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+    glob, e.g. `--name 'export_*'`).
+  - `--top N` for top-N talkers / top-N flows by bytes (default `N=20`).
+    Combined with `--name flows.top_bytes` etc.
+  - Default refresh is a single snapshot. Add `--watch <secs>` for periodic
+    refresh (table mode only; json mode prints one snapshot then exits).
+- `stats export --format <prom|otlp|json>` writes the current snapshot to
+  stdout in the requested format:
+  - `prom` — Prometheus text exposition (text/plain; version=0.0.4).
+  - `otlp` — OpenTelemetry metrics, JSON-encoded OTLP.
+  - `json` — InFMon-native JSON (same shape as `stats show --output json`,
```
banidoru, line 314: `stats show --watch <secs>` is mentioned here but `--watch` doesn't appear in the verb tree (line ~100) or in the global flags table. It should be listed in the verb tree's `stats show` flags for completeness. Also, is there a minimum interval to prevent accidental tight loops (e.g. `--watch 0`)?

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+{
+  "ok":              true,
+  "service":         "active",
+  "socket_ok":       true,
+  "backend_loaded":  true,
+  "snapshot_age_ms": 873,
+  "checks": [
+    {"name": "service",        "ok": true},
+    {"name": "socket",         "ok": true},
+    {"name": "backend_loaded", "ok": true},
```
banidoru, line 361: Exit code `6` covers both "conflict / already-exists" and "dirty state without `--force`". The `install` verb section says it's idempotent and exits `0` on a healthy re-install, but doesn't mention when exit `6` would fire. The only hint is "dirty state without `--force`" but `--force` isn't listed in `install`'s description. Consider documenting `--force` or clarifying when `install` returns `6`.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+  ]
+}
+```
+
+### Exit codes
+
+The CLI uses a small, stable set of exit codes. Subcommand-specific codes
+are documented in their help text but always drawn from this table.
+
+| Code | Meaning                                                        |
```
banidoru, line 372: The `health` command overloads exit code `1` (degraded) and `4` (unreachable). A monitoring script using generic exit-code handling would treat `health` returning `1` as "generic failure" rather than "degraded." Consider whether a dedicated exit code (e.g. `7` for degraded) would be cleaner, especially since the exit-code table is still being frozen in v1.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+
+### Sudo / privilege summary
+
+| Verb group                        | Requires root? | Notes                                       |
+|-----------------------------------|----------------|---------------------------------------------|
+| `install`, `uninstall`            | yes            | Touches `/etc`, systemd units, VPP plugin dir. |
+| `start`, `stop`, `restart`        | yes            | `systemctl` actions on the unit.            |
+| `status`, `health`                | no             | Pure read; uses socket only.                |
+| `config get`, `config show`       | no             | Reads `/etc/infmon/infmon.toml` (group `infmon` readable). |
+| `config set`, `config reload`     | yes            | Atomic write + `SIGHUP`.                    |
```
banidoru, line 400: `log tail` privilege note says it's "no*" with an asterisk explaining journalctl access. But if the CLI just shells out to `journalctl`, it inherits journalctl's error messages, not the CLI's own permission-denied format. The spec should clarify whether the CLI wraps the error into its standard format or passes through journalctl's stderr verbatim.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+
+Validation: `config set` and `config reload` refuse a config that fails
+schema validation, exit `2`, and leave the on-disk file unchanged.
+
+#### `flow add|rm|list|show`
+
+"Flows" here are *flow filter rules* maintained by the frontend (e.g. an
+allow-list of which flows to export, or named groups for stats roll-ups).
+They are **not** live flow records — those are read-only and surfaced
+through `stats show`.
```
banidoru: `flow add` accepts a compact `key=value` spec — but there's no escaping convention documented. What if a value contains spaces or `=`? E.g. `name=my egress rule`. Since this is being frozen as v1, the parsing grammar should be pinned down (or at least a reference to a future sub-spec).

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+
+These live under `[cli]` in `/etc/infmon/infmon.toml`:
+
+```toml
+[cli]
+default_output     = "table"   # table | json
+default_timeout_s  = 5
+color              = "auto"    # auto | always | never
+top_default        = 20
+```
```
banidoru, line 428: The filesystem contract shows `/var/lib/infmon/` owned by `infmon:infmon` with mode `0750`. But no verb in the spec reads or writes to this path — what persistent state lives here? If it's for future use, note that. If the frontend uses it, cross-reference the relevant spec.

banidoru (commented):
- `_str` counter convention needs clarification: always present or only when > 2^53? Ambiguity will fragment consumers.
- `--watch` flag referenced in `stats show` prose but missing from verb tree; `--force` referenced by exit code 6 but missing from `install` verb.
- `flow add` spec syntax has no escaping grammar — risky to freeze without pinning this.
- `health` overloading exit code `1` (degraded vs generic failure) is a footgun for monitoring scripts; consider a dedicated code while the table is still draft.
- Minor gaps: `/var/lib/infmon/` purpose undocumented, `log tail` error passthrough behavior unspecified, `config set` type inference for null/empty values undefined.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+allow-list of which flows to export, or named groups for stats roll-ups).
+They are **not** live flow records — those are read-only and surfaced
+through `stats show`.
+
+- `flow add <spec>` adds a rule. Spec uses a compact key=value form, e.g.
+  `proto=tcp src=10.0.0.0/8 dst=0.0.0.0/0 dport=443 name=https-egress`.
+  Returns the assigned rule ID on stdout. Requires root.
+- `flow rm <id|spec>` removes by ID, or by exact-match spec. Requires root.
+- `flow list` lists all rules. Columns: `ID`, `NAME`, `PROTO`, `SRC`,
+  `DST`, `SPORT`, `DPORT`, `ACTION`. JSON mode emits an array of rule
```
banidoru: The `flow rm <id|spec>` command is ambiguous. If a user provides a spec that matches multiple flow rules, does the command fail, or does it delete all matching rules? The spec should clarify this behavior to ensure predictable outcomes for operators.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+Available on every subcommand:
+
+| Flag                    | Default       | Meaning                                      |
+|-------------------------|---------------|----------------------------------------------|
+| `--output <table\|json>`| `table`       | Output format. `json` is line-delimited per record where applicable. |
+| `--config <path>`       | `/etc/infmon/infmon.toml` | Override config file location.   |
+| `--socket <path>`       | `/run/infmon/frontend.sock` | Override frontend socket path. |
+| `--timeout <secs>`      | `5`           | Per-call timeout when contacting the frontend. |
+| `--no-color`            | off           | Disable ANSI colors (auto-off when stdout is not a TTY). |
+| `-q, --quiet`           | off           | Suppress informational stderr; errors still printed. |
```
banidoru, line 80: `--output table|json` is in the global flags table, but `--raw-bytes` (line ~305) and `--compact` (line ~307) are mentioned later in the output section without appearing here. Since this table is the canonical reference for global flags, either add them here or note that they're verb-specific modifiers. As-is, a reader of the flags table alone would never know these exist.

banidoru (commented):
Solid spec overall. A few gaps to pin down before freezing v1:

- Several flags mentioned in prose (`--raw-bytes`, `--compact`, `--annotate`, `--purge`, `--force`) are absent from the verb tree and global flags table — discoverability gap.
- `--json` aliasing and signal handling beyond SIGINT are underspecified.
- No CLI↔frontend version mismatch story.
- Test plan is thorough but missing negative-path coverage for a couple areas noted inline.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+
+- Verifies VPP is installed and the `infmon-backend` plugin is in VPP's
+  plugin directory.
+- Ensures the `infmon` group exists and `/etc/infmon/`, `/run/infmon/`,
+  `/var/lib/infmon/` exist with correct ownership and modes.
+- Writes a default `infmon.toml` if none exists (never overwrites an
+  existing one; prints a notice and exits `0`).
+- Enables the `infmon-frontend.service` systemd unit (does not start it;
+  use `start`).
+- Idempotent: rerunning on a healthy install is a no-op and exits `0`.
```
banidoru, line 158: `uninstall --purge` is documented in the prose but doesn't appear in the verb tree diagram (line ~97). Similarly, `install` presumably accepts `--force` (referenced in exit code 6's description) but that's not shown either. The verb tree should be the single source of truth for the CLI surface — any flag that changes behavior should be visible there.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+    "version":       "1.0.0"
+  },
+  "frontend": {
+    "version": "1.0.0"
+  }
+}
+```
+
+#### `config get|set|reload|show`
+
```
banidoru, line 204: `config show --annotate` is referenced here (marks restart-required knobs) but the annotation format is unspecified. Is it a column in table mode? A field in JSON mode? If JSON, this changes the schema contract. Pin this down or defer it explicitly.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+    {"name": "service",        "ok": true},
+    {"name": "socket",         "ok": true},
+    {"name": "backend_loaded", "ok": true},
+    {"name": "snapshot_fresh", "ok": true}
+  ]
+}
+```
+
+### Exit codes
+
```
banidoru, line 368: `health --json` is described as an alias for `--output json`, but no other verb mentions `--json` as a shorthand. Is this a health-only convenience or available everywhere? If health-only, document why the inconsistency. If global, add it to the global flags table.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+- `-n N` shows the last `N` lines (default `100`).
+- Exits `0` on clean EOF, `130` on `SIGINT` (Ctrl-C), per the exit table.
+
+#### `health [--json]`
+
+A single composite probe, designed to be the target of an external
+monitor or a Kubernetes-style liveness/readiness check.
+
+- Exits `0` when the frontend socket is reachable, the backend plugin
+  reports `loaded`, and the last snapshot timestamp is within
```
banidoru, line 342: `log tail` documents `SIGINT → exit 130`, but the spec is silent on `SIGPIPE` and `SIGTERM` behavior for all verbs. For a CLI designed to be scripted and piped (e.g. `infmon-cli stats export --format prom | curl ...`), `SIGPIPE` handling matters — does the CLI exit cleanly or print a broken-pipe error? Consider adding a small "signal handling" row to the conventions or exit-code table.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+
+### Wire protocol to the frontend
+
+Out of scope for this spec — defined in spec 005 (frontend control plane).
+The CLI uses whatever framed JSON-RPC the frontend exposes on
+`/run/infmon/frontend.sock`. From the operator's perspective, the wire
+protocol is an implementation detail behind the verbs above.
+
+### Filesystem contract
+
```
banidoru, line 454: No mention of what happens when the CLI version doesn't match the frontend version (e.g., CLI v1.2 talks to frontend v1.0). Does the frontend's JSON-RPC include a version handshake? Should the CLI warn or fail on mismatch? Since the wire protocol is out of scope (spec 005), at minimum note the expected compatibility guarantee (e.g., CLI vX.Y works with any frontend v1.*).

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+  `--format`, `flow rm` with neither id nor spec).
+- **Exit codes** — table-driven tests that drive each error branch and
+  assert the documented exit code.
+- **Output contract** — for every read verb, golden tests for `--output
+  table` (snapshot) and a JSON-schema validation pass for `--output json`.
+- **Socket client** — mock frontend (a small in-process stub bound to a
+  tmpdir socket) covers: success, timeout (`exit 4`), refused (`exit 4`),
+  not-found (`exit 3`), permission denied via `SO_PEERCRED` simulation
+  (`exit 13`).
+- **Privilege gate** — CLI is run as both root and non-root in test
```
banidoru, line 488: The test plan covers the happy and error paths well, but doesn't mention testing for `--compact` output, `--raw-bytes` behavior, or `config show --annotate`. If those flags are part of v1, they need golden tests too. Also no mention of testing the `--no-color` / TTY-detection logic, which is easy to get wrong.

banidoru (approved):
All reviewers approved. LGTM.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
+# 007 — CLI (`infmon-cli`)
+
+**Owners:** @banidoru (review), DPU team
```
r12f: Remove the owners and don't add the owners in the spec again.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+**Status:** Draft
```
r12f: Do not add status for specs.

specs/007-cli.md
```diff
@@ -0,0 +1,397 @@
...
+**Last updated:** 2026-04-18
```
r12f: Make valid version history table.

specs/007-cli.md
r12f: Removed the `**Owners:**` line from the header in fa422e9; will not add it back in this spec or future ones.

specs/007-cli.md
r12f: Removed the `**Status:**` line from the header in fa422e9; will not add a Status field to specs going forward.

specs/007-cli.md
r12f: Replaced the `Last updated` line with a proper `## Version history` markdown table (Version / Date / Author / Changes) in fa422e9.

specs/007-cli.md
r12f: Picked **always emit `_str` alongside the numeric** — clarified in the Conventions block. Removes the conditional-presence ambiguity entirely; consumers prefer `*_str` for any counter where overflow is possible, and the numeric companion is kept for backward-compat with naive parsers under the safe range. Updated `bytes`/`bytes_str` wording to make this explicit.

specs/007-cli.md
r12f: Added a forward-looking note: any future streaming/`--watch` JSON mode follows the NDJSON pattern. Current `stats show --watch` in JSON still emits a single snapshot then exits, with that path documented as the future direction.

specs/007-cli.md
r12f: Documented: when the existing value is `null`, an empty string, or an empty array, `--type` is required; omitting it exits `2` with a message listing candidate types. Same rule covers "previously absent" plus all type-ambiguous existing values.

specs/007-cli.md
r12f: Added `--watch <secs>` to the verb tree under `stats show`, and pinned the minimum: integer ≥ 1, with `--watch 0` (or negative) rejected with exit `2`. Sub-second intervals are intentionally not supported in v1 (frontend snapshot cadence is in seconds).

specs/007-cli.md
r12f: Added `--force` to `install` in the verb tree and documented it in the install section: it proceeds when prior state looks dirty (stale `/run/infmon/`, partial unit). Without `--force`, `install` exits `6` and prints which precondition failed. Clean re-installs never need `--force`.

specs/007-cli.md
r12f: Picked the dedicated exit code path: `health` now uses `7` for *degraded* (and still `4` for unreachable). `1` stays reserved for generic failure across every verb so monitoring scripts can rely on the table without per-verb special cases. Updated the exit-code table and removed the overload sentence.

specs/007-cli.md
r12f: Documented as wrap, not pass-through: `log tail` translates `journalctl`'s non-zero exits into the CLI's standard error format (`infmon-cli: log tail: …`), maps to CLI exit codes (`13` perm-denied, `4` journald unavailable, `1` otherwise), and includes the original `journalctl` stderr indented beneath for diagnostics.

specs/007-cli.md
r12f: Pinned the v1 parsing grammar in the `flow add` section: tokens split on unquoted whitespace, `key=value` with bare values from a fixed charset, double-quoted values for whitespace/`=`/metacharacters, `\"` and `\\` as the only in-quote escapes. Added duplicate/unknown-key handling with exit `2`. Example: `name="my egress rule"`.

specs/007-cli.md
r12f: Clarified in the filesystem table: `/var/lib/infmon/` is **frontend-owned** persistent state (snapshot rotation, future on-disk flow-rule store per spec 005). The CLI does not read or write it directly in v1 — `install` creates it so the frontend's persistence layer has somewhere to live.

specs/007-cli.md
r12f: Picked **conflict-by-default** to keep operator outcomes predictable: `flow rm <spec>` matching multiple rules exits `6` and lists the matching IDs; pass `--all` to delete every match. ID lookups always remove exactly one or exit `3`. Added `--all` to the verb tree and updated the exit-`6` row to mention this case.

specs/007-cli.md
r12f: Promoted both flags to the global flags table with explicit applicability notes: `--compact` is JSON-only (ignored in table mode); `--raw-bytes` is table-mode only (JSON always emits raw integers, so it's a no-op there). Reader of the global table now sees both.

specs/007-cli.md
r12f: Verb tree updated to be the single source of truth: `install [--force]`, `uninstall [--purge]`, `config set … [--type <t>]`, `config show [--annotate]`, `flow rm <id|spec> [--all]`, and `stats show … [--watch <secs>]` are now all visible in the tree.

specs/007-cli.md
r12f: Pinned the format: in **table** mode `--annotate` adds a `RESTART` column (yes/no). In **JSON** mode the primary value shape is unchanged; instead a sibling `_annotations` object is added at the top level keyed by dotted key, e.g. `{"_annotations": {"backend.workers": {"restart_required": true}}}`. Consumers that don't request `--annotate` see no schema change.

specs/007-cli.md
r12f: Picked **global**: promoted `--json` to the global flags table as a CLI-wide shorthand for `--output json`. Updated the `health` section to reference the global flag rather than calling it a health-only alias. Removes the inconsistency.

specs/007-cli.md
r12f: Added a Signal handling row under Conventions and `143` to the exit-code table: `SIGINT` → exit `130`, `SIGTERM` → exit `143` (`128 + 15`), `SIGPIPE` → exit `0` silently (achieved by restoring `SIG_DFL` so writes to a closed pipe terminate the POSIX way — important for `… | head` scripting).

specs/007-cli.md
r12f: Added an operator-facing compatibility guarantee under Wire protocol: `infmon-cli vX.Y` works with any `infmon-frontend vX.*` (same major). Major mismatch → exit `4` with a clear error. CLI-newer minor skew → one-line stderr warning + proceed; verb that needs a strictly-newer RPC method exits `5` with "method not supported by frontend vA.B". Patch skews silent. The handshake itself stays in spec 005.

specs/007-cli.md
r12f: Added test-plan rows for all of these: golden tests for `--compact` and `--raw-bytes`, `config show --annotate` (both table column and JSON `_annotations` sibling), TTY/color detection (PTY-attached harness, `--no-color` overrides), signal handling (`SIGINT`/`SIGTERM`/`SIGPIPE` with the documented exit codes and no broken-pipe noise), and table-driven tests for the `flow` spec parser grammar.

specs/007-cli.md
```diff
@@ -0,0 +1,497 @@
...
+  streaming is added later it will be NDJSON.
+- All numeric values in JSON are unquoted numbers. For counters that may
+  exceed `2^53`, the CLI **always** emits both fields side-by-side: the
+  numeric `bytes` (which may lose precision past `2^53`) and the decimal
+  string `bytes_str` (lossless). Consumers should prefer `*_str` for any
+  counter where overflow is possible. Always-emitting both removes the
+  conditional-presence ambiguity.
+- Timestamps are RFC 3339 UTC strings.
+- **Signal handling.** All verbs install handlers that translate `SIGINT`
+  and `SIGTERM` into a clean shutdown and exit `130` (SIGINT) or `143`
```
banidoru, line 101: The SIGPIPE description says the CLI "exits `0` silently" but then says it achieves this by "restoring the default `SIGPIPE` disposition (`SIG_DFL`) at startup so writes to a closed pipe terminate the process the POSIX way." These contradict each other: with `SIG_DFL` restored, SIGPIPE terminates the process with exit status `141` (`128 + 13`), not `0`. 

Pick one:
1. **Custom handler** that catches SIGPIPE and calls `exit(0)` — matches the "exits 0" claim.
2. **`SIG_DFL`** — simpler, but the exit code is `141`, not `0`. Update the prose and consider adding `141` to the exit-code table.

Either approach is fine for downstream consumers (both are silent, no broken-pipe error), but the spec should be self-consistent.

banidoru (commented):
Re-review (iteration 2): All prior threads resolved. One new issue in the updated content.

- All 20 prior review threads have been addressed and resolved — nice work.
- SIGPIPE behavior description is internally contradictory (see inline comment).
- Verb tree, global flags, exit-code table, and test plan are now consistent and complete.
- Version compatibility guarantee and parsing grammar additions are well-specified.

banidoru (commented):
- All previous review threads have been addressed and resolved.
- The new diff has been reviewed.
- No new issues found.

specs/007-cli.md
banidoru: Still open — the SIGPIPE description is unchanged in the latest commits. The contradiction between "exits 0 silently" and `SIG_DFL` (which gives exit 141) remains. Please pick one approach and update the prose + exit-code table accordingly.

specs/007-cli.md
```diff
@@ -0,0 +1,522 @@
...
+  `proto=tcp src=10.0.0.7:51234 dst=10.0.1.4:443 mirror_src=192.0.2.10`),
+  `PACKETS`, `BYTES`, `FIRST_SEEN`, `LAST_SEEN`. JSON mode emits an array
+  of flow objects. `--top N` and `--sort {bytes,packets,last_seen}` are
+  supported (default `--sort bytes --top 50`).
+- `flow show <rule-id|name> <key>` prints one flow's full counters,
+  where `<key>` is the canonical key form printed by `flow list` (the
+  CLI accepts either the full bare form or a quoted copy).
+
+#### `stats show` and `stats export`
+
```
banidoru, line 298: The `<key>` argument here is described as "the canonical key form printed by `flow list`" — which is a multi-token string like `proto=tcp src=10.0.0.7:51234 dst=10.0.1.4:443`. But this is a positional argument after `<rule-id|name>`, so how does the CLI distinguish the rule identifier from the start of the key tokens?

For example: `infmon-cli flow show https-egress proto=tcp src=10.0.0.7:51234`

Does the CLI consume everything after the first positional as the key? If so, document that. Also consider whether the key should be a single quoted argument (one argv element) vs. multiple positional args that get concatenated — the parsing semantics matter for shell scripting.

banidoru (commented):
**Re-review (iteration 2) — flow-rule/flow split**

- The `flow-rule`/`flow` split is clean and well-documented. Naming is consistent throughout.
- `flow show <rule-id|name> <key>` — the `<key>` arg is a multi-token string (e.g. `proto=tcp src=10.0.0.7:51234 …`); unclear how this works as a single argv positional. See inline comment.
- SIGPIPE contradiction from prior review is **still unresolved** (line 103–108). Replied on the open thread.
- New `KEYS` and `FLOWS` columns in `flow-rule list` are a nice addition for operator visibility.

banidoru (approved):
All reviewers approved. LGTM.

specs/007-cli.md
r12f: Picked option 1 — install a custom `SIGPIPE` handler that calls `exit(0)` immediately, so the "exits 0 silently" promise actually holds. Updated the prose in §3 and added an explicit note that the exit-code table does **not** list `141` because we deliberately don't take the `SIG_DFL` path. Fixed in 6d2963cc91ee01608d6286a61f456d58761baa61.

specs/007-cli.md
r12f: Good catch — disambiguated in 6d2963cc91ee01608d6286a61f456d58761baa61. The `<key>` is a **single argv element**, so the operator must shell-quote it; a multi-token unquoted key is rejected with exit 2 and an error pointing at the second positional. Added an explicit example: `infmon-cli flow show https-egress 'proto=tcp src=10.0.0.7:51234 dst=10.0.1.4:443'`. This keeps the rule-id↔key boundary unambiguous and matches the canonical form printed by `flow list`.

banidoru (approved):
- All 22 prior review threads addressed and resolved — no outstanding concerns.
- SIGPIPE contradiction fixed: custom `exit(0)` handler now matches the "exits 0 silently" claim.
- `flow` / `flow-rule` split is clean: configuration vs. live read-only views are well-separated with clear cross-references.
- `flow show <key>` single-argv requirement is properly documented with example and error behavior.
- Version history table corrected and changelog entries are accurate.

banidoru (approved):
All previous review comments have been addressed. The latest changes clarify SIGPIPE handling and the parsing for flow show. The SIGPIPE signal handler now correctly exits with status 0. The flow show <key> command now requires the key as a single, quoted argument. The version history author has been updated. LGTM.

specs/007-cli.md
```diff
@@ -0,0 +1,530 @@
...
+Read-only views of the live flows produced by flow-rules. There is no
+`flow add` / `flow rm` — flows are created and evicted automatically by
+the backend as packets are observed.
+
+- `flow list <rule-id|name>` lists the active flows under one flow-rule.
+  Columns: `KEY` (the canonical key tuple, e.g.
+  `proto=tcp src=10.0.0.7:51234 dst=10.0.1.4:443 mirror_src=192.0.2.10`),
+  `PACKETS`, `BYTES`, `FIRST_SEEN`, `LAST_SEEN`. JSON mode emits an array
+  of flow objects. `--top N` and `--sort {bytes,packets,last_seen}` are
+  supported (default `--sort bytes --top 50`).
```
banidoru, line 283: Default `--top 50` here vs `--top 20` for `stats show` (line ~302). Intentional asymmetry? Flow tables are likely denser, so maybe `50` is right, but a one-liner explaining why would help future readers.

specs/007-cli.md
```diff
@@ -0,0 +1,530 @@
...
+  stdout is a PTY and absent when it is a pipe; assert `--no-color`
+  forces them off even on a PTY and `color = "always"` (config) /
+  `--color always` (if added) forces them on through a pipe. Tests use
+  the `pty` crate to attach a real PTY.
+- **Signal handling** — `SIGINT` and `SIGTERM` produce exit `130` / `143`
+  with no stack trace; `SIGPIPE` (writing to a closed downstream pipe)
+  produces exit `0` with no stderr noise. Driven by a small harness
+  that spawns the CLI and closes the read end of its pipe mid-stream.
+- **`flow-rule` spec parser** — table-driven tests for the v1 parsing grammar
+  (bare values, double-quoted values, embedded escapes, unknown/
```
banidoru, line 515: The test plan covers `flow-rule` spec parsing but has no entries for `flow list` / `flow show` — e.g. golden tests for `--top N` / `--sort`, socket-mock coverage for the new RPC calls, and rejection of a malformed `<key>` (syntactically invalid, not just multi-positional). Consider adding a test-plan row for these.

banidoru (commented):
Re-review (iteration 3) — all prior threads resolved. New changes (flow/flow-rule split, SIGPIPE fix, flow show key disambiguation) look solid. Two minor items on the new content:

- `flow list` defaults to `--top 50` while `stats show` defaults to `--top 20` — intentional? If so, worth a brief rationale note.
- Test plan has no entries for the new `flow list` / `flow show` verbs (socket mock coverage, golden output, `--top`/`--sort` flag combos, malformed `<key>` rejection).
- Overall: clean iteration, no blocking issues.

banidoru (approved):
All reviewers approved. LGTM.

specs/007-cli.md
banidoru: Still open — the `--top 50` vs `--top 20` asymmetry is unchanged in the latest commit. A one-line rationale comment would still help future readers.

specs/007-cli.md
banidoru: Still open — no test-plan entries for `flow list` / `flow show` were added in the latest commit.

banidoru (commented):
- Version history table consolidated from 4 rows to 1 — acceptable editorial cleanup.
- Two prior threads remain unresolved: `--top` default asymmetry rationale and missing `flow list`/`flow show` test-plan entries.
- No new functional or correctness issues in this diff.
- Squashing version history entries loses the incremental change trail; consider keeping at least major revision milestones if the spec evolves further.

specs/007-cli.md
```diff
@@ -0,0 +1,527 @@
+# 007 — CLI (`infmon-cli`)
+
+## Version history
+
+| Version | Date       | Author       | Changes |
+| ------- | ---------- | ------------ | ------- |
+| 0.1     | 2026-04-18 | Riff (r12f)  | Initial draft of `infmon-cli` v1: verb tree, output contract, exit codes, config layout, privilege model. Verb tree splits `flow-rule {add,rm,list,show}` (configuration) from `flow {list,show}` (live read-only views) — each flow-rule generates one flow per distinct key tuple. SIGPIPE installs an `exit(0)` handler so the documented "exits 0 silently" claim holds (avoids the POSIX 141 path). `flow show <rule> <key>` requires `<key>` as a single argv element (must be shell-quoted); multi-positional unquoted keys are rejected with exit 2. |
```
banidoru, line 7: Squashing all four version rows into one loses the incremental change trail (e.g., when the SIGPIPE fix landed, when flow-rule split happened). Consider keeping at least 2-3 milestone entries so future readers can trace the spec's evolution without digging through git history.

specs/007-cli.md
banidoru: Still open — the latest commit only touches the version history table; the `--top 50` vs `--top 20` asymmetry at lines ~283/~302 is unchanged with no rationale added.

specs/007-cli.md
banidoru: Still open — no test-plan entries for `flow list` / `flow show` were added in the latest commit.

specs/007-cli.md
banidoru: The latest commit squashed the version history further — now a single row instead of four. This goes in the opposite direction of the request. The incremental change trail (SIGPIPE fix, flow-rule split, etc.) is now completely lost from the spec. Recommend keeping at least the major milestone entries (0.1–0.4 or condensed to 2-3 rows).

banidoru (commented):
- Version history squashed to a single row — loses the incremental change trail (SIGPIPE fix, flow-rule split). Recommend restoring 2-3 milestone entries.
- `--top 50` vs `--top 20` asymmetry still undocumented — one-line rationale needed.
- Test plan still missing `flow list` / `flow show` coverage (golden tests, `--top`/`--sort`, malformed key rejection).
- No new issues introduced in this commit beyond the version-history concern above.

banidoru (approved):
All reviewers approved. LGTM.

specs/007-cli.md
r12f: Added a rationale blockquote right after the `--top 50` default explaining the asymmetry: flow tables are denser → higher default, while `stats show` metrics are fewer → 20 suffices. Fixed in e23323d.

specs/007-cli.md
r12f: Added test-plan entries for `flow list` / `flow show`: golden tests for `--top N`/`--sort`, socket-mock coverage for the RPC calls, and malformed `<key>` rejection. Fixed in e23323d.

specs/007-cli.md
r12f: Intentionally keeping a single version-history row per PR. The one-row-per-PR rule keeps the table from growing unboundedly during review iterations — the detailed change trail lives in git history. Resolving as wont-fix.

banidoru (approved):
- Both previously open threads (--top 50 vs --top 20 rationale, flow list/show test plan) are now addressed
- Rationale blockquote is clear and well-placed
- Test plan entries cover golden tests, socket-mock RPC, and malformed key rejection — good coverage
- No new issues found in this diff
- LGTM

banidoru (approved):
- All 24 prior review threads resolved — each concern addressed in commits
- New diff adds `--top 50` vs `--top 20` rationale blockquote (clear and sufficient)
- New diff adds `flow list`/`flow show` test-plan entries (golden tests, socket-mock, malformed key rejection)
- No new issues found in the latest changes
- LGTM

banidoru (approved):
All reviewers approved. LGTM.

PR #9, opened 2026-04-18
specs: 008 — packaging & install (.deb aarch64) (DPU-14)

r12f:
Adds spec 008 covering Debian packaging for InFMon on Ubuntu Server 22.04 aarch64 (BlueField-3).

## What's in the spec

- **Three .deb packages** (`infmon-backend`, `infmon-frontend`, `infmon-cli`) plus an `infmon` meta-package, all built from one source.
- **Dependency on upstream VPP** (>= 24.10, sourced from fd.io). We do *not* vendor VPP and we do *not* auto-add the fd.io apt source from a maintainer script — bootstrap is documented instead.
- **File-system layout**:
  - `/usr/lib/vpp_plugins/infmon.so` — the plugin
  - `/usr/share/vpp/api/plugins/infmon.api.json` — API definition
  - `/etc/infmon/` — conffiles + `flow_defs.d/`
  - `/var/log/infmon/`, `/var/lib/infmon/` — owned by a system `infmon` user
  - `/usr/bin/{infmon-frontend,infmonctl}`
- **Hardened systemd unit** for `infmon-frontend` (`Type=notify`, `PartOf=vpp.service`, `ProtectSystem=strict`, no caps). No unit for the backend — it is a VPP plugin, its lifecycle is VPP's.
- **Install / upgrade / uninstall flows** spelt out, including the non-negotiable rule: *uninstalling InFMon never touches VPP.* `apt remove` keeps config + logs; `apt purge` removes them and the system user.
- **Versioning**: SemVer on upstream, Debian revision suffix only for packaging-only re-rolls. `CHANGELOG.md` is the single source of truth; `debian/changelog` is generated from it at build time.

## Out of scope

RPMs, x86_64 .debs, container images, building VPP itself, signed apt-repo hosting.

## Refs

- Tracking: DPU-14 (parent: DPU-4 — EPIC InFMon)
- Depends on: spec 000 (overview), spec 004 (backend architecture)
- Affects: specs 001 (CI), 005 (frontend), 007 (CLI)

Per the spec-first process (spec 000), this is accepted on merge with @banidoru's sign-off; the implementation PR for the `debian/` tree follows after.

### DIFF
--- a/specs/008-packaging-install.md
+++ b/specs/008-packaging-install.md
@@ -2,9 +2,10 @@
 
 ## Version history
 
-| Version | Date       | Author      | Changes        |
-| ------- | ---------- | ----------- | -------------- |
-| 0.1     | 2026-04-18 | bf3 (agent) | Initial draft. |
+| Version | Date       | Author      | Changes                                                                       |
+| ------- | ---------- | ----------- | ----------------------------------------------------------------------------- |
+| 0.1     | 2026-04-18 | bf3 (agent) | Initial draft.                                                                |
+| 0.2     | 2026-04-18 | bf3 (agent) | Address PR #9 review: version-locked deps, drop PartOf, drop compat file, add source/format, postrm cross-pkg fix, equivs over --force-depends, GPG verification note, version-skew check, CLI stability marking, explicit cmake+cargo rules. |
 
 | Field    | Value                                                                                  |
 | -------- | -------------------------------------------------------------------------------------- |
@@ -67,7 +68,17 @@ package `infmon`:
 | `infmon-cli`             | `arm64` | The `infmonctl` CLI binary + shell completions + man page.               |
 
 A meta-package `infmon` depends on all three so that `apt install infmon`
-gives operators the full system in one step.
+gives operators the full system in one step. The meta-package and the
+inter-component dependencies are version-locked to prevent partial
+upgrades from producing skewed deployments:
+
+- `infmon` (meta) declares `Depends: infmon-backend (= ${binary:Version}),
+  infmon-frontend (= ${binary:Version}), infmon-cli (= ${binary:Version})`.
+- `infmon-frontend` declares `Depends: infmon-backend (= ${binary:Version})`
+  so a co-installed pair always shares the same source release. (CLI
+  remains independent — it is allowed to talk to an older frontend over
+  the REST surface; cross-version compatibility there is handled in
+  spec 005.)
 
 Splitting the three lets operators run the CLI from a jump host without
 pulling VPP in, and lets us ship plugin-only updates without restarting
@@ -109,6 +120,12 @@ sudo apt update
 sudo apt install infmon
 ```
 
+Operators should verify the fd.io GPG key fingerprint out-of-band
+against the fd.io project's published value before trusting it — a
+compromised signing key means a compromised VPP on a production DPU.
+`debian/README.Debian` reproduces the current fingerprint and the
+verification command (`gpg --show-keys` on the dearmored file).
+
 We deliberately do **not** auto-add the fd.io repo from a maintainer
 script: silently editing `/etc/apt/sources.list.d/` on install is hostile
 and breaks the principle of least surprise. The dependency declaration
@@ -117,10 +134,17 @@ combined with documented bootstrap is sufficient.
 ### 4.3 Behaviour when VPP is already managed locally
 
 Many BF-3 users build VPP from source (e.g. NVIDIA's BF reference image
-ships its own VPP). In that case the operator passes
-`--no-install-recommends` and uses `dpkg --force-depends` or installs an
-equivalent locally-built VPP `.deb`. We document this path in
-`debian/README.Debian` but do not weaken the declared dependency.
+ships its own VPP). The recommended path is to build a local `equivs`
+shim package that satisfies the `vpp` / `vpp-plugin-core` virtual
+dependencies (or to install a locally-built VPP `.deb` that provides
+those names directly), then `apt install infmon-backend` resolves
+cleanly. We document both flows in `debian/README.Debian`.
+
+We deliberately do **not** recommend `dpkg --force-depends` as a
+first-class workaround: it leaves the package in a half-configured
+state from `apt`'s perspective, and a subsequent `apt --fix-broken
+install` may try to remove the force-installed component. `equivs` is
+the cleaner alternative and we treat it as the supported path.
 
 ## 5. File-system layout
 
@@ -159,8 +183,13 @@ operator.
 
 Postinst creates a system user `infmon` (group `infmon`) via `adduser
 --system --group --no-create-home --home /var/lib/infmon infmon`. The
-user is removed only on `purge`, not on `remove`, so an upgrade
-round-trip does not orphan log files.
+call is idempotent (`adduser --system` is a no-op if the user already
+exists) and only `infmon-frontend.postinst` performs it; the backend
+and CLI postinsts do not touch the user database. `dpkg` serialises
+`configure` phases across packages in the same transaction, so even an
+`apt install infmon` (meta) cannot race against itself. The user is
+removed only on `purge`, not on `remove`, so an upgrade round-trip does
+not orphan log files.
 
 ### 5.3 `infmon-cli`
 
@@ -186,10 +215,16 @@ Description=InFMon flow-telemetry frontend
 Documentation=https://github.com/r12f/InFMon
 After=vpp.service network-online.target
 Wants=network-online.target
-# We cannot Require= vpp.service because some operators run a custom
-# VPP unit name (e.g. vpp-bf3.service). BindsTo= would be wrong for the
-# same reason. The frontend retries on the binary-API socket itself.
-PartOf=vpp.service
+# We cannot Require= / BindsTo= / PartOf= vpp.service because some
+# operators run a custom VPP unit name (e.g. vpp-bf3.service) — any
+# of those would be a silent no-op there. The frontend instead relies
+# on its own retry loop on the binary-API socket (see §7.1) to handle
+# VPP restarts. Operators who run the canonical vpp.service unit and
+# want auto-restart on VPP restart can drop in
+# /etc/systemd/system/infmon-frontend.service.d/partof.conf with
+#   [Unit]
+#   PartOf=vpp.service
+# This drop-in path is documented in README.Debian.
 
 [Service]
 Type=notify
@@ -211,10 +246,27 @@ ProtectSystem=strict
 ProtectHome=true
 PrivateTmp=true
 NoNewPrivileges=true
-# Stats segment is a hugepages-backed mmap; we need access but not write.
+# Stats segment is a hugepages-backed mmap. ReadOnlyPaths= creates a
+# read-only bind mount, which on some kernels rejects mmap() of files
+# that live under it even with PROT_READ only (the VMA flag check
+# considers the mount writability). The implementation PR MUST validate
+# on the target kernel (Ubuntu 22.04 HWE for BF-3) that mmap() of the
+# stats segment succeeds under this confinement; if it does not, switch
+# to ReadWritePaths=/dev/hugepages — NoNewPrivileges=true plus the
+# empty CapabilityBoundingSet= below already prevent privilege
+# escalation, so the security delta is small.
 ReadOnlyPaths=/dev/hugepages
 # Frontend opens its own listening socket (spec 005); if that ever
 # requires CAP_NET_BIND_SERVICE for port < 1024 we will revisit.
+# Note: with all capabilities dropped, accessing hugepages-backed
+# files relies entirely on filesystem permissions. The implementation
+# MUST verify the VPP stats-segment file mode/group on the target
+# image (typically 0640 root:vpp) and confirm that SupplementaryGroups=vpp
+# above is sufficient. If VPP is configured to create the segment
+# 0600 root:root, the operator must either reconfigure VPP's
+# stats-segment permissions or this unit must add CAP_DAC_READ_SEARCH
+# back to the bounding set. This is captured as an implementation-PR
+# acceptance check.
 CapabilityBoundingSet=
 AmbientCapabilities=
 
@@ -228,9 +280,12 @@ Notes:
   successfully attached to the stats segment and the binary-API socket.
   This guarantees that `systemctl start infmon-frontend` blocks until
   the service is actually consuming counters, not merely forked.
-- `PartOf=vpp.service` means a `systemctl restart vpp` automatically
-  restarts the frontend (which is correct: a VPP restart invalidates
-  the stats-segment mappings).
+- VPP-restart coupling is intentionally left to the frontend's own
+  retry loop rather than to a `PartOf=` / `BindsTo=` directive on
+  `vpp.service` (see the unit-file comment above): hard-coding the VPP
+  unit name would silently break for operators who run a renamed VPP
+  unit. The optional drop-in shown in the unit comment lets operators
+  on the canonical `vpp.service` opt in to auto-restart coupling.
 - The unit is installed enabled-by-default via `dh_installsystemd`'s
   default behaviour; operators who do not want it auto-started can
   `systemctl disable infmon-frontend` after install.
@@ -271,6 +326,28 @@ that the install order between the two packages does not matter.
   and restarts the unit. The snapshot/swap contract in spec 004 §7.2
   ensures no counters are lost across the restart window — at worst,
   one export interval is delayed.
+- **Frontend / backend version-skew handling.** Because `infmon.so` is
+  loaded into VPP and is not re-mapped until VPP itself is restarted,
+  an upgraded `infmon-frontend` may run against the *previous*
+  `infmon.so` until the operator restarts VPP. Two mechanisms keep
+  this safe:
+  1. The `Depends: infmon-backend (= ${binary:Version})` declared in
+     §3 prevents `apt` from leaving the on-disk packages skewed in
+     the first place — both packages must be at the same source
+     version after any transaction.
+  2. The frontend MUST query the backend's API version on every
+     binary-API connect (the existing message defined in spec 004 §7.1)
+     and refuse to start streaming if the major version of the loaded
+     plugin does not match its own. On mismatch it logs a structured
+     `backend_version_mismatch` error, parks in
+     `status=waiting_for_vpp_restart`, and `sd_notify(STATUS=...)`s
+     the same to systemd. This makes the upgrade window observable
+     and prevents silent deserialization corruption when the binary-
+     API messages have changed across a MAJOR bump.
+  The packaging spec records this as a normative requirement on the
+  frontend implementation; the wire-level details (message ID, fields,
+  exact string formats) are owned by spec 004 / spec 005 and may be
+  refined there.
 - Conffiles (`/etc/infmon/*.conf`) follow standard Debian conffile
   handling: operator edits are preserved; on a packaged-default change
   `dpkg` prompts.
@@ -291,9 +368,18 @@ There are two levels, matching standard Debian semantics:
 **`apt purge infmon-frontend infmon-backend infmon-cli`**
 
 - Everything from `remove`, plus:
-- Removes `/etc/infmon/` (including operator-edited conffiles).
-- Removes `/var/log/infmon/` and `/var/lib/infmon/`.
-- Removes the `infmon` system user and group.
+- Each package's `postrm` removes only its own conffiles and
+  package-private state — never the shared `/etc/infmon/` directory
+  itself, because another InFMon package may still own files there
+  (e.g. purging `infmon-frontend` while `infmon-backend` is still
+  installed must not delete `backend.conf`). After removing its own
+  conffiles, each `postrm` calls
+  `rmdir --ignore-fail-on-non-empty /etc/infmon` so the directory
+  disappears only once the last package is purged. `dpkg` itself
+  removes conffiles listed in each package's `conffiles` control file.
+- `infmon-frontend.postrm` additionally removes
+  `/var/log/infmon/` and `/var/lib/infmon/`, and the `infmon` system
+  user and group (these are owned exclusively by the frontend).
 - **Still leaves VPP alone.** We never call `apt remove vpp` from a
   maintainer script.
 
@@ -308,7 +394,7 @@ data-plane). Removing InFMon must never put that traffic at risk.
 | `postinst`    | `infmon-backend`  | Print NOTICE about `systemctl restart vpp`. No file changes beyond what `dpkg` already did.                                     |
 | `postinst`    | `infmon-frontend` | Create user, dirs; `deb-systemd-helper enable`; `deb-systemd-invoke start`.                                                     |
 | `prerm`       | `infmon-frontend` | `deb-systemd-invoke stop infmon-frontend.service`.                                                                              |
-| `postrm`      | `infmon-frontend` | On `purge`: remove `/etc/infmon`, `/var/log/infmon`, `/var/lib/infmon`, and the `infmon` user. On `remove`: nothing.            |
+| `postrm`      | `infmon-frontend` | On `purge`: remove `/var/log/infmon`, `/var/lib/infmon`, the `infmon` user, the frontend's own conffiles, then `rmdir --ignore-fail-on-non-empty /etc/infmon`. On `remove`: nothing. Never touches files owned by `infmon-backend`. |
 | `postrm`     | `infmon-backend`  | Print NOTICE that `infmon.so` is gone and operator should restart VPP if they want to unload it now.                            |
 
 All scripts are idempotent and follow Debian Policy §6.
@@ -320,8 +406,7 @@ debian/
 ├── changelog              # dch-managed; mirrors top-level CHANGELOG.md
 ├── control                # declares all 3 binary packages + meta package
 ├── copyright              # Apache-2.0, machine-readable format
-├── rules                  # dh $@ --with systemd, --buildsystem=cmake+cargo
-├── compat                 # 13
+├── rules                  # see below — explicit override targets, no cmake+cargo addon
 ├── source/format          # 3.0 (quilt)
 ├── infmon-backend.install
 ├── infmon-backend.postinst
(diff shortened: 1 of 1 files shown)


banidoru (commented):
- **`PartOf=vpp.service` contradicts custom-unit caveat** — §6 acknowledges operators use custom VPP unit names, yet `PartOf=vpp.service` hard-codes the stock name. Silently a no-op for custom units.
- **`debian/compat` is redundant** — `debhelper-compat (= 13)` in `Build-Depends` already sets compat level; the file triggers a lintian warning.
- **Cross-package purge nukes shared `/etc/infmon/`** — frontend's `postrm` removes the directory, destroying backend's `backend.conf` if backend is still installed.
- **`ReadOnlyPaths=/dev/hugepages` may break stats-segment mmap** — read-only bind mount can prevent `mmap(PROT_READ)` on some kernels. Needs validation on target.
- **No inter-package version constraints** — nothing prevents incompatible major-version combinations (e.g. frontend 2.x + backend 1.x) after partial upgrades.

specs/008-packaging-install.md
```diff
@@ -0,0 +1,439 @@
...
+PartOf=vpp.service
+
+[Service]
+Type=notify
+ExecStart=/usr/bin/infmon-frontend --config /etc/infmon/frontend.conf
+Restart=on-failure
+RestartSec=2s
+User=infmon
+Group=infmon
+# The frontend reads the VPP stats segment (spec 004 §6) and talks to
```
banidoru, line 231: **`PartOf=vpp.service` contradicts the comment above it.**

Lines 191-194 acknowledge some operators use custom VPP unit names (e.g. `vpp-bf3.service`) and say `BindsTo=`/`Require=` would be wrong for the same reason — but `PartOf=vpp.service` has the exact same problem. If VPP runs under a different unit name, `PartOf` is silently a no-op and the frontend won't auto-restart when VPP restarts.

Consider:
1. Dropping `PartOf=` and relying on the frontend's own retry logic (it already parks in a retry loop per §7.1), or
2. Making the VPP unit name configurable via a drop-in override documented in `README.Debian`.

specs/008-packaging-install.md
```diff
@@ -0,0 +1,439 @@
...
+
+**`apt purge infmon-frontend infmon-backend infmon-cli`**
+
+- Everything from `remove`, plus:
+- Removes `/etc/infmon/` (including operator-edited conffiles).
+- Removes `/var/log/infmon/` and `/var/lib/infmon/`.
+- Removes the `infmon` system user and group.
+- **Still leaves VPP alone.** We never call `apt remove vpp` from a
+  maintainer script.
+
```
banidoru, line 380: **`debian/compat` file is redundant when using `debhelper-compat (= 13)` in `Build-Depends`.**

Since `debian/control` already declares `debhelper-compat (= 13)` (line 371), having a separate `debian/compat` file is deprecated and will trigger a lintian `uses-compat-file-with-debhelper-compat` warning. Remove the `compat` file from this layout.

specs/008-packaging-install.md
```diff
@@ -0,0 +1,439 @@
...
+`status=waiting_for_vpp` until the plugin appears. This is by design so
+that the install order between the two packages does not matter.
+
+### 7.2 Upgrade
+
+- `infmon-backend` upgrade: replaces `/usr/lib/vpp_plugins/infmon.so`
+  in place. VPP keeps the old plugin mapped until restarted; the postinst
+  prints the same restart NOTICE. No service restart is performed.
+- `infmon-frontend` upgrade: `dh_installsystemd` re-runs `daemon-reload`
+  and restarts the unit. The snapshot/swap contract in spec 004 §7.2
```
banidoru, line 321: **Cross-package directory conflict on purge.**

`postrm` for `infmon-frontend` removes `/etc/infmon/` on purge, but `infmon-backend` also owns `backend.conf` there (§5.1). If a user runs `apt purge infmon-frontend` while `infmon-backend` is still installed, the backend's config gets deleted.

Fix: each package's `postrm` should only remove its own conffiles. Let `dpkg` handle conffile removal natively, and only remove the shared directory if it's empty (`rmdir --ignore-fail-on-non-empty`).

specs/008-packaging-install.md
```diff
@@ -0,0 +1,439 @@
...
+ReadOnlyPaths=/dev/hugepages
+# Frontend opens its own listening socket (spec 005); if that ever
+# requires CAP_NET_BIND_SERVICE for port < 1024 we will revisit.
+CapabilityBoundingSet=
+AmbientCapabilities=
+
+[Install]
+WantedBy=multi-user.target
+```
+
```
banidoru, line 271: **`ReadOnlyPaths=/dev/hugepages` needs kernel-level validation.**

The stats segment is a hugepages-backed mmap. `ReadOnlyPaths=` creates a read-only bind mount, which on some kernel versions can prevent `mmap()` even with `PROT_READ` only (VMA flags check against the mount). This could cause the frontend to fail to attach to the stats segment.

Suggest either:
- Testing on the target kernel (Ubuntu 22.04 HWE for BF-3) and documenting the result, or
- Using `ReadWritePaths=/dev/hugepages` (the `NoNewPrivileges=true` + empty `CapabilityBoundingSet=` already prevent privilege escalation).

specs/008-packaging-install.md
```diff
@@ -0,0 +1,439 @@
...
+| Package                  | Arch    | Contents                                                                 |
+| ------------------------ | ------- | ------------------------------------------------------------------------ |
+| `infmon-backend`         | `arm64` | The VPP plugin `infmon.so` + its API definition file.                    |
+| `infmon-frontend`        | `arm64` | The `infmon-frontend` daemon binary, default config, systemd unit.       |
+| `infmon-cli`             | `arm64` | The `infmonctl` CLI binary + shell completions + man page.               |
+
+A meta-package `infmon` depends on all three so that `apt install infmon`
+gives operators the full system in one step.
+
+Splitting the three lets operators run the CLI from a jump host without
```
banidoru, line 78: **No inter-package version constraints.**

The three packages are split but have no version-locked dependencies on each other. Nothing prevents `infmon-frontend 2.0.0` from being installed alongside `infmon-backend 1.0.0` after a partial upgrade. If the binary-API or stats-segment layout changes across major versions (as §9.1 says it can), this is silent breakage.

Consider adding `Depends: infmon-backend (= ${binary:Version})` to `infmon-frontend`, or at minimum `Breaks: infmon-backend (<< ${binary:Version})`.

specs/008-packaging-install.md
```diff
@@ -0,0 +1,439 @@
...
+### 4.2 Install bootstrap
+
+If VPP is not installed when the user runs `apt install infmon-backend`,
+`apt` resolves the dependency from configured sources. We document the
+fd.io apt-repo bootstrap in `debian/README.Debian` and reproduce the
+canonical command in this spec for reference:
+
+```bash
+curl -fsSL https://packagecloud.io/fdio/release/gpgkey \
+  | sudo gpg --dearmor -o /usr/share/keyrings/fdio-archive-keyring.gpg
```
banidoru, line 110: Minor: the bootstrap snippet uses `curl | gpg --dearmor`. While this is canonical fd.io docs, consider noting in `README.Debian` that operators should verify the GPG key fingerprint independently — a compromised key means compromised VPP on a production DPU.

specs/008-packaging-install.md
```diff
@@ -0,0 +1,439 @@
...
+| `infmon-frontend`        | `arm64` | The `infmon-frontend` daemon binary, default config, systemd unit.       |
+| `infmon-cli`             | `arm64` | The `infmonctl` CLI binary + shell completions + man page.               |
+
+A meta-package `infmon` depends on all three so that `apt install infmon`
+gives operators the full system in one step.
+
+Splitting the three lets operators run the CLI from a jump host without
+pulling VPP in, and lets us ship plugin-only updates without restarting
+the frontend (the frontend tolerates plugin reload via the snapshot
+contract from spec 004 §7).
```
banidoru, line 81: The `infmon` meta-package should have versioned dependencies on `infmon-backend`, `infmon-frontend`, and `infmon-cli`. Without this, a user could upgrade the meta-package without upgrading the components, leading to a version mismatch.

specs/008-packaging-install.md
```diff
@@ -0,0 +1,439 @@
...
+| `/usr/bin/infmon-frontend`                      | root           | 0755 | Daemon binary.                                            |
+| `/etc/infmon/frontend.conf`                     | root           | 0644 | Conffile.                                                 |
+| `/etc/infmon/flow_defs.d/`                      | root           | 0755 | Operator drops flow_def YAML here (consumed per spec 002).|
+| `/lib/systemd/system/infmon-frontend.service`   | root           | 0644 | Unit file (§6).                                           |
+| `/var/log/infmon/`                              | infmon:infmon  | 0750 | Created by postinst; rotated by `/etc/logrotate.d/infmon`.|
+| `/etc/logrotate.d/infmon`                       | root           | 0644 | Daily rotation, 14 days, compressed.                      |
+| `/var/lib/infmon/`                              | infmon:infmon  | 0750 | Reserved for future state (currently empty).              |
+| `/usr/share/doc/infmon-frontend/changelog.gz`   | root           | 0644 |                                                           |
+
```
banidoru, line 178: The `postinst` script creates the `infmon` user and group. If multiple packages are installed in parallel, they might all try to create the user and group simultaneously, which could cause the script to fail. This creation should be atomic or handled by one package only.

specs/008-packaging-install.md
```diff
@@ -0,0 +1,439 @@
...
+**`apt remove infmon-frontend infmon-backend infmon-cli`**
+
+- Stops and disables `infmon-frontend.service`.
+- Removes the plugin `.so` from `/usr/lib/vpp_plugins/` (so the next
+  VPP restart will not load it).
+- Removes binaries under `/usr/bin/` and the systemd unit file.
+- **Leaves untouched:** `/etc/infmon/`, `/var/log/infmon/`,
+  `/var/lib/infmon/`, the `infmon` system user, and **VPP itself**.
+
+**`apt purge infmon-frontend infmon-backend infmon-cli`**
```
banidoru, line 363: The `debian` directory layout is missing the `debian/source/format` file, which is required by `dpkg-source` to specify the source package format (e.g., `3.0 (quilt)`).

banidoru (changes requested):
- The `infmon` meta-package lacks versioned dependencies on its components.
- The `postinst` script has a potential race condition when creating the `infmon` user and directories.
- The spec omits `debian/source/format`.

specs/008-packaging-install.md
```diff
@@ -0,0 +1,439 @@
...
+
+Notes:
+
+- `Type=notify`: the frontend calls `sd_notify(READY=1)` once it has
+  successfully attached to the stats segment and the binary-API socket.
+  This guarantees that `systemctl start infmon-frontend` blocks until
+  the service is actually consuming counters, not merely forked.
+- `PartOf=vpp.service` means a `systemctl restart vpp` automatically
+  restarts the frontend (which is correct: a VPP restart invalidates
+  the stats-segment mappings).
```
banidoru: **Empty `CapabilityBoundingSet=` may break hugepages access.**

The unit drops all capabilities (`CapabilityBoundingSet=` and `AmbientCapabilities=` both empty), but `ReadOnlyPaths=/dev/hugepages` on line 224 implies the frontend needs to `mmap` hugepages-backed shared memory for the stats segment. On many kernels, accessing `/dev/hugepages` requires `CAP_DAC_OVERRIDE` or at minimum the file permissions must align exactly with the `vpp` supplementary group. If VPP's stats segment file is created with restrictive permissions (e.g., 0600 owned by root), group membership alone won't suffice.

Consider explicitly listing required capabilities or documenting the exact permission contract VPP must satisfy on the stats segment file for this to work.

specs/008-packaging-install.md
```diff
@@ -0,0 +1,439 @@
...
+script: silently editing `/etc/apt/sources.list.d/` on install is hostile
+and breaks the principle of least surprise. The dependency declaration
+combined with documented bootstrap is sufficient.
+
+### 4.3 Behaviour when VPP is already managed locally
+
+Many BF-3 users build VPP from source (e.g. NVIDIA's BF reference image
+ships its own VPP). In that case the operator passes
+`--no-install-recommends` and uses `dpkg --force-depends` or installs an
+equivalent locally-built VPP `.deb`. We document this path in
```
banidoru: **`dpkg --force-depends` recommendation needs stronger guardrails.**

Suggesting `dpkg --force-depends` in documentation is risky — it bypasses dependency resolution entirely and can leave the system in a broken state that `apt` refuses to fix without manual intervention (`apt --fix-broken install` may try to remove the force-installed package). At minimum, this section should:

1. Warn that `--force-depends` puts the package in a `dpkg` half-configured state from `apt`'s perspective.
2. Recommend `equivs` as the cleaner alternative (create a dummy `vpp` package that satisfies the dependency).
3. Note that `apt-mark hold` on the force-installed package prevents `apt` from "fixing" it away.

specs/008-packaging-install.md
```diff
@@ -0,0 +1,439 @@
...
+
+The "leave VPP alone" rule is non-negotiable: VPP on a DPU may be
+carrying production traffic for unrelated services (e.g. an SDN
+data-plane). Removing InFMon must never put that traffic at risk.
+
+### 7.4 Maintainer-script summary
+
+| Script        | Package           | Action                                                                                                                          |
+| ------------- | ----------------- | ------------------------------------------------------------------------------------------------------------------------------- |
+| `postinst`    | `infmon-backend`  | Print NOTICE about `systemctl restart vpp`. No file changes beyond what `dpkg` already did.                                     |
```
banidoru, line 389: **`--buildsystem=cmake+cargo` is not a real debhelper buildsystem.**

Debhelper ships `cmake`, `makefile`, `meson`, etc., but there is no `cmake+cargo` buildsystem. This will fail with `dh: unable to load addon cmake+cargo`. The `debian/rules` file will need explicit `override_dh_auto_build` / `override_dh_auto_install` targets to invoke cmake for the backend and cargo for the frontend/CLI separately. Worth clarifying this in the spec so the implementer doesn't discover it at build time.

specs/008-packaging-install.md
```diff
@@ -0,0 +1,439 @@
...
+not a service. Its lifecycle is VPP's.
+
+## 7. Install / upgrade / uninstall flows
+
+### 7.1 Install
+
+```
+apt install infmon
+  └─ resolves vpp >= 24.10 from fd.io repo (or local)
+  └─ unpacks infmon-backend → /usr/lib/vpp_plugins/infmon.so
```
banidoru, line 298: **Frontend/backend version skew during upgrade is unaddressed.**

The spec says `dh_installsystemd` restarts the frontend on upgrade, but the backend plugin isn't reloaded until the operator manually restarts VPP. This means the upgraded frontend may run against the *old* backend plugin for an unbounded period. If the binary-API messages changed between versions (which §9.1 says triggers a MAJOR bump), the frontend could hit deserialization errors or silent data corruption.

The spec should either:
1. Require the frontend to negotiate/verify the backend API version on connect and fail clearly if mismatched, or
2. Document the supported version-skew matrix (e.g., frontend N works with backend N and N-1).

specs/008-packaging-install.md
```diff
@@ -0,0 +1,439 @@
...
+  vpp-dev (>= 24.10),
+  libvppinfra-dev (>= 24.10),
+  cargo (>= 1.75),
+  rustc (>= 1.75),
+  clang,
+```
+
+Spec 001's CI builds the source package with `dpkg-buildpackage -us -uc
+-b -aarm64` on an aarch64 runner (or `sbuild` in a BF-3 chroot) and
+publishes the resulting `.deb`s as workflow artefacts.
```
banidoru, line 482: **"CLI argument grammar that the help text marks as stable" — how?**

This references a stability-marking mechanism for CLI arguments, but no such mechanism is defined here or in spec 007. How does the help text distinguish stable from unstable arguments? A `[UNSTABLE]` prefix? A separate section? Without this, the MAJOR-bump criterion is unenforceable — any CLI change could be argued either way.

banidoru (commented):
- `CapabilityBoundingSet=` empty may conflict with hugepages access needs despite `SupplementaryGroups=vpp`.
- `dpkg --force-depends` recommendation in §4.3 is risky — should come with stronger caveats.
- No buildsystem `cmake+cargo` exists in debhelper — `debian/rules` override needs clarification.
- Frontend upgrade restart + old backend plugin compatibility gap not addressed.
- CLI stability marking mechanism referenced but undefined.

banidoru (approved):
All reviewers approved. LGTM.

r12f:

specs/008-packaging-install.md
banidoru: Still open — the new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields). This concern is not addressed.

specs/008-packaging-install.md
banidoru: Still open — the new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields). This concern is not addressed.

specs/008-packaging-install.md
banidoru: Still open — the new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields). This concern is not addressed.

specs/008-packaging-install.md
banidoru: Still open — the new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields). This concern is not addressed.

specs/008-packaging-install.md
banidoru: Still open — the new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields). This concern is not addressed.

specs/008-packaging-install.md
banidoru: Still open — the new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields). This concern is not addressed.

specs/008-packaging-install.md
banidoru: Still open — the new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields). This concern is not addressed.

specs/008-packaging-install.md
banidoru: Still open — the new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields). This concern is not addressed.

specs/008-packaging-install.md
banidoru: Still open — the new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields). This concern is not addressed.

specs/008-packaging-install.md
banidoru: Still open — the new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields). This concern is not addressed.

specs/008-packaging-install.md
banidoru: Still open — the new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields). This concern is not addressed.

specs/008-packaging-install.md
banidoru: Still open — the new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields). This concern is not addressed.

specs/008-packaging-install.md
banidoru: Still open — the new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields). This concern is not addressed.

specs/008-packaging-install.md
banidoru: Still open — the new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields). This concern is not addressed.

banidoru (changes requested):
- **None of the 14 prior review concerns are addressed** — new commits only changed header metadata (added version history table, removed Status/Owner/Reviewer fields).
- All open threads remain valid and unresolved.
- No new substantive issues introduced by the metadata-only change.
- Please address the prior feedback before requesting another review.
- Key outstanding items: inter-package version constraints, `PartOf=vpp.service` contradiction, `cmake+cargo` buildsystem, hugepages capability/mount issues, purge directory conflict.

banidoru:
All previous review comments remain unaddressed in the latest push. This update only changes document metadata.

banidoru (commented):
Re-review (iteration 2) — new commits only changed header metadata (version history table, removed Status/Owner/Reviewer fields).

- **All 14 prior review threads remain unaddressed** — no substantive spec changes in this push
- No new issues introduced by the metadata-only diff
- Key open concerns: `PartOf=vpp.service` misconfiguration, `cmake+cargo` non-existent buildsystem, missing inter-package version constraints, cross-package purge directory conflict, `CapabilityBoundingSet` vs hugepages access
- Recommend addressing the prior feedback before next iteration

specs/008-packaging-install.md
r12f: Agreed — `PartOf=vpp.service` is silently a no-op for renamed VPP units. Removed the unconditional `PartOf=`; the unit now relies on the frontend's own retry loop for VPP restarts and documents an opt-in drop-in (`/etc/systemd/system/infmon-frontend.service.d/partof.conf`) for operators on the canonical `vpp.service`. See unit comment + §6 notes.

specs/008-packaging-install.md
r12f: Right — dropped `debian/compat` from the §8 layout and added a note that compat 13 is declared via `debhelper-compat (= 13)` in `Build-Depends` to avoid the lintian warning.

specs/008-packaging-install.md
r12f: Good catch. Rewrote §7.3 purge and §7.4 table: each package's `postrm` removes only its own conffiles (dpkg handles them) and ends with `rmdir --ignore-fail-on-non-empty /etc/infmon`, so the shared dir survives until the last InFMon package is purged.

specs/008-packaging-install.md
r12f: Acknowledged — kernel-VMA semantics here are real. Kept `ReadOnlyPaths=` as the default but added a normative note in the unit + §6 that the implementation PR MUST validate mmap() under this confinement on Ubuntu 22.04 HWE for BF-3 and switch to `ReadWritePaths=/dev/hugepages` if it fails (NoNewPrivileges + empty caps make the security delta small).

specs/008-packaging-install.md
r12f: Agreed. Added explicit `Depends: infmon-backend (= ${binary:Version})` from `infmon-frontend`, and version-locked the meta-package's deps. CLI is intentionally left independent (cross-version tolerated via REST). See §3.

specs/008-packaging-install.md
r12f: Added a §4.2 note: operators should verify the fd.io GPG key fingerprint out-of-band, and `debian/README.Debian` will reproduce the fingerprint and the `gpg --show-keys` command.

specs/008-packaging-install.md
r12f: Done — meta-package now declares `Depends: infmon-backend (= ${binary:Version}), infmon-frontend (= ${binary:Version}), infmon-cli (= ${binary:Version})` in §3.

specs/008-packaging-install.md
r12f: Clarified in §5.2: only `infmon-frontend.postinst` creates the user (backend/CLI postinsts don't touch the user db), `adduser --system` is idempotent, and dpkg serialises configure phases so even `apt install infmon` (meta) can't race against itself.

specs/008-packaging-install.md
r12f: Added — `debian/source/format` (containing `3.0 (quilt)`) is now in the §8 layout with a note that it's required by `dpkg-source`.

specs/008-packaging-install.md
r12f: Tied this to the same review thread on hugepages/caps. Added a normative implementation-PR check in §6: verify the VPP stats-segment file mode/group on the target image; if VPP creates it 0600 root:root, either reconfigure VPP's stats-segment perms or add CAP_DAC_READ_SEARCH back. Default assumption is 0640 root:vpp + `SupplementaryGroups=vpp`.

specs/008-packaging-install.md
r12f: Agreed. §4.3 rewritten: `equivs` (or a locally-built VPP `.deb` providing the dependency names) is now the recommended path; `dpkg --force-depends` is explicitly called out as not first-class because it leaves dpkg half-configured and `apt --fix-broken install` may try to remove the package.

specs/008-packaging-install.md
r12f: You're right — there is no `cmake+cargo` debhelper buildsystem. Replaced the §8 one-liner with explicit `debian/rules` showing `dh $@` plus `override_dh_auto_configure/build/install` driving cmake (backend) and cargo (frontend/CLI) by hand. Removes the build-time surprise.

specs/008-packaging-install.md
r12f: Addressed in §7.2: (1) §3's version-locked `Depends: infmon-backend (= ${binary:Version})` keeps on-disk packages in lockstep, and (2) the frontend MUST query the backend API version on each binary-API connect and refuse to stream on MAJOR mismatch, parking in `status=waiting_for_vpp_restart` and surfacing via `sd_notify(STATUS=...)`. Wire-level details remain owned by spec 004 / 005.

specs/008-packaging-install.md
r12f: Made it mechanically enforceable in §9.1: stable CLI args are those whose `clap` definition does NOT carry `#[arg(hide = true)]` or a doc-comment `UNSTABLE:` prefix; `infmonctl --help` lists unstable flags in a separate section. Spec 007 will be updated to make this normative when this spec lands.

banidoru (approved):
- All 14 prior review concerns addressed in this iteration — version-locked deps, PartOf removal with drop-in, debian/compat cleanup, cross-package purge fix, equivs recommendation, explicit cmake+cargo override targets, and more.
- The new diff is clean: changes are well-scoped to the issues raised, no regressions introduced.
- Minor: file ends without a trailing newline (`\ No newline at end of file`) — not a blocker but worth fixing for consistency.
- Overall spec quality is solid; the normative implementation-PR checks (hugepages mmap validation, VPP stats-segment permissions) are well-placed.
- LGTM.

banidoru (approved):
- All 14 prior review threads addressed and resolved — version-locked deps, PartOf removal, postrm cross-pkg fix, compat file removal, source/format addition, equivs recommendation, GPG verification, hugepages validation notes, version-skew handling, CLI stability marking, and explicit debian/rules all look correct.
- The explicit `override_dh_auto_*` targets are the right pattern for mixed cmake+cargo builds; the `librsvg` reference is apt.
- Minor: file ends without a trailing newline (last line of diff). Non-blocking.
- Version history table properly attributes authorship and documents changes per iteration.
- LGTM — no new issues found in this iteration.

banidoru (approved):
All reviewers approved. LGTM.

banidoru (approved):
- All 14 prior review threads have been addressed and resolved — fixes look correct.
- New diff (c65956b..1d34c89) is cosmetic only: version history table consolidation and metadata field reformatting.
- LGTM.

banidoru (approved):
- All prior review comments have been addressed.
- The `PartOf=` and `compat` file issues are resolved.
- Inter-package dependencies are now correctly version-locked.
- The build rules and systemd unit file have been corrected.
- Documentation has been updated to reflect best practices for GPG verification and dependency management.

banidoru (approved):
- All 14 prior review threads addressed and resolved by author — fixes look correct
- New commits are metadata-only: version history table consolidation, field→bullet-point reformatting
- No substantive spec content changes in the new diff — no new issues
- LGTM

banidoru (approved):
All reviewers approved. LGTM.

PR #11, opened 2026-04-18
specs: 006 — OTLP exporter (v1)

r12f:
Drafts `specs/006-exporter-otlp.md`, the v1 OpenTelemetry exporter contract for InFMon.

Tracking: **DPU-12** (parent: DPU-4 — EPIC InFMon)
Depends on: 000-overview, 002-flow-tracking-model

## What's in the spec

- **Signal type.** Metrics only. Per-bucket: `infmon.flow.packets` /
  `infmon.flow.bytes` as cumulative monotonic Sums, `infmon.flow.last_seen`
  as a Gauge. `first_seen_ns` is folded into `start_time_unix_nano`.
- **Per-tracker observability** (Spec 002 §8) shipped on the same stream:
  buckets gauge, evictions/drops/packets/bytes counters, scoped by
  `{tracker}` (and `reason` where relevant).
- **Attribute mapping.** Reserved `flow.*` namespace, one-to-one with the
  tracker's ordered field list. IPs as canonical text, dscp/ip_proto as
  ints. Trackers only emit attributes for fields they're keyed on.
- **Resource attributes.** `service.name`/`namespace`/`version`/`instance.id`,
  `host.name`/`host.id`/`host.arch`, plus operator-set
  `infmon.dpu.id` / `infmon.dpu.platform`. Stable for the process lifetime.
- **Endpoint config.** gRPC default (4317), http/protobuf opt-in (4318),
  optional mTLS, gzip, headers, full TOML block + per-field defaults +
  all-or-nothing validation rules (incl. `export_timeout < export_interval`).
- **Batching, retry, timeout.** Single periodic tick, bounded send queue,
  oldest-batch-drop on overflow, jittered exponential backoff (3 retries,
  cap 30s), per-batch total bound of `4 × export_interval`, graceful
  shutdown drains.
- **Cardinality.** Operator-facing guidance (eviction rate as the load
  signal, prefer narrow trackers) plus hard exporter-side caps:
  `max_export_points_per_tick = 2_000_000`, 256B attribute-value cap,
  no free-form per-bucket attributes, full self-observability metric set.
- **Test plan** (unit / integration with a tonic OTLP stub / E2E via
  `otelcol-contrib`) and a perf guideline (1M buckets × 8 trackers ≤ 2s
  on BF-3).
- **Open questions** documented (host.id fallback, delta temporality,
  required vs default `infmon.dpu.id`, exemplars).

## Review focus

- Is the cumulative-only choice acceptable for the first wave of
  collectors we expect (Q2)?
- Is the `flow.*` attribute namespace agreeable, or should we align
  with an existing OTel semantic-conventions group?
- Cardinality safeguards — are the defaults sane for the BF-3 footprint?

Reviewer: @banidoru

### DIFF
--- a/specs/006-exporter-otlp.md
+++ b/specs/006-exporter-otlp.md
@@ -1,0 +1,406 @@
+# Spec 006 — OTLP exporter (v1)
+
+Status: Draft
+Tracking issue: DPU-12 (project InFMon)
+Parent epic: DPU-4 (EPIC: InFMon — flow telemetry service on BF-3)
+Depends on: 000-overview (system overview),
+            002-flow-tracking-model (defines trackers, keys, buckets,
+            and the per-tracker observability counters this exporter ships)
+Related: 004-backend-architecture (owns the snapshot mechanism the
+         exporter consumes), 005-frontend (hosts the exporter process),
+         007-cli (`infmon-cli exporter ...` surface)
+
+## 1. Purpose
+
+Define the **v1 OpenTelemetry (OTLP) exporter** for InFMon: how flow-tracker
+state held by the backend (Spec 002) is turned into OTLP metrics and shipped
+off-DPU to a collector.
+
+The OTLP exporter is the **only** standard-format exporter in v1. It is the
+sole production sink for flow data; everything else (CLI `flow show`, ad-hoc
+JSON dumps) is for debugging. Getting this contract right means downstream
+collectors, dashboards, and alerting can be built without waiting for InFMon
+to stabilise further.
+
+## 2. Concepts
+
+### 2.1 Exporter instance
+
+A single exporter instance is configured per `infmon-frontend` process.
+Multiple destinations are not supported in v1 — operators who need fan-out
+should run an OpenTelemetry Collector adjacent to the DPU.
+
+### 2.2 Source of truth
+
+The exporter never touches the data plane. It consumes the **snapshot**
+surface owned by the backend (Spec 004) on a fixed cadence and translates
+each tracker's live buckets into OTLP data points.
+
+### 2.3 Wire shape
+
+OTLP / metrics. v1 ships the gRPC transport by default and accepts
+http/protobuf as a configuration switch. JSON-over-HTTP is **not** supported
+(noisier on the wire, no observed collector that requires it).
+
+## 3. Signal type and metric model
+
+### 3.1 Signal: metrics
+
+InFMon exports **metrics** only in v1. Logs and traces are out of scope —
+flow data is intrinsically aggregate counter state, and shoehorning it
+through `Logs` would lose the collector's native aggregation pipeline.
+
+### 3.2 Per-bucket data points
+
+Each live bucket in each tracker becomes a fixed set of OTLP data points,
+all sharing the bucket's attribute set (§4):
+
+| OTLP metric name             | Instrument | Unit  | Source field      | Notes                                     |
+|------------------------------|------------|-------|-------------------|-------------------------------------------|
+| `infmon.flow.packets`        | Sum (cumulative, monotonic) | `{packets}` | `bucket.packets` | Total packets attributed to this bucket. |
+| `infmon.flow.bytes`          | Sum (cumulative, monotonic) | `By`        | `bucket.bytes`   | Total bytes (L3 length per Spec 003).    |
+| `infmon.flow.last_seen`      | Gauge      | `ns`  | `bucket.last_seen_ns` | Wall-clock ns of the most recent packet that updated this bucket. |
+
+`first_seen_ns` is **not** exported as its own data point. It is folded into
+the data-point `start_time_unix_nano` of the two `Sum` metrics, which is the
+OTLP-native way to express "this counter started accumulating at time T".
+
+### 3.3 Aggregation temporality
+
+`Cumulative` for both `Sum` metrics. Rationale:
+
+- Cumulative is the OTLP default and what most collector pipelines (Prometheus
+  remote-write, Mimir, Cortex) prefer.
+- The backend owns the bucket lifetime; the exporter is stateless. Cumulative
+  + a stable `start_time_unix_nano` lets a collector compute deltas without
+  the exporter having to track previous-export state.
+- On bucket eviction (Spec 002 §6) the counter is gone. Receivers that
+  compute deltas will see a counter reset; OTLP Sum semantics handle that
+  correctly when `start_time_unix_nano` advances.
+
+### 3.4 Per-tracker observability points
+
+The counters defined in Spec 002 §8 are exported alongside the per-bucket
+data points. They are scoped to the tracker, not the key:
+
+| OTLP metric name                  | Instrument | Unit         | Source                                    |
+|-----------------------------------|------------|--------------|-------------------------------------------|
+| `infmon.tracker.buckets`          | Gauge      | `{buckets}`  | `infmon_tracker_buckets`                  |
+| `infmon.tracker.evictions`        | Sum (cum.) | `{evictions}`| `infmon_tracker_evictions_total`          |
+| `infmon.tracker.drops`            | Sum (cum.) | `{drops}`    | `infmon_tracker_drops_total{reason}`      |
+| `infmon.tracker.packets`          | Sum (cum.) | `{packets}`  | `infmon_tracker_packets_total`            |
+| `infmon.tracker.bytes`            | Sum (cum.) | `By`         | `infmon_tracker_bytes_total`              |
+
+Their attribute set is `{tracker}` (plus `reason` for `drops`), nothing else.
+
+## 4. Attribute mapping (per-bucket points)
+
+Each per-bucket data point carries a fixed attribute set derived from
+the tracker's field list (Spec 002 §3) plus a `tracker` label:
+
+| Attribute         | Type        | Source                                  |
+|-------------------|-------------|-----------------------------------------|
+| `tracker`         | string      | tracker `name` (Spec 002 §2.1)          |
+| `flow.src_ip`     | string      | `enc(src_ip)` rendered as canonical text (IPv4 if v4-mapped, else v6) |
+| `flow.dst_ip`     | string      | same rule for `dst_ip`                  |
+| `flow.ip_proto`   | int         | numeric `ip_proto`, 0–255               |
+| `flow.dscp`       | int         | numeric `dscp`, 0–63                    |
+
+Rules:
+
+1. The exporter MUST emit **only** attributes that correspond to fields
+   actually present in the tracker's `ordered_field_list`. A tracker keyed
+   on `[dscp]` produces points with attributes `{tracker, flow.dscp}` —
+   nothing else.
+2. The mapping is fixed (§9 reserves the `flow.*` namespace for future
+   fields). Attribute names MUST NOT vary between exports.
+3. IP addresses are rendered as text (canonical IPv4 dotted-quad if the
+   16-byte key is v4-mapped, RFC 5952 lowercase v6 otherwise), not as raw
+   bytes. Text form is what every viable collector / TSDB indexes natively.
+4. `dscp` and `ip_proto` are emitted as integers, not numeric strings.
+
+### 4.1 Why prefix with `flow.`
+
+Reserves a namespace so future signals (e.g. `link.*` for interface stats,
+`parser.*` for the upstream parser) do not collide. Keeps the collector-side
+schema obvious at a glance.
+
+### 4.2 Attribute order
+
+OTLP attribute sets are unordered by spec; the exporter MAY emit them in
+any order. Implementations SHOULD sort them by attribute key for stable
+on-the-wire bytes (helps gRPC compression and snapshot diffs in tests).
+
+## 5. Resource attributes
+
+Every export carries a `Resource` describing the producer. v1 set:
+
+| Attribute                       | Source                                        | Notes |
+|---------------------------------|-----------------------------------------------|-------|
+| `service.name`                  | constant `"infmon-frontend"`                  | OTel-mandated. |
+| `service.namespace`             | constant `"infmon"`                           | Lets a multi-tenant collector group all InFMon producers. |
+| `service.version`               | InFMon build version (`CARGO_PKG_VERSION` of `infmon-frontend`, or the package version baked in at build time) | The single "infmon version" mentioned in the issue. |
+| `service.instance.id`           | stable per-process UUID, generated at frontend startup, persisted in `/var/lib/infmon/instance_id` | Distinguishes restarts from re-deploys. |
+| `host.name`                     | `gethostname(2)` at startup                   | Human-readable host identity. |
+| `host.id`                       | `/etc/machine-id` if readable, else absent    | Stable across reboots; omit silently if unavailable. |
+| `host.arch`                     | constant `"arm64"` for the v1 .deb build      | Matches Spec 000 §"Build & Release Model". |
+| `infmon.dpu.id`                 | configured value (see §6); falls back to `host.name` if unset | The "dpu" attribute the issue calls for. Free-form short string operators set per-DPU. |
+| `infmon.dpu.platform`           | configured value (e.g. `"bluefield-3"`)        | Optional; omitted if unset. |
+
+`host.*` attributes follow OpenTelemetry [semantic conventions for host].
+`service.*` likewise. `infmon.*` is our own namespace and is documented here.
+
+Resource attributes are computed once at frontend startup and reused on
+every export; they MUST NOT vary between exports of a single process
+lifetime.
+
+## 6. Endpoint and transport configuration
+
+The exporter is configured in the same `infmon-frontend` config file
+(format owned by Spec 005). The OTLP block:
+
+### 6.1 TOML
+
+```toml
+[exporter.otlp]
+enabled            = true
+protocol           = "grpc"            # "grpc" | "http_protobuf"
+endpoint           = "otel-collector.infra.local:4317"
+                                       # gRPC default 4317, http/protobuf default 4318
+insecure           = false             # true => plaintext (TCP, not mTLS)
+ca_file            = "/etc/infmon/tls/ca.pem"      # optional; system trust if unset
+cert_file          = "/etc/infmon/tls/client.pem"  # optional client mTLS
+key_file           = "/etc/infmon/tls/client.key"  # optional client mTLS
+compression        = "gzip"            # "none" | "gzip"
+export_interval    = "10s"             # how often to take a snapshot and ship
+export_timeout     = "5s"              # per-request timeout
+max_batch_points   = 8192              # see §7
+queue_size         = 4                 # see §7
+
+[exporter.otlp.headers]
+# arbitrary key=value, sent on every request (gRPC metadata or HTTP headers)
+"x-tenant" = "dpu-east-1"
+
+[exporter.otlp.resource]
+# operator-set resource attributes, merged on top of §5 defaults
+"infmon.dpu.id"       = "bf3-rack17-u4"
+"infmon.dpu.platform" = "bluefield-3"
+```
+
+### 6.2 Defaults
+
+Field-by-field defaults if absent:
+
+- `enabled = false` — the exporter is opt-in.
+- `protocol = "grpc"`.
+- `endpoint` — required when `enabled = true`; no default.
+- `insecure = false`.
+- `compression = "gzip"`.
+- `export_interval = "10s"`.
+- `export_timeout = "5s"`.
+- `max_batch_points = 8192`.
+- `queue_size = 4`.
+- `headers` and `resource` default to empty.
+
+### 6.3 Validation
+
+Reject the config (and refuse to start) if:
+
+1. `enabled = true` and `endpoint` is missing or unparseable.
+2. `protocol` is not in `{"grpc", "http_protobuf"}`.
+3. `compression` is not in `{"none", "gzip"}`.
+4. Any of `cert_file` / `key_file` is set without the other.
+5. Any TLS file path is set but unreadable at startup.
+6. `export_interval`, `export_timeout` are not positive durations.
+7. `export_timeout >= export_interval` (would let backpressure run away).
+8. `max_batch_points <= 0`.
+9. `queue_size <= 0`.
+
+Configuration validation is all-or-nothing, mirroring Spec 002 §5.3.
+
+## 7. Batching, retry, timeout
+
+### 7.1 Tick
+
+The exporter runs on a single periodic tick of period `export_interval`.
+On each tick:
+
+1. Ask the backend for a fresh snapshot (Spec 004 surface). The snapshot
+   is read-only and lock-free w.r.t. the data plane.
+2. Walk every tracker. For each live bucket emit the data points defined
+   in §3.2. Append the per-tracker points (§3.4).
+3. Group data points into OTLP `ResourceMetrics` → `ScopeMetrics` →
+   `Metric` according to standard OTLP shape (one `ResourceMetrics` per
+   process; one `ScopeMetrics` per InstrumentationScope `infmon`; one
+   `Metric` per metric name).
+4. Slice into batches of at most `max_batch_points` data points and hand
+   them to the send queue.
+
+If a tick is still busy when the next tick fires, the new tick is **dropped**
+and `infmon_exporter_ticks_dropped_total` is incremented. The exporter
+never queues unbounded snapshots.
+
+### 7.2 Send queue
+
+A bounded in-memory queue of size `queue_size` batches sits between the
+tick loop and the network sender. If the queue is full when a new batch
+arrives, the **oldest** batch is dropped (the freshest cumulative data
+supersedes it anyway) and `infmon_exporter_batches_dropped_total` is
+incremented.
+
+### 7.3 Retry
+
+Per-batch retry policy:
+
+- Retry only on transport-level errors and OTLP `RetryableError` /
+  HTTP 429 / HTTP 5xx.
+- Exponential backoff with full jitter: `min(cap, base * 2^attempt) *
+  rand(0, 1)`, with `base = 1s`, `cap = 30s`.
+- Maximum 3 retries per batch. After the final failure the batch is
+  dropped and `infmon_exporter_batches_failed_total{reason}` is bumped.
+- The per-request timeout is `export_timeout`. The total time spent on
+  a single batch (including retries) is bounded by
+  `4 * export_interval` to keep the queue head from rotting.
+
+Non-retryable errors (auth failure, schema rejection, HTTP 4xx other than
+429) drop the batch immediately and bump
+`infmon_exporter_batches_failed_total{reason="non_retryable"}`. The
+exporter does **not** poison-pill subsequent batches.
+
+### 7.4 Shutdown
+
+On graceful shutdown the exporter MUST:
+
+1. Stop the tick loop.
+2. Drain the send queue, with a hard ceiling of one `export_timeout` per
+   batch and at most `queue_size` batches.
+3. Close the gRPC channel / HTTP client.
+
+A signal-driven hard exit MAY skip drain. The cumulative semantics
+guarantee a restart will resume cleanly with a fresh
+`start_time_unix_nano`.
+
+## 8. Cardinality guidance and safeguards
+
+OTLP cardinality is the single sharpest foot-gun in this exporter, since
+each live bucket becomes ≥ 3 unique time series in the receiving TSDB.
+
+### 8.1 Operator-facing guidance
+
+Document in the `infmon-frontend` operator README and CLI `--help`:
+
+- A tracker's worst-case time-series contribution to the collector is
+  `max_keys × distinct_data_points_per_bucket`. With v1's 3 per-bucket
+  metrics, a tracker with `max_keys = 1_048_576` can land 3 Mi series in
+  the receiving TSDB **per scrape interval**.
+- The four v1 fields are *all* high-cardinality except `dscp`. A tracker
+  keyed on `[src_ip, dst_ip, ip_proto, dscp]` is appropriate for an
+  always-on production deployment **only** if the receiving TSDB is sized
+  for it.
+- Prefer narrower trackers (e.g. `[src_ip]` for talker tables, `[dscp]`
+  for class-of-service rollups) for default-on deployments.
+- The growth signal is `infmon_tracker_evictions_total`. A non-zero,
+  growing eviction rate means cardinality exceeds the configured budget;
+  either grow `max_keys` or narrow the field list.
+
+### 8.2 Exporter-side safeguards
+
+The exporter MUST enforce these, independently of the backend:
+
+1. **Hard per-export point cap** — `max_export_points_per_tick`
+   (compile-time default `2_000_000`). If a tick would emit more, the
+   exporter ships the first cap-many points sorted by tracker name then
+   bucket recency and **drops the rest**, bumping
+   `infmon_exporter_points_dropped_total{reason="export_cap"}`. This
+   protects the collector from a backend misconfiguration.
+2. **Per-attribute length cap** — string attribute values are truncated
+   to 256 bytes (UTF-8 safe) with a trailing `…`. Bumps
+   `infmon_exporter_attrs_truncated_total`.
+3. **Drop empty trackers** — a tracker with zero live buckets emits its
+   per-tracker points (§3.4) but no per-bucket points. This is normal,
+   not an error.
+4. **No metadata explosion** — the exporter MUST NOT add free-form
+   per-bucket attributes beyond §4. There is no escape hatch in v1.
+
+### 8.3 Self-observability
+
+The exporter exports its own health on the same OTLP stream:
+
+| OTLP metric name                              | Instrument | Unit       |
+|-----------------------------------------------|------------|------------|
+| `infmon.exporter.ticks_dropped`               | Sum (cum.) | `{ticks}`  |
+| `infmon.exporter.batches_sent`                | Sum (cum.) | `{batches}`|
+| `infmon.exporter.batches_dropped`             | Sum (cum.) | `{batches}`|
+| `infmon.exporter.batches_failed`              | Sum (cum.) | `{batches}`|
+| `infmon.exporter.points_emitted`              | Sum (cum.) | `{points}` |
+| `infmon.exporter.points_dropped`              | Sum (cum.) | `{points}` |
+| `infmon.exporter.attrs_truncated`             | Sum (cum.) | `{attrs}`  |
+| `infmon.exporter.export_duration`             | Gauge      | `s`        |
+| `infmon.exporter.queue_depth`                 | Gauge      | `{batches}`|
+
+These carry the §5 resource attributes and a `reason` attribute where
+relevant; they do **not** carry tracker or bucket attributes.
+
+## 9. Future extension hooks (non-normative)
+
+- **L4 attributes.** When Spec 002 §9.1 adds `src_port`, `dst_port`,
+  `tcp_flags`, the exporter gains `flow.src_port`, `flow.dst_port`,
+  `flow.tcp_flags` under the same `flow.*` namespace, no schema break.
+- **RoCE attributes.** `flow.roce_dest_qp`, `flow.roce_opcode` etc.,
+  one-to-one with Spec 002 §9.2 fields.
+- **Histograms.** RTT samples (Spec 000 component map mentions them as
+  a future bucket field) will export as OTLP `Histogram`. The instrument
+  type column in §3.2 / §3.4 is precisely why instrument is per-metric,
+  not global.
+- **Multiple destinations.** v1 explicitly forbids fan-out; v2 can add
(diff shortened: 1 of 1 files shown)


specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,406 @@
...
+
+Each live bucket in each tracker becomes a fixed set of OTLP data points,
+all sharing the bucket's attribute set (§4):
+
+| OTLP metric name             | Instrument | Unit  | Source field      | Notes                                     |
+|------------------------------|------------|-------|-------------------|-------------------------------------------|
+| `infmon.flow.packets`        | Sum (cumulative, monotonic) | `{packets}` | `bucket.packets` | Total packets attributed to this bucket. |
+| `infmon.flow.bytes`          | Sum (cumulative, monotonic) | `By`        | `bucket.bytes`   | Total bytes (L3 length per Spec 003).    |
+| `infmon.flow.last_seen`      | Gauge      | `ns`  | `bucket.last_seen_ns` | Wall-clock ns of the most recent packet that updated this bucket. |
+
```
banidoru, line 82: The `last_seen` gauge uses unit `ns` (nanoseconds since epoch), which will be a very large integer (~19 digits). Many TSDBs and dashboards have trouble with nanosecond-epoch gauges (float64 precision loss, Grafana rendering). Consider:
- Using seconds (with fractional) as the unit, consistent with OTel's own timestamp conventions
- Or documenting explicitly that this is wall-clock nanoseconds and noting the float64 precision risk for backends that store gauge values as doubles

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,406 @@
...
+| `service.version`               | InFMon build version (`CARGO_PKG_VERSION` of `infmon-frontend`, or the package version baked in at build time) | The single "infmon version" mentioned in the issue. |
+| `service.instance.id`           | stable per-process UUID, generated at frontend startup, persisted in `/var/lib/infmon/instance_id` | Distinguishes restarts from re-deploys. |
+| `host.name`                     | `gethostname(2)` at startup                   | Human-readable host identity. |
+| `host.id`                       | `/etc/machine-id` if readable, else absent    | Stable across reboots; omit silently if unavailable. |
+| `host.arch`                     | constant `"arm64"` for the v1 .deb build      | Matches Spec 000 §"Build & Release Model". |
+| `infmon.dpu.id`                 | configured value (see §6); falls back to `host.name` if unset | The "dpu" attribute the issue calls for. Free-form short string operators set per-DPU. |
+| `infmon.dpu.platform`           | configured value (e.g. `"bluefield-3"`)        | Optional; omitted if unset. |
+
+`host.*` attributes follow OpenTelemetry [semantic conventions for host].
```
banidoru, line 188: The `host.arch` is hardcoded to `"arm64"`. If InFMon is ever built for x86 dev/test (even unofficially), this becomes a lie. Safer to derive from `std::env::consts::ARCH` at runtime — trivial in Rust and future-proof.

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,406 @@
...
+
+### 6.3 Validation
+
+Reject the config (and refuse to start) if:
+
+1. `enabled = true` and `endpoint` is missing or unparseable.
+2. `protocol` is not in `{"grpc", "http_protobuf"}`.
+3. `compression` is not in `{"none", "gzip"}`.
+4. Any of `cert_file` / `key_file` is set without the other.
+5. Any TLS file path is set but unreadable at startup.
```
banidoru, line 257: Rule 7 says `export_timeout >= export_interval` is rejected. But the boundary `export_timeout == export_interval` is already problematic — a single request consuming the full interval leaves zero headroom for snapshot acquisition + serialization. Consider tightening to `export_timeout > export_interval / 2` or at minimum document the reasoning for allowing `timeout = interval - 1ns`.

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,406 @@
...
+### 7.2 Send queue
+
+A bounded in-memory queue of size `queue_size` batches sits between the
+tick loop and the network sender. If the queue is full when a new batch
+arrives, the **oldest** batch is dropped (the freshest cumulative data
+supersedes it anyway) and `infmon_exporter_batches_dropped_total` is
+incremented.
+
+### 7.3 Retry
+
```
banidoru, line 315: §7.3 says the per-batch total time is bounded by `4 * export_interval`. With default `export_interval = 10s` and `max_retries = 3`, the total wall time for a single batch could be up to 40s. But the send queue is only 4 batches deep, so a stuck batch blocks the sender for 4 ticks. This interacts poorly with the tick-drop logic in §7.1 — you could silently lose 3 ticks of data while retrying a single batch. Worth documenting this trade-off explicitly, or consider a lower bound like `2 * export_interval`.

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,406 @@
...
+A signal-driven hard exit MAY skip drain. The cumulative semantics
+guarantee a restart will resume cleanly with a fresh
+`start_time_unix_nano`.
+
+## 8. Cardinality guidance and safeguards
+
+OTLP cardinality is the single sharpest foot-gun in this exporter, since
+each live bucket becomes ≥ 3 unique time series in the receiving TSDB.
+
```
banidoru, line 362: The cardinality cap (`max_export_points_per_tick = 2_000_000`) is compile-time only. Given that this is the primary safety valve against collector overload, it should be operator-configurable at runtime (via the TOML config). A compile-time-only default forces a rebuild to adjust — impractical on a deployed DPU fleet.

banidoru (commented):
Solid spec overall — thorough on semantics and failure modes. Key concerns:

- **`last_seen` gauge precision risk**: nanosecond-epoch integers exceed float64 precision in many TSDBs — document or mitigate
- **Cardinality cap is compile-time only**: `max_export_points_per_tick` should be operator-configurable, not require a rebuild
- **Retry/tick interaction**: `4 × export_interval` retry ceiling can silently drop 3 ticks of data while retrying one batch
- **`export_duration` as Gauge loses distribution info**: histogram would be far more actionable for latency debugging
- **`host.arch` hardcoded to arm64**: derive at runtime for dev/test correctness

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,406 @@
...
+### 8.1 Operator-facing guidance
+
+Document in the `infmon-frontend` operator README and CLI `--help`:
+
```
banidoru, line 366: Points are sorted by "tracker name then bucket recency" when the cap is hit. Bucket recency sort requires comparing `last_seen_ns` across potentially millions of buckets. This is O(n log n) on the full set every tick when the cap is exceeded. Consider a cheaper strategy like per-tracker proportional allocation (each tracker gets `cap * tracker_buckets / total_buckets` points) to avoid a global sort.

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,406 @@
...
+The exporter MUST enforce these, independently of the backend:
+
+1. **Hard per-export point cap** — `max_export_points_per_tick`
+   (compile-time default `2_000_000`). If a tick would emit more, the
+   exporter ships the first cap-many points sorted by tracker name then
+   bucket recency and **drops the rest**, bumping
+   `infmon_exporter_points_dropped_total{reason="export_cap"}`. This
+   protects the collector from a backend misconfiguration.
+2. **Per-attribute length cap** — string attribute values are truncated
+   to 256 bytes (UTF-8 safe) with a trailing `…`. Bumps
```
banidoru, line 403: `export_duration` is a Gauge in seconds. A histogram would be far more useful for understanding latency distribution and tail behavior. A single gauge gives you only the latest value — you lose min/max/p99 at the collector. At minimum, consider making this a Summary or Histogram, or add separate `export_duration_max` / `export_duration_bucket` metrics.

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,406 @@
...
+### 7.1 Tick
+
+The exporter runs on a single periodic tick of period `export_interval`.
+On each tick:
+
+1. Ask the backend for a fresh snapshot (Spec 004 surface). The snapshot
+   is read-only and lock-free w.r.t. the data plane.
+2. Walk every tracker. For each live bucket emit the data points defined
+   in §3.2. Append the per-tracker points (§3.4).
+3. Group data points into OTLP `ResourceMetrics` → `ScopeMetrics` →
```
banidoru, line 287: The send queue drops the **oldest** batch when full, which makes sense for cumulative counters (newest supersedes). However, this should note that for the `last_seen` Gauge, dropping older batches means a receiver might see `last_seen` jump forward without seeing intermediate values. For a gauge this is generally fine, but worth a one-line acknowledgment since the doc is otherwise very precise about semantics.

banidoru (commented):
Thorough spec with solid failure-mode coverage. A few things to tighten:

- **Naming inconsistency**: §7/§8 internal references use underscore names (`infmon_exporter_ticks_dropped_total`) while §8.3 OTLP table uses dot names (`infmon.exporter.ticks_dropped`). Clarify which is canonical and that the other is just shorthand.
- **Missing `reason` enum for tracker drops**: `drops{reason}` attribute values are never enumerated.
- **No eviction-then-reappearance semantics**: What `start_time_unix_nano` does a bucket get if the same flow key reappears after eviction?
- **Tick jitter**: No jitter on the export tick across DPU fleet — potential thundering herd on the collector.
- **TLS key format unspecified**: `key_file` doesn't state accepted formats (PEM-wrapped PKCS#8? PKCS#1?).

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,406 @@
...
+
+## 4. Attribute mapping (per-bucket points)
+
+Each per-bucket data point carries a fixed attribute set derived from
+the tracker's field list (Spec 002 §3) plus a `tracker` label:
+
+| Attribute         | Type        | Source                                  |
+|-------------------|-------------|-----------------------------------------|
+| `tracker`         | string      | tracker `name` (Spec 002 §2.1)          |
+| `flow.src_ip`     | string      | `enc(src_ip)` rendered as canonical text (IPv4 if v4-mapped, else v6) |
```
banidoru, line 141: `infmon.tracker.drops` carries a `reason` attribute, but no enum of valid `reason` values is defined anywhere in the spec. This will lead to inconsistent implementation — define the set (e.g. `table_full`, `parse_error`, `rate_limit`) here or reference the Spec 002 section that owns them.

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,406 @@
...
+
+- Cumulative is the OTLP default and what most collector pipelines (Prometheus
+  remote-write, Mimir, Cortex) prefer.
+- The backend owns the bucket lifetime; the exporter is stateless. Cumulative
+  + a stable `start_time_unix_nano` lets a collector compute deltas without
+  the exporter having to track previous-export state.
+- On bucket eviction (Spec 002 §6) the counter is gone. Receivers that
+  compute deltas will see a counter reset; OTLP Sum semantics handle that
+  correctly when `start_time_unix_nano` advances.
+
```
banidoru, line 105: The eviction/reset story covers a bucket disappearing, but what happens when the **same flow key reappears** after eviction? The new bucket gets a fresh `start_time_unix_nano`, but there's a gap where cumulative counters jumped to zero. OTel Sum handles resets via `start_time_unix_nano` advancing, but this relies on the receiver correctly detecting the reset. Worth a one-liner confirming that the new bucket is treated as a brand-new time series (new `start_time_unix_nano`), not a continuation.

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,406 @@
...
+Configuration validation is all-or-nothing, mirroring Spec 002 §5.3.
+
+## 7. Batching, retry, timeout
+
+### 7.1 Tick
+
+The exporter runs on a single periodic tick of period `export_interval`.
+On each tick:
+
```
banidoru, line 282: The tick fires on a fixed `export_interval` with no jitter. If a fleet of DPUs all start around the same time (e.g. after a rolling upgrade), they'll hammer the collector in sync. Consider adding a startup jitter of `rand(0, export_interval)` on the first tick, similar to what the OTel SDK does by default.

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,406 @@
...
+- Prefer narrower trackers (e.g. `[src_ip]` for talker tables, `[dscp]`
+  for class-of-service rollups) for default-on deployments.
+- The growth signal is `infmon_tracker_evictions_total`. A non-zero,
+  growing eviction rate means cardinality exceeds the configured budget;
+  either grow `max_keys` or narrow the field list.
+
+### 8.2 Exporter-side safeguards
+
+The exporter MUST enforce these, independently of the backend:
+
```
banidoru, line 384: Naming inconsistency: this table uses OTel dot-separated names (`infmon.exporter.ticks_dropped`), but §7.1 and §8.2 reference the same metrics with Prometheus-style underscores (`infmon_exporter_ticks_dropped_total`, `infmon_exporter_batches_dropped_total`). Pick one as canonical (the dot-form is correct for OTLP) and add a note that the underscore form is just prose shorthand, or vice versa. As-is, an implementer won't know which to emit.

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,406 @@
...
+"x-tenant" = "dpu-east-1"
+
+[exporter.otlp.resource]
+# operator-set resource attributes, merged on top of §5 defaults
+"infmon.dpu.id"       = "bf3-rack17-u4"
+"infmon.dpu.platform" = "bluefield-3"
+```
+
+### 6.2 Defaults
+
```
banidoru, line 234: `key_file` is validated for existence but the spec doesn't state accepted key formats. PEM-wrapped PKCS#8? PKCS#1 RSA? EC? This matters for `rustls`/`native-tls` compatibility — specify at least "PEM-encoded" and which key types are supported, or state "any format accepted by the TLS backend."

banidoru (approved):
All reviewers approved. LGTM.

r12f:

specs/006-exporter-otlp.md
banidoru: Still open — the new commits only rename tracker→flow-rule / bucket→flow but do not address this concern.

specs/006-exporter-otlp.md
banidoru: Still open — the new commits only rename tracker→flow-rule / bucket→flow but do not address this concern.

specs/006-exporter-otlp.md
banidoru: Still open — the new commits only rename tracker→flow-rule / bucket→flow but do not address this concern.

specs/006-exporter-otlp.md
banidoru: Still open — the new commits only rename tracker→flow-rule / bucket→flow but do not address this concern.

specs/006-exporter-otlp.md
banidoru: Still open — the new commits only rename tracker→flow-rule / bucket→flow but do not address this concern.

specs/006-exporter-otlp.md
banidoru: Still open — the new commits only rename tracker→flow-rule / bucket→flow but do not address this concern.

specs/006-exporter-otlp.md
banidoru: Still open — the new commits only rename tracker→flow-rule / bucket→flow but do not address this concern.

specs/006-exporter-otlp.md
banidoru: Still open — the new commits only rename tracker→flow-rule / bucket→flow but do not address this concern.

specs/006-exporter-otlp.md
banidoru: Still open — the new commits only rename tracker→flow-rule / bucket→flow but do not address this concern.

specs/006-exporter-otlp.md
banidoru: Still open — the new commits only rename tracker→flow-rule / bucket→flow but do not address this concern.

specs/006-exporter-otlp.md
banidoru: Still open — the new commits only rename tracker→flow-rule / bucket→flow but do not address this concern.

specs/006-exporter-otlp.md
banidoru: Still open — the new commits only rename tracker→flow-rule / bucket→flow but do not address this concern.

specs/006-exporter-otlp.md
banidoru: Still open — the new commits only rename tracker→flow-rule / bucket→flow but do not address this concern.

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,412 @@
...
+Depends on: 000-overview (system overview),
+            002-flow-tracking-model (defines flow-rules, keys, flows,
+            and the per-flow-rule observability counters this exporter ships)
+Related: 004-backend-architecture (owns the snapshot mechanism the
+         exporter consumes), 005-frontend (hosts the exporter process),
+         007-cli (`infmon-cli exporter ...` surface)
+
+## 1. Purpose
+
+Define the **v1 OpenTelemetry (OTLP) exporter** for InFMon: how flow-flow-rule
```
banidoru: Garbled text from bad find-and-replace: `how flow-flow-rule` should be `how flow-rule`. Looks like `flow-tracker` was replaced with `flow-flow-rule` instead of just `flow-rule`.

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,412 @@
...
+  correctly when `start_time_unix_nano` advances.
+
+### 3.4 Per-flow-rule observability points
+
+The counters defined in Spec 002 §8 are exported alongside the per-flow
+data points. They are scoped to the flow-rule, not the key:
+
+| OTLP metric name                  | Instrument | Unit         | Source                                    |
+|-----------------------------------|------------|--------------|-------------------------------------------|
+| `infmon.flow-rule.flows`          | Gauge      | `{flows}`  | `infmon_tracker_buckets`                  |
```
banidoru: The OTLP metric names were updated to `infmon.flow-rule.*` but the Source column still references the old internal names (`infmon_tracker_buckets`, `infmon_tracker_evictions_total`, etc.). These should be updated to match the new terminology — e.g. `infmon_flow_rule_flows`, `infmon_flow_rule_evictions_total`, etc.

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,412 @@
...
+  `max_keys × distinct_data_points_per_bucket`. With v1's 3 per-flow
+  metrics, a flow-rule with `max_keys = 1_048_576` can land 3 Mi series in
+  the receiving TSDB **per scrape interval**.
+- The four v1 fields are *all* high-cardinality except `dscp`. A flow-rule
+  keyed on `[src_ip, dst_ip, ip_proto, dscp]` is appropriate for an
+  always-on production deployment **only** if the receiving TSDB is sized
+  for it.
+- Prefer narrower flow-rules (e.g. `[src_ip]` for talker tables, `[dscp]`
+  for class-of-service rollups) for default-on deployments.
+- The growth signal is `infmon_tracker_evictions_total`. A non-zero,
```
banidoru: Still references `infmon_tracker_evictions_total` — should use the renamed form consistent with §3.4.

banidoru (changes requested):
- All 13 prior review comments remain unaddressed — replied to each noting they are still open.
- The tracker→flow-rule rename introduced a garbled string "flow-flow-rule" on line 21 (double substitution).
- The Source column in §3.4 table (lines 94-98) and §8.1 (line 308) still references old `infmon_tracker_*` names — should be `infmon_flow_rule_*` or equivalent.
- No substantive spec changes beyond the rename — all prior design concerns (nanosecond precision, hardcoded arch, timeout validation, cardinality cap configurability, etc.) still need to be addressed.

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. Drop reason enum added in §3.4: table_full, parse_error, rate_limit, key_rejected.

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. Flow re-appearance after eviction documented in §3.3 — new start_time_unix_nano, fresh time series.

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. Startup jitter added in §7.1: uniform random delay in [0, export_interval).

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. Naming convention note added in §8.3 clarifying dot-form as canonical OTLP, underscore as prose shorthand.

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. key_file now specifies: PEM-encoded, PKCS#8 unencrypted (RSA or EC P-256/P-384), PKCS#1 RSA also accepted (§6).

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. Fixed: §8.1 now uses infmon_flow_rule_evictions_total.

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. Source column updated: infmon_tracker_* → infmon_flow_rule_* throughout §3.4.

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. Precision note for last_seen added (§3.2) — documents float64 precision risk and justifies keeping ns unit.

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. host.arch now derived from std::env::consts::ARCH at runtime (§5).

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. Validation rule 7 now explicitly documents why timeout == interval is rejected and the headroom rationale (§6.3).

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. Trade-off note added in §7.3 documenting the 40s retry / 3-tick data loss interaction.

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. Fixed: flow-flow-rule → flow-rule in §1.

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. max_export_points_per_tick is now operator-configurable via [exporter.otlp] TOML (§6, §8.2).

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. Cap enforcement now uses per-flow-rule proportional allocation instead of global O(n log n) sort (§8.2).

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. Acknowledged as Gauge-in-v1 with rationale note; Histogram listed as §9 extension hook.

specs/006-exporter-otlp.md
banidoru: Addressed in latest commits. Gauge note added in §7.2 explicitly acknowledging last_seen jump-forward semantics on batch drop.

banidoru (approved):
- All 16 prior review threads addressed and resolved — precision notes, runtime config, jitter, naming, drop-reason enum, key format, garbled text, stale references all fixed
- Terminology rename (tracker→flow-rule, bucket→flow) consistently applied; no stale `infmon_tracker_*` or `flow-flow-rule` references remain
- Cardinality cap now operator-configurable with proportional allocation strategy — cleaner than the global sort approach
- New clarifying notes (gauge semantics on batch drop, retry trade-off, export_duration rationale) add precision without bloat

specs/006-exporter-otlp.md
r12f: Kept the unit `ns` for consistency with OTel timestamp conventions, but added an explicit precision note to §3.2 calling out the float64 mantissa issue (~16 decimal digits) and pointing operators at the underlying `time_unix_nano` field for sub-microsecond use cases. Operators using `last_seen` for staleness/age (the only intended use) aren't affected.

specs/006-exporter-otlp.md
r12f: Switched §5 to derive `host.arch` from `std::env::consts::ARCH` at runtime (resolves to `aarch64` on the BF-3 .deb, `x86_64` on a dev box). No more hardcoded lie.

specs/006-exporter-otlp.md
r12f: Kept rule 7 as `>=` (which already excludes the equal boundary you flagged) and added the rationale inline: zero headroom for snapshot acquisition + serialization is not OK. Didn't tighten further to `> interval/2` — operators that genuinely want long timeouts should widen `export_interval` rather than have us pick a one-size factor.

specs/006-exporter-otlp.md
r12f: Added a Trade-off note in §7.3 spelling out the worst-case (~40s × queue_size = up to 3 ticks of silent aging on a flapping collector), the design choice (freshness > delivery, cumulative + last_seen self-heal), and the right knob to turn (lower `export_interval`, not enlarge `queue_size`). Kept the `4 * export_interval` ceiling — lowering it to `2x` would force collectors to choose between full retry budget and full export interval, which is worse for transient collector pain.

specs/006-exporter-otlp.md
r12f: Promoted `max_export_points_per_tick` to a TOML-configurable field (added to §6.1, §6.2 defaults, §6.3 validation rule 10, and §8.2). Default unchanged at 2_000_000. Operators can now tune it without a rebuild.

specs/006-exporter-otlp.md
r12f: Switched §8.2 to a per-flow-rule proportional allocation (`floor(cap × flows_in_rule / total_flows)`) — drops the global O(n log n) sort and gives each flow-rule representation roughly proportional to its live size. Cheaper at million-flow scale and arguably fairer.

specs/006-exporter-otlp.md
r12f: Kept Gauge in v1 to keep self-observability cheap and the cardinality story trivial, but added an explicit note in §8.3 explaining the limitation and listed Histogram migration as a §9 future hook. Will revisit when an operator actually needs p50/p99 — not today.

specs/006-exporter-otlp.md
r12f: Added an explicit gauge-jump note in §7.2: cumulative Sums are unambiguously safe under oldest-drop, but `last_seen` may skip intermediate samples — acceptable for a 'most recent packet' gauge but now stated.

specs/006-exporter-otlp.md
r12f: Defined a closed enum in §3.4: `table_full`, `parse_error`, `rate_limit`, `key_rejected`. Marked Spec 002 §8 as the authoritative list (this spec is a mirror) so the two specs can't drift.

specs/006-exporter-otlp.md
r12f: Added an explicit one-liner in §3.3: a reappearing flow key gets a fresh `first_seen_ns`, cumulative counters restart at zero, `start_time_unix_nano` advances, and receivers MUST treat it as a new time series rather than splicing.

specs/006-exporter-otlp.md
r12f: Added a startup jitter to §7.1: first tick is delayed by `rand(0, export_interval)` (computed once at startup), matching upstream OTel SDK default. Steady-state cadence remains exact `export_interval`.

specs/006-exporter-otlp.md
r12f: Added a Naming convention note in §8.3 declaring the dot form (`infmon.exporter.ticks_dropped`) canonical/wire-format, and explaining that the underscore forms in §7.1 / §8.2 / Spec 002 §8 are Prometheus-style prose shorthand that pipelines may end up exposing after the OTLP→Prometheus naming transform. Implementers emit the dot form.

specs/006-exporter-otlp.md
r12f: Documented in §6.1 inline comment: PEM-encoded; PKCS#8 unencrypted (RSA or EC P-256/P-384) is the canonical form; PKCS#1 RSA is also accepted on rustls; encrypted keys rejected at startup. Narrow enough to avoid backend-detection ambiguity.

specs/006-exporter-otlp.md
r12f: Fixed in §1: `flow-flow-rule` → `flow-rule`. Good catch.

specs/006-exporter-otlp.md
r12f: Updated the Source column in §3.4 to use the renamed counters: `infmon_flow_rule_flows`, `infmon_flow_rule_evictions_total`, `infmon_flow_rule_drops_total{reason}`, `infmon_flow_rule_packets_total`, `infmon_flow_rule_bytes_total`. Also updated the §8.1 reference to `infmon_flow_rule_evictions_total` for consistency.

specs/006-exporter-otlp.md
r12f: Fixed in §8.1 — updated to `infmon_flow_rule_evictions_total` for consistency with the §3.4 source rename.

r12f:

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,516 @@
...
+
+### 8.2 Exporter-side safeguards
+
+The exporter MUST enforce these, independently of the backend:
+
+1. **Hard per-export point cap** — `max_export_points_per_tick` (operator
+   configurable in the `[exporter.otlp]` TOML, see §6; default
+   `2_000_000`). If a tick would emit more, the exporter ships
+   approximately the cap-many points using a **per-flow-rule proportional
+   allocation**: each flow-rule gets a budget of
```
banidoru, line 389: Unit mismatch: `max_export_points_per_tick` is denominated in **points**, but the formula `floor(cap × flow_rule_flows / total_flows)` produces a per-flow-rule budget in **flows**, and then "each flow-rule emits its first `budget` flows" — each of which generates 3 data points (packets, bytes, last_seen). So the actual emitted point count can reach `3 × cap`, exceeding the configured safety limit.

Either:
- Divide the cap by the points-per-flow multiplier (currently 3) before allocation: `effective_cap = cap / points_per_flow`, then allocate flows from that, or
- State the budget in points (not flows) and have each flow-rule convert back to a flow count internally.

As written, a `max_export_points_per_tick = 2_000_000` config could emit up to 6M points.

specs/006-exporter-otlp.md
```diff
@@ -0,0 +1,516 @@
...
+  `/etc/machine-id` is missing, to keep the resource set complete? v1
+  default: omit. Revisit if collector-side joins prove painful.
+- **Q2.** Do we expose `temporality = "delta"` as a v1 hidden flag for
+  collectors that prefer it (Datadog, some SaaS APMs)? v1 default: no,
+  ship cumulative only. Re-evaluate once we have a concrete user.
+- **Q3.** Should `infmon.dpu.id` be a *required* operator-set value
+  (refuse to start if unset) rather than falling back to `host.name`?
+  Defer to operator feedback after first deploy.
+- **Q4.** OTLP exemplars on the `Sum` metrics, linking back to a future
+  trace span? Out of scope for v1; revisit when a tracing story exists.
```
banidoru: Nit: file is missing a trailing newline (`\ No newline at end of file`). Most linters and POSIX tools expect one.

banidoru (commented):
All prior review threads addressed and resolved. New diff review:

- **Unit mismatch in §8.2 cap allocation**: `max_export_points_per_tick` is in *points*, but the proportional formula allocates *flows* — each flow emits 3 data points, so the actual point count can be up to 3× the cap.
- **Missing trailing newline** at end of `006-exporter-otlp.md`.
- §1.1 mental-model paragraph and `flow.mirror_src_ip` attribute are clean additions, no issues.
- Version history and metadata cleanup looks correct.

banidoru (commented):
Iteration 3 review:

- All 15 previously-resolved threads verified — fixes look correct, no regressions
- §1.1 mental-model paragraph is a good addition; clear and concise
- `flow.mirror_src_ip` attribute properly scoped as opt-in with Spec 003 cross-ref
- **Still open:** §8.2 unit mismatch (points vs flows) — cap can emit up to 3× configured limit
- **Still open:** missing trailing newline

specs/006-exporter-otlp.md
banidoru: Still open in this iteration. The formula allocates `budget` in flows but the cap is denominated in points. With 3 data points per flow, the actual emitted point count can reach `3 × cap`. Either divide the cap by points-per-flow before allocation, or rename the config to `max_export_flows_per_tick` and document that it controls flow count (not point count).

specs/006-exporter-otlp.md
banidoru: Still open — file still lacks a trailing newline.

specs/006-exporter-otlp.md
banidoru: Still open — this commit only reformats the version history table and dependency links. The points-vs-flows unit mismatch in §8.2 remains unaddressed.

specs/006-exporter-otlp.md
banidoru: Still open — file still lacks a trailing newline.

banidoru (commented):
- Version history collapsed from 4 rows into 1 — loses the incremental change trail; consider keeping per-version entries for auditability
- Dependency links now use relative markdown (`[002-flow-tracking-model](002-flow-tracking-model.md)`) which is good, but the detailed dependency descriptions (e.g. "defines flow-rules, keys, flows…") were dropped — those were useful context for readers unfamiliar with the spec tree
- Two prior open threads remain unaddressed: (1) points-vs-flows unit mismatch in §8.2 cap formula, (2) missing trailing newline
- No functional spec changes in this commit — cosmetic only

banidoru:
Still open — this commit only reformats the version history table and dependency links. The points-vs-flows unit mismatch in §8.2 remains unaddressed.

banidoru:
Still open — file still lacks a trailing newline.

banidoru (changes requested):
- The unit mismatch in `max_export_points_per_tick` remains unaddressed.
- The file is still missing a trailing newline.

specs/006-exporter-otlp.md
banidoru: Still open — this commit only reformats the version history table and linkifies dependency references. The points-vs-flows unit mismatch in §8.2 remains unaddressed.

specs/006-exporter-otlp.md
banidoru: Still open — file still lacks a trailing newline.

banidoru (commented):
**Iteration 4 re-review — version-history cleanup commit only**

- ✅ Version history collapsed to a single row per PR — cleaner.
- ✅ Depends/Related now linkified to sibling spec files — good navigability improvement.
- ⚠️ **Points-vs-flows unit mismatch** (§8.2) still unaddressed — `max_export_points_per_tick` cap is denominated in points but allocation formula produces flow budgets, so actual emitted points can reach `3 × cap`.
- ⚠️ **Missing trailing newline** — still no newline at end of file.
- No new issues introduced by this commit.

specs/006-exporter-otlp.md
r12f: Fixed in latest commit: added `effective_cap = floor(cap / points_per_flow)` so the budget is allocated in flows but the total emitted points never exceed the configured cap. The formula now correctly accounts for the 3 data points per flow.

specs/006-exporter-otlp.md
r12f: Fixed — trailing newline added.

banidoru (approved):
- Points-vs-flows unit mismatch in §8.2 fixed: `effective_cap = floor(cap / points_per_flow)` correctly ensures emitted points never exceed the configured cap
- Trailing newline added
- All 18 prior review threads addressed and resolved across iterations
- No new issues found in this diff
- LGTM — spec is solid

banidoru (approved):
- All prior review threads resolved — fixes verified in latest commits
- points-vs-flows unit mismatch in §8.2 fixed via `effective_cap = floor(cap / points_per_flow)` — correct
- Trailing newline added
- No new issues in this diff
- LGTM

banidoru (approved):
All reviewers approved. LGTM.

PR #14, opened 2026-04-18
specs: fix systemd unit — disabled by default, add --run install flag

r12f:
## Summary

Resolves the contradiction between spec 000 (disabled by default) and spec 008 §6 (enabled-by-default).

**Decision:** install as disabled by default. Add a `--run` parameter to install as enabled.

### Changes to spec 008

- **§6 notes:** Replace enabled-by-default with disabled-by-default. Document `--run` flag and `dh_installsystemd --no-enable --no-start`.
- **§7.1 install flow:** Service is NOT enabled or started unless `--run` is passed.
- **§7.4 maintainer-script summary:** Updated postinst description to match.

Spec 000 already says disabled by default — no change needed there.

Resolves: DPU-29

### DIFF
--- a/specs/008-packaging-install.md
+++ b/specs/008-packaging-install.md
@@ -5,7 +5,7 @@
 | Version | Date       | Author       | Changes |
 | ------- | ---------- | ------------ | ------- |
 | 0.1     | 2026-04-18 | Riff (r12f)  | Initial draft of the `.deb` packaging contract for Ubuntu aarch64 on BlueField-3. Version-locked deps, source/format declared, postrm cross-package fix, equivs preferred over `--force-depends`, GPG verification note, version-skew check, CLI stability marking, explicit cmake + cargo rules; PartOf and compat file dropped per PR #9 review. |
-| 0.2     | 2026-04-18 | BF-3 (bf3)   | Fix systemd unit default: disabled by default instead of enabled. Add `--run` install flag to opt in to enable+start. Update §6 notes, §7.1 install flow, §7.4 maintainer-script summary. Resolves contradiction with spec 000. |
+| 0.2     | 2026-04-18 | Riff (r12f)   | Fix systemd unit default: disabled by default instead of enabled. Add `--run` install flag to opt in to enable+start. Update §6 notes, §7.1 install flow, §7.4 maintainer-script summary. Resolves contradiction with spec 000. |
 
 - **Depends on:** [`000-overview`](000-overview.md), [`004-backend-architecture`](004-backend-architecture.md)
 - **Affects:** [`005-frontend-architecture`](005-frontend-architecture.md), [`007-cli`](007-cli.md), [`001-ci-and-precommit`](001-ci-and-precommit.md) (build job produces the `.deb`)
--- a/specs/008-packaging-install.md
+++ b/specs/008-packaging-install.md
@@ -5,6 +5,7 @@
 | Version | Date       | Author       | Changes |
 | ------- | ---------- | ------------ | ------- |
 | 0.1     | 2026-04-18 | Riff (r12f)  | Initial draft of the `.deb` packaging contract for Ubuntu aarch64 on BlueField-3. Version-locked deps, source/format declared, postrm cross-package fix, equivs preferred over `--force-depends`, GPG verification note, version-skew check, CLI stability marking, explicit cmake + cargo rules; PartOf and compat file dropped per PR #9 review. |
+| 0.2     | 2026-04-18 | BF-3 (bf3)   | Fix systemd unit default: disabled by default instead of enabled. Add `--run` install flag to opt in to enable+start. Update §6 notes, §7.1 install flow, §7.4 maintainer-script summary. Resolves contradiction with spec 000. |
 
 - **Depends on:** [`000-overview`](000-overview.md), [`004-backend-architecture`](004-backend-architecture.md)
 - **Affects:** [`005-frontend-architecture`](005-frontend-architecture.md), [`007-cli`](007-cli.md), [`001-ci-and-precommit`](001-ci-and-precommit.md) (build job produces the `.deb`)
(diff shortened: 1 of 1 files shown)


specs/008-packaging-install.md
```diff
@@ -281,9 +282,15 @@ Notes:
...
-- The unit is installed enabled-by-default via `dh_installsystemd`'s
-  default behaviour; operators who do not want it auto-started can
-  `systemctl disable infmon-frontend` after install.
+- The unit is installed **disabled by default** — `dh_installsystemd
+  --no-enable --no-start` prevents the service from starting or being
+  enabled on first install. Operators who want the service to start
+  immediately can pass `--run` to the installer script
+  (`dpkg --configure -D infmon-frontend -- --run`, or the equivalent
+  `apt` invocation documented in `debian/README.Debian`), which causes
+  the postinst to `systemctl enable --now infmon-frontend`. This
```
banidoru, line 291: `dpkg --configure -D` sets the **debug level**, not a pass-through argument. This invocation won't deliver `--run` to the postinst script.

Common alternatives for passing options to maintainer scripts:
1. **debconf preseed** — `echo 'infmon-frontend infmon/enable-now boolean true' | debconf-set-selections` before install.
2. **Environment variable** — `INFMON_RUN=1 apt install infmon` with the postinst checking `$INFMON_RUN`.
3. **Post-install CLI** — `infmon enable` or `systemctl enable --now infmon-frontend` documented as a manual step.

Please pick a mechanism and update the spec (and the §7.1 / §7.4 references that depend on it).

banidoru (commented):
Spec review — iteration 1:

- **Good:** Resolves the spec 000 vs spec 008 contradiction cleanly; disabled-by-default is the right call for production DPUs.
- **`dpkg --configure -D` is a debug flag**, not a mechanism for passing arguments to postinst — the documented invocation won't work as described. Needs a different approach (e.g. debconf, env var, or a separate `infmon enable` CLI command).
- **`--run` pass-through mechanism is unspecified** — how does `dh_installsystemd --no-enable --no-start` interact with a custom `--run` flag in postinst? The spec should clarify exactly how the flag reaches the maintainer script.
- Minor: version row has trailing whitespace inconsistency (extra space before `|`).
- Overall direction is sound; the flag delivery mechanism needs rework.

banidoru (commented):
- The `dpkg --configure -D` mechanism for passing `--run` is incorrect (already noted in prior review) — this needs a workable delivery mechanism before the spec can be finalized.
- Upgrade path from v0.1 (enabled-by-default) to v0.2 (disabled-by-default) is unspecified — will `--no-enable` on upgrade disable a previously-enabled service? This needs explicit handling in the postinst.
- §7.1 pseudocode references the broken `--run` flag mechanism; must be updated together with §6.
- Minor: version table row 0.2 has an extra space in the Author column vs row 0.1 (cosmetic).

specs/008-packaging-install.md
```diff
@@ -281,9 +282,15 @@ Notes:
...
+  conservative default avoids surprising operators on a production DPU
+  where VPP may not be ready yet.
 
 There is **no** unit for `infmon-backend` — the backend is a VPP plugin,
 not a service. Its lifecycle is VPP's.
```
banidoru, line 296: The spec doesn't address the **upgrade path**. If an operator installed v0.1 (where the service was enabled-by-default and running), what happens when they upgrade to v0.2?

`dh_installsystemd --no-enable --no-start` only suppresses enable/start on **first install** — on upgrade, `dpkg` invokes postinst with `configure <old-version>`, and `deb-systemd-helper` typically preserves existing enable state. But since the postinst logic is being rewritten, the spec should explicitly state:
1. On upgrade from a version that enabled-by-default: leave existing enable state untouched (don't forcibly disable a running service).
2. On fresh install: disabled by default (as written).

Without this, operators upgrading may get inconsistent behaviour depending on whether the postinst migration logic is correct.

specs/008-packaging-install.md
```diff
@@ -301,8 +308,8 @@ apt install infmon
   └─ postinst (frontend):
        creates infmon user/group (idempotent)
        creates /var/log/infmon, /var/lib/infmon
-       deb-systemd-helper enable infmon-frontend.service
-       deb-systemd-invoke start infmon-frontend.service
+       (service is NOT enabled or started by default)
+       if --run flag was passed: systemctl enable --now infmon-frontend.service
   └─ postinst (backend):
        prints NOTICE asking operator to restart VPP
 ```
```
banidoru, line 315: This pseudocode says `if --run flag was passed` but the mechanism for passing `--run` is broken (see prior review on the `dpkg --configure -D` issue at §6). This section and §6 need to be updated together once a delivery mechanism is chosen — otherwise the spec has two sections referencing a non-functional interface.

specs/008-packaging-install.md
```diff
@@ -5,6 +5,7 @@
 | Version | Date       | Author       | Changes |
 | ------- | ---------- | ------------ | ------- |
 | 0.1     | 2026-04-18 | Riff (r12f)  | Initial draft of the `.deb` packaging contract for Ubuntu aarch64 on BlueField-3. Version-locked deps, source/format declared, postrm cross-package fix, equivs preferred over `--force-depends`, GPG verification note, version-skew check, CLI stability marking, explicit cmake + cargo rules; PartOf and compat file dropped per PR #9 review. |
+| 0.2     | 2026-04-18 | Riff (r12f)   | Fix systemd unit default: disabled by default instead of enabled. Add `--run` install flag to opt in to enable+start. Update §6 notes, §7.1 install flow, §7.4 maintainer-script summary. Resolves contradiction with spec 000. |
```
banidoru, line 8: Nit: extra trailing space in the Author column (`Riff (r12f)   |` — 3 spaces vs 2 in the row above). Cosmetic only.

r12f:

specs/008-packaging-install.md
r12f: Good catch — dpkg --configure -D does not pass --run through to the postinst. Of the three alternatives proposed, I recommend **option 2 (environment variable INFMON_RUN=1)** as the simplest and most Debian-idiomatic approach. Environment variables survive through dpkg invocation naturally, and this pattern is used by other packages (e.g. DEBIAN_FRONTEND). Leaving unresolved as this is an architectural decision that changes the install API surface and needs explicit sign-off.

specs/008-packaging-install.md
r12f: Acknowledged — the upgrade path semantics are missing. On upgrade, the existing enable/start state should be preserved; the postinst must not forcibly disable a service that was previously enabled. Recommend adding explicit upgrade semantics (e.g. checking deb-systemd-helper was-enabled or equivalent). Leaving unresolved as this affects the packaging contract across versions.

specs/008-packaging-install.md
r12f: This is blocked on the --run delivery mechanism decision in the thread at line 291. Once that is resolved, the downstream logic here can be updated accordingly. Leaving unresolved.

specs/008-packaging-install.md
r12f: Fixed — removed the extra trailing space in the version history table (line 8).

PR #15, opened 2026-04-18
specs: fix eviction policy contradiction — adopt lru_drop for v1

r12f:
## Summary

Spec 004 §5.2 stated *"We deliberately do not evict; eviction policy is a v2 conversation"* which directly contradicts spec 002 §6 that defines `lru_drop` as the v1 eviction policy.

**Decision (per DPU-25):** spec 002 is correct — `lru_drop` is the v1 policy.

### Spec 004 (backend architecture)
- **§5.2**: Replaced the no-eviction statement with a reference to spec 002 §6's `lru_drop` policy
- **§11**: Updated `counter_table_full` error counter description to note that eviction is triggered

### Spec 000 (overview)
- **§5 step 5**: Fixed cross-reference from "spec 003 to be written" → "spec 002" 
- **Open questions item 3**: Replaced open question about eviction policy with the decided answer (`lru_drop`)

## Verification
- Searched all specs for eviction/lru references — no other contradictions found
- Specs 002, 005, 006, 007 already correctly reference `lru_drop` / eviction from spec 002

Resolves: DPU-25

### DIFF
--- a/specs/000-overview.md
+++ b/specs/000-overview.md
@@ -5,6 +5,7 @@
 | Version | Date       | Author      | Changes |
 | ------- | ---------- | ----------- | ------- |
 | 0.1     | 2026-04-18 | Riff (r12f) | Initial draft. Establishes mission, scope, component map, repo layout, build/release model, glossary (incl. flow / flow-rule), and pointers to the canonical spec template at [`TEMPLATE.md`](TEMPLATE.md). |
+| 0.2     | 2026-04-18 | BF-3 (bf3)  | Fix eviction policy references: §5 step 5 now points to spec 002 (not 003); §open-questions item 3 updated to reflect `lru_drop` decision. |
 
 ---
 
@@ -151,7 +152,7 @@ packets         |     v                                         |
 4. **Match & key.** L2–L4 parser produces a flow key (5-tuple + VRF/VNI when
    available) and per-packet metadata (length, TCP flags, timestamp).
 5. **Update.** Flow table lookup; on hit, update counters in place; on miss,
-   insert a new flow record (with eviction policy per spec 003 to be written).
+   insert a new flow record (with eviction policy per spec 002).
 6. **Snapshot.** Frontend periodically reads the backend's exposed snapshot
    (lock-free / RCU-style; details in spec 002) and builds an aggregate view.
 7. **Export.** Frontend pushes records to configured exporters and serves
@@ -333,8 +334,7 @@ explanations of each rule.
 2. **Snapshot transport** — shared memory ring vs. local Unix-domain socket
    vs. gRPC. Decided in spec 002. *Default:* shared-memory ring (lowest
    overhead on DPU).
-3. **Flow table eviction policy** — LRU, time-based, or hybrid. Decided in
-   spec 003. *Default:* time-based with active/idle timeouts (IPFIX-aligned).
+3. **Flow table eviction policy** — `lru_drop` (decided in spec 002 §6).
 4. **Distro target** — Ubuntu 22.04 only, or also DOCA's recommended base?
    Decided in spec 005. *Default:* Ubuntu 22.04 arm64.
 5. **Frontend API authn/authz** — v1 ships with UNIX-group-based access on a
--- a/specs/000-overview.md
+++ b/specs/000-overview.md
@@ -5,7 +5,7 @@
 | Version | Date       | Author      | Changes |
 | ------- | ---------- | ----------- | ------- |
 | 0.1     | 2026-04-18 | Riff (r12f) | Initial draft. Establishes mission, scope, component map, repo layout, build/release model, glossary (incl. flow / flow-rule), and pointers to the canonical spec template at [`TEMPLATE.md`](TEMPLATE.md). |
-| 0.2     | 2026-04-18 | BF-3 (bf3)  | Fix eviction policy references: §5 step 5 now points to spec 002 (not 003); §open-questions item 3 updated to reflect `lru_drop` decision. |
+| 0.2     | 2026-04-18 | Riff (r12f)  | Fix eviction policy references: §5 step 5 now points to spec 002 (not 003); §open-questions item 3 updated to reflect `lru_drop` decision. |
 
 ---
 
--- a/specs/004-backend-architecture.md
+++ b/specs/004-backend-architecture.md
@@ -5,7 +5,7 @@
 | Version | Date       | Author       | Changes |
 | ------- | ---------- | ------------ | ------- |
 | 0.1     | 2026-04-18 | Riff (r12f)  | Initial draft of `infmon-backend` (VPP plugin). Linear-probing flow table with offset-based descriptors, memory ordering, epoch-based RCU, scratch cap, alloc-failed recovery; internal identifiers use `flow_rule*` per Spec 002 mental model; per-worker scratch-triple and §6 emit format use `flow_rule_index` (u32 handle), keeping the 24 B/entry estimate. |
-| 0.2     | 2026-04-18 | BF-3 (bf3)   | Fix eviction policy contradiction: §5.2 now adopts `lru_drop` (spec 002 §6) instead of deferring eviction to v2. Updated §11 `counter_table_full` description accordingly. |
+| 0.2     | 2026-04-18 | Riff (r12f)   | Fix eviction policy contradiction: §5.2 now adopts `lru_drop` (spec 002 §6) instead of deferring eviction to v2. Updated §11 `counter_table_full` description accordingly. |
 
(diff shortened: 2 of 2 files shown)


banidoru (commented):
Spec-only fix, changes are clear and well-scoped. A few issues:

- `infmon_flow_rule_evictions_total` referenced in §5.2 but not defined in §11 error counter table — add it or cross-ref where it lives
- Eviction failure reuses `counter_insert_retry_exhausted` (CAS retry counter) — may conflate two distinct failure modes
- `counter_table_full` description says eviction is triggered, but the counter name implies the table stayed full — clarify semantics (fires before or after eviction attempt?)

specs/004-backend-architecture.md
```diff
@@ -158,8 +159,11 @@ struct infmon_slot {
   exhaustion we increment a per-table `insert_failed` counter and drop
   the contribution from this packet (the packet itself still goes to
   `drop`, like every other packet — InFMon never forwards).
-- Table full → contribution dropped, `table_full` counter incremented.
-  We deliberately do not evict; eviction policy is a v2 conversation.
+- Table full → the `lru_drop` eviction policy (spec 002 §6) applies:
+  evict the least-recently-updated key, drop its residual counters,
+  increment `infmon_flow_rule_evictions_total`, and insert the new key.
```
banidoru, line 164: `infmon_flow_rule_evictions_total` is referenced here but doesn't appear in the §11 error counter table. Should it be added there, or is it exposed via a different stats mechanism (e.g. Prometheus gauge)? Either way, the spec should say where this counter lives.

specs/004-backend-architecture.md
```diff
@@ -158,8 +159,11 @@ struct infmon_slot {
...
+  If eviction itself fails, the contribution is dropped and
+  `counter_insert_retry_exhausted` is incremented.
```
banidoru, line 166: Reusing `counter_insert_retry_exhausted` for eviction failure conflates two distinct failure modes: (1) CAS retries exhausted during normal insert, and (2) eviction-then-insert failure. Operators debugging table-full events would benefit from a dedicated counter (e.g. `counter_eviction_failed`) to distinguish the two paths.

specs/004-backend-architecture.md
```diff
@@ -373,7 +377,7 @@ The plugin exposes the following error counters via the standard VPP
 | `inner_parse_failed`          | Inner L2/L3/L4 parse error after decap.              |
 | `flow_rule_no_match`           | Packet matched zero flow_rules (informational).       |
 | `counter_insert_retry_exhausted` | CAS retries exceeded `INFMON_INSERT_RETRY`.        |
-| `counter_table_full`          | Table reached `max_keys_per_flow_rule`.               |
+| `counter_table_full`          | Table reached `max_keys_per_flow_rule`; `lru_drop` eviction triggered (spec 002 §6). |
```
banidoru, line 380: The description now says eviction is triggered, but the counter name `counter_table_full` suggests the table *remained* full. Consider clarifying the semantics: does this counter fire when the table is full *before* attempting eviction (i.e., every time eviction is attempted), or only when eviction fails and the table is still full? If the former, the name is slightly misleading post-eviction.

specs/000-overview.md
```diff
@@ -151,7 +152,7 @@ packets         |     v                                         |
 4. **Match & key.** L2–L4 parser produces a flow key (5-tuple + VRF/VNI when
    available) and per-packet metadata (length, TCP flags, timestamp).
 5. **Update.** Flow table lookup; on hit, update counters in place; on miss,
-   insert a new flow record (with eviction policy per spec 003 to be written).
+   insert a new flow record (with eviction policy per spec 002).
 6. **Snapshot.** Frontend periodically reads the backend's exposed snapshot
    (lock-free / RCU-style; details in spec 002) and builds an aggregate view.
 7. **Export.** Frontend pushes records to configured exporters and serves
```
banidoru, line 158: The link was updated to spec 002, but the text still implies the spec is not yet written. The phrase "to be written" should be removed, as the policy is now defined.

banidoru (commented):
- The new `lru_drop` eviction policy is a significant improvement, but the PR introduces minor inconsistencies in counter naming and documentation.
- The existing review comments correctly identify the main issues in `spec-004` regarding the new `infmon_flow_rule_evictions_total` counter, the reuse of `counter_insert_retry_exhausted`, and the ambiguous naming of `counter_table_full`.
- I have added one minor inline comment regarding a dangling reference in the overview spec.

specs/000-overview.md
```diff
@@ -333,8 +334,7 @@ explanations of each rule.
 2. **Snapshot transport** — shared memory ring vs. local Unix-domain socket
    vs. gRPC. Decided in spec 002. *Default:* shared-memory ring (lowest
    overhead on DPU).
-3. **Flow table eviction policy** — LRU, time-based, or hybrid. Decided in
-   spec 003. *Default:* time-based with active/idle timeouts (IPFIX-aligned).
+3. **Flow table eviction policy** — `lru_drop` (decided in spec 002 §6).
```
banidoru, line 337: This item is now decided (`lru_drop` per spec 002 §6), yet it still sits in the **Open Questions** list. Consider either moving it to a "Resolved" section or striking it with a note, so readers don't treat it as still open.

specs/004-backend-architecture.md
```diff
@@ -158,8 +159,11 @@ struct infmon_slot {
   exhaustion we increment a per-table `insert_failed` counter and drop
   the contribution from this packet (the packet itself still goes to
   `drop`, like every other packet — InFMon never forwards).
-- Table full → contribution dropped, `table_full` counter incremented.
-  We deliberately do not evict; eviction policy is a v2 conversation.
+- Table full → the `lru_drop` eviction policy (spec 002 §6) applies:
+  evict the least-recently-updated key, drop its residual counters,
```
banidoru, line 163: "evict the least-recently-updated key" — how is LRU ordering tracked? A linear-probing table doesn't inherently maintain insertion/access order. If you're adding a `last_updated` timestamp per slot, that increases per-entry size and should be accounted for in the 24 B/entry estimate from §0.1. If relying on a different mechanism, it should be specified here.

banidoru, line 163: "drop its residual counters" — does the evicted flow's accumulated data get silently discarded, or is there a best-effort export/flush before eviction? For a monitoring tool, silently losing flow records on eviction could be a significant data-integrity concern. The spec should state the intended behavior explicitly.

banidoru (commented):
- Open question #3 is now decided but still lives in the "Open Questions" section — should be marked resolved or moved.
- §5.2 says "evict the least-recently-updated key" but doesn't specify the LRU tracking mechanism (timestamp per slot? separate list?) — important given the 24 B/entry budget.
- "Drop its residual counters" on eviction — unclear if evicted flow data is silently lost or flushed/exported first; data loss implications should be documented.
- Agree with existing feedback on the missing `infmon_flow_rule_evictions_total` row in §11 and the conflated failure counter.

banidoru (approved):
All reviewers approved. LGTM.

r12f:

specs/004-backend-architecture.md
r12f: Done — added `infmon_flow_rule_evictions_total` to the §11 counter table with description "Number of successful LRU evictions performed."

specs/004-backend-architecture.md
r12f: Done — added dedicated `counter_eviction_failed` counter in both §5.2 and §11 table. `counter_insert_retry_exhausted` now only covers non-eviction insert failures.

specs/004-backend-architecture.md
r12f: Done — clarified `counter_table_full` description: fires when table is full and eviction is initiated, not on eviction success/failure (those are tracked separately by `infmon_flow_rule_evictions_total` and `counter_eviction_failed`).

specs/000-overview.md
r12f: Already addressed in this PR — the "to be written" text was replaced with a reference to spec 002 in the eviction policy fix.

specs/000-overview.md
r12f: Done — struck through the resolved eviction policy question (item 3) in the Open Questions list and marked it as **Resolved**.

specs/004-backend-architecture.md
r12f: Done — added a note in §5.2 that LRU ordering is tracked via a `last_updated` timestamp stored in each slot (adds 8 B per entry to the §5.1 layout). Simplest approach that avoids a separate LRU linked list.

r12f: Done — stated explicitly that residual counters are silently dropped on eviction (v1 simplification, consistent with "drop" in `lru_drop` name). Flushing evicted counters before discard is deferred to v2.

PR #23, opened 2026-04-18
specs: add VPP disable snippet to spec 008 file layout

r12f:
Spec 000 describes `/etc/vpp/startup.d/10-infmon.conf` (disables plugin autoload by default), but this file was missing from spec 008's authoritative file layout (§5.1) and the `debian/` source tree.

**Changes:**
- Add `/etc/vpp/startup.d/10-infmon.conf` row to §5.1 `infmon-backend` file table
- Add `debian/10-infmon.conf` and `debian/infmon-backend.conffiles` to the §8 source-tree layout
- Add `install` command in `override_dh_auto_install` to place the snippet
- Update version history (v0.2)


specs/008-packaging-install.md
```diff
@@ -404,8 +406,10 @@ debian/
 ├── rules                  # see below — explicit override targets, no cmake+cargo addon
 ├── source/format          # 3.0 (quilt)
 ├── infmon-backend.install
+├── infmon-backend.conffiles
```
banidoru, line 409: Nit: with debhelper compat ≥ 12, files installed under `/etc/` are automatically marked as conffiles. If the package already uses compat 12+, an explicit `.conffiles` could create duplicate `conffiles` entries in the binary package. Consider adding a note here clarifying why it's declared explicitly (e.g., targeting older compat levels), or dropping it if compat ≥ 12 is assumed.

banidoru (commented):
- Spec alignment with spec 000 looks correct — the missing VPP snippet file is now accounted for in both the file layout table and the source tree.
- `install -Dm0644` invocation is correct for a config file.
- Minor: version history row has a trailing space before the `|` delimiter (cosmetic only).
- Consider whether an explicit `infmon-backend.conffiles` is needed — debhelper compat ≥ 12 auto-detects files under `/etc/` as conffiles. An explicit file risks duplicate entries if the compat level is already ≥ 12. Worth a note in the spec clarifying why it's explicit.
- No correctness or security concerns.

specs/008-packaging-install.md
banidoru, line 409: This PR adds another file under `/etc` (`/etc/vpp/startup.d/10-infmon.conf`) and also adds an explicit `infmon-backend.conffiles` file. As noted in a previous review comment on this line, with debhelper compat level 12 or higher, files installed under `/etc` are automatically marked as conffiles. Explicitly creating this file is likely redundant and could cause packaging issues. It would be better to rely on the default debhelper behavior unless there is a specific reason to support older compatibility levels.

banidoru (commented):
- The explicit creation of `debian/infmon-backend.conffiles` seems redundant, as debhelper (compat 12+) automatically treats files in `/etc` as conffiles. This could lead to duplicate entries, echoing a concern from a previous review. Consider removing the explicit conffile unless targeting older compat levels.

specs/008-packaging-install.md
```diff
@@ -5,6 +5,7 @@
 | Version | Date       | Author       | Changes |
 | ------- | ---------- | ------------ | ------- |
 | 0.1     | 2026-04-18 | Riff (r12f)  | Initial draft of the `.deb` packaging contract for Ubuntu aarch64 on BlueField-3. Version-locked deps, source/format declared, postrm cross-package fix, equivs preferred over `--force-depends`, GPG verification note, version-skew check, CLI stability marking, explicit cmake + cargo rules; PartOf and compat file dropped per PR #9 review. |
+| 0.2     | 2026-04-18 | Riff (r12f)   | Add `/etc/vpp/startup.d/10-infmon.conf` (VPP plugin disable snippet) to §5.1 file layout and `debian/` source tree per spec 000 alignment. |
```
banidoru, line 8: Minor formatting nit: there's an extra trailing space before the `|` in this row compared to the v0.1 row above. Not a functional issue, but inconsistent column alignment.

specs/008-packaging-install.md
```diff
@@ -154,6 +155,7 @@ files outside `/usr`, `/etc`, `/var`, `/lib/systemd/system`, or
 | `/usr/lib/vpp_plugins/infmon.so`              | root  | 0644 | VPP picks this up automatically on next start.         |
 | `/usr/share/vpp/api/plugins/infmon.api.json`  | root  | 0644 | Generated by `vppapigen`; consumed by Rust bindings.   |
 | `/etc/infmon/backend.conf`                    | root  | 0644 | Conffile — `dpkg` preserves operator edits on upgrade. |
+| `/etc/vpp/startup.d/10-infmon.conf`           | root  | 0644 | Disables plugin autoload by default (see spec 000). Operators flip `disable` → `enable` on instances where InFMon should run. |
```
banidoru, line 158: The description says operators flip `disable` → `enable`, but the spec doesn't show the actual file contents anywhere. Consider adding the expected snippet (e.g., `plugin infmon { disable }`) either inline here or in a dedicated subsection, so implementers don't have to cross-reference spec 000 for a one-liner.

specs/008-packaging-install.md
```diff
@@ -404,8 +406,10 @@ debian/
 ├── rules                  # see below — explicit override targets, no cmake+cargo addon
 ├── source/format          # 3.0 (quilt)
 ├── infmon-backend.install
+├── infmon-backend.conffiles
 ├── infmon-backend.postinst
 ├── infmon-backend.postrm
+├── 10-infmon.conf               # VPP startup.d snippet (installed to /etc/vpp/startup.d/)
 ├── infmon-frontend.install
```
banidoru, line 413: Placing `10-infmon.conf` directly under `debian/` (not in a subdirectory) is fine for a single snippet, but if more VPP startup.d files are added later this could get cluttered. A `debian/vpp-startup.d/` subdirectory might be worth considering for future-proofing — though not strictly necessary for this PR.

banidoru (commented):
- Clean, well-scoped change aligning spec 008 with spec 000.
- The `infmon-backend.conffiles` addition is likely redundant with debhelper compat ≥ 12 (already flagged by prior reviewers).
- Version history row has a trailing extra space before the `|` delimiter — minor formatting nit.
- The `10-infmon.conf` snippet is installed but its contents aren't shown in the spec; consider adding the expected file contents (e.g., `plugin infmon { disable }`) so implementers have a single source of truth.
- No concerns with the `install -Dm0644` approach — correct mode and idiomatic for `debian/rules`.

banidoru (approved):
All reviewers approved. LGTM.

r12f:

specs/008-packaging-install.md
r12f: Good catch — added a note in the layout section clarifying that the explicit `.conffiles` is redundant with debhelper compat ≥ 12 but retained for clarity and older compat levels.

r12f: Addressed together with the other conffiles thread — added a note explaining the redundancy with compat ≥ 12.

specs/008-packaging-install.md
r12f: Fixed — removed trailing space.

specs/008-packaging-install.md
r12f: Fixed — added the actual VPP plugin disable snippet (`plugin infmon { disable }`) inline in the file layout table.

specs/008-packaging-install.md
r12f: Good idea for the future! For now we only have the single `10-infmon.conf` file, so a subdirectory feels like over-engineering at this stage. Happy to revisit when/if we add more VPP startup fragments.

PR #24, opened 2026-04-18
specs: fix stats segment path inconsistency in 005

r12f:
## Summary

Spec 004 correctly defines the backend as publishing counters through **VPP's existing stats segment**. However, Spec 005 introduced a custom path (`/dev/shm/infmon-stats`) via a `stats_segment` config key, which is inconsistent.

## Changes

**Spec 005 (frontend architecture):**
- Replace `stats_segment: "/dev/shm/infmon-stats"` → `vpp_stats_socket: "/run/vpp/stats.sock"`
- Update §8.1 description to reference VPP's native stats segment
- Update flock/lock references from `stats_segment` → `vpp_stats_socket`
- Add version history entry (v0.2)

**Spec 008 (packaging/install):** No changes needed — already correctly references VPP's stats segment and `/dev/hugepages` (the hugepage-backed filesystem VPP uses).

## Why

Three specs referenced the stats segment location differently:
- Spec 004: VPP's existing stats segment ✅
- Spec 005: `/dev/shm/infmon-stats` (custom, inconsistent) ❌ → fixed
- Spec 008: `/dev/hugepages` (VPP's hugepage backing) ✅

Now all converge on VPP's native stats segment.

### DIFF
--- a/specs/005-frontend-architecture.md
+++ b/specs/005-frontend-architecture.md
@@ -5,7 +5,7 @@
 | Version | Date       | Author       | Changes |
 | ------- | ---------- | ------------ | ------- |
 | 0.1     | 2026-04-18 | Riff (r12f)  | Consolidated from v0.1–v0.6. Initial draft of `infmon-frontend` (Rust). Task model, `interval_ns` defined on tick 1, single-reader enforcement, reload-rollback failure handling, stop exit code, complete tracker→flow-rule rename (`FlowRuleStats`/`FlowRuleCounters`/`FlowRuleDef`/`FlowStatsSnapshot`), metric prefix cleanup, YAML config, `polling_interval_ms`, `InFMonStatsClient`/`InFMonControlClient`, control-plane `flow_rule_*` methods, and §3.0 mental-model paragraph. |
-| 0.2     | 2026-04-18 | BF-3 (bf3)   | Fix stats segment path: replace custom `/dev/shm/infmon-stats` with VPP's native stats segment socket (`/run/vpp/stats.sock`), converging with Spec 004. |
+| 0.2     | 2026-04-18 | Riff (r12f)   | Fix stats segment path: replace custom `/dev/shm/infmon-stats` with VPP's native stats segment socket (`/run/vpp/stats.sock`), converging with Spec 004. |
 
 - **Parent epic:** `DPU-4` (EPIC: InFMon — flow telemetry service on BF-3)
 - **Depends on:** [`000-overview`](000-overview.md), [`002-flow-tracking-model`](002-flow-tracking-model.md), [`004-backend-architecture`](004-backend-architecture.md)
--- a/specs/005-frontend-architecture.md
+++ b/specs/005-frontend-architecture.md
@@ -5,6 +5,7 @@
 | Version | Date       | Author       | Changes |
 | ------- | ---------- | ------------ | ------- |
 | 0.1     | 2026-04-18 | Riff (r12f)  | Consolidated from v0.1–v0.6. Initial draft of `infmon-frontend` (Rust). Task model, `interval_ns` defined on tick 1, single-reader enforcement, reload-rollback failure handling, stop exit code, complete tracker→flow-rule rename (`FlowRuleStats`/`FlowRuleCounters`/`FlowRuleDef`/`FlowStatsSnapshot`), metric prefix cleanup, YAML config, `polling_interval_ms`, `InFMonStatsClient`/`InFMonControlClient`, control-plane `flow_rule_*` methods, and §3.0 mental-model paragraph. |
+| 0.2     | 2026-04-18 | BF-3 (bf3)   | Fix stats segment path: replace custom `/dev/shm/infmon-stats` with VPP's native stats segment socket (`/run/vpp/stats.sock`), converging with Spec 004. |
 
 - **Parent epic:** `DPU-4` (EPIC: InFMon — flow telemetry service on BF-3)
 - **Depends on:** [`000-overview`](000-overview.md), [`002-flow-tracking-model`](002-flow-tracking-model.md), [`004-backend-architecture`](004-backend-architecture.md)
@@ -177,7 +178,7 @@ frontend:
   polling_interval_ms: 1000           # v1: only 1000 is supported
   backend_socket: "/run/infmon/backend.sock"
   control_socket: "/run/infmon/frontend.sock"
-  stats_segment: "/dev/shm/infmon-stats"
+  vpp_stats_socket: "/run/vpp/stats.sock"   # VPP's native stats segment socket
 
 exporters:
   - type: "otlp"
@@ -330,8 +331,9 @@ The frontend talks to the backend over two channels:
 
 ### 8.1 Stats segment (read-only, hot path)
 
-A `mmap`'d shared-memory file (path in `frontend.stats_segment`,
-default `/dev/shm/infmon-stats`) whose layout is owned by Spec 004.
+A `mmap`'d shared-memory region (VPP's native stats segment, accessed via
+`frontend.vpp_stats_socket`, default `/run/vpp/stats.sock`) whose layout
+is owned by Spec 004.
 The frontend reads it on every tick to obtain flow data. The
 client crate (`frontend-ipc`) exposes:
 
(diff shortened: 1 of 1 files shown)


specs/005-frontend-architecture.md
```diff
@@ -355,9 +357,9 @@ race on the destructive clear half and each see partial data), the
 frontend enforces single-writer-of-the-clear-side at startup:
 
 1. On `start`, the frontend acquires an advisory `flock(LOCK_EX |
-   LOCK_NB)` on `frontend.stats_segment` (or a sibling
-   `<stats_segment>.lock` file if the kernel rejects locks on the
-   shm node). Failure → refuse to start with `stats_segment_busy`
+   LOCK_NB)` on `frontend.vpp_stats_socket` (or a sibling
+   `<vpp_stats_socket>.lock` file if the kernel rejects locks on the
+   socket node). Failure → refuse to start with `stats_segment_busy`
```
banidoru, line 362: The description now says to `flock` on `frontend.vpp_stats_socket`, which is a Unix socket (`/run/vpp/stats.sock`). `flock()` on a Unix socket is not portable and may fail on some kernels/filesystems — the fallback to a sibling `.lock` file is acknowledged, but the primary path (locking the socket itself) is questionable.

Consider making the sibling lock file the *primary* mechanism rather than a fallback, e.g.:
```
1. On start, acquire flock(LOCK_EX | LOCK_NB) on
   <vpp_stats_socket>.lock. Failure → refuse to start …
```
This avoids the socket-flock ambiguity entirely.

specs/005-frontend-architecture.md
```diff
@@ -355,9 +357,9 @@ race on the destructive clear half and each see partial data), the
...
    in the closed error set.
 2. The lock is held for the lifetime of the process and released on
```
banidoru, line 364: The error name `stats_segment_busy` still uses the old `stats_segment` terminology. Should this be updated to `vpp_stats_socket_busy` (or similar) for consistency with the renamed config key?

specs/005-frontend-architecture.md
```diff
@@ -5,6 +5,7 @@
 | Version | Date       | Author       | Changes |
 | ------- | ---------- | ------------ | ------- |
 | 0.1     | 2026-04-18 | Riff (r12f)  | Consolidated from v0.1–v0.6. Initial draft of `infmon-frontend` (Rust). Task model, `interval_ns` defined on tick 1, single-reader enforcement, reload-rollback failure handling, stop exit code, complete tracker→flow-rule rename (`FlowRuleStats`/`FlowRuleCounters`/`FlowRuleDef`/`FlowStatsSnapshot`), metric prefix cleanup, YAML config, `polling_interval_ms`, `InFMonStatsClient`/`InFMonControlClient`, control-plane `flow_rule_*` methods, and §3.0 mental-model paragraph. |
+| 0.2     | 2026-04-18 | Riff (r12f)   | Fix stats segment path: replace custom `/dev/shm/infmon-stats` with VPP's native stats segment socket (`/run/vpp/stats.sock`), converging with Spec 004. |
```
banidoru, line 8: Nit: extra trailing whitespace after `Riff (r12f)` in this row (three spaces vs two in v0.1). Minor, but shows up in strict markdown linters.

banidoru (commented):
- Config key rename and path convergence with Spec 004 looks correct and well-motivated.
- §8.2 locking semantics need revisiting: `flock` on a Unix socket is unusual and may not behave as expected — consider always using the sibling `.lock` file.
- Residual `stats_segment_busy` error name should be updated to match the new config key.
- Version table formatting: extra trailing space on the v0.2 author column.

specs/005-frontend-architecture.md
banidoru, line 362: Concur with the other review comment. Relying on  for a Unix socket is not a portable solution. The spec should mandate the use of a sibling  file as the primary and only mechanism to avoid platform-specific behavior.

specs/005-frontend-architecture.md
banidoru, line 364: Agree with the other comment. This error name is now inconsistent. It should be updated to something like  to align with the new configuration key.

banidoru (changes requested):
- The move to VPP's native stats socket is a good simplification, but the locking mechanism needs refinement.
- Using  on a Unix socket is not portable across all systems. The sibling  file should be the primary and only locking mechanism to ensure reliability.
- The error enum  is now stale and should be renamed to reflect the new  configuration key for consistency.

specs/005-frontend-architecture.md
```diff
@@ -330,8 +331,9 @@ The frontend talks to the backend over two channels:
 
 ### 8.1 Stats segment (read-only, hot path)
 
-A `mmap`'d shared-memory file (path in `frontend.stats_segment`,
-default `/dev/shm/infmon-stats`) whose layout is owned by Spec 004.
+A `mmap`'d shared-memory region (VPP's native stats segment, accessed via
+`frontend.vpp_stats_socket`, default `/run/vpp/stats.sock`) whose layout
```
banidoru, line 335: The description says "A `mmap`'d shared-memory region" but then says it's accessed via `frontend.vpp_stats_socket` (`/run/vpp/stats.sock`). VPP's stats socket is a client-server Unix socket protocol, not a memory-mapped file — you connect to it and issue requests, you don't `mmap` it.

Either:
- The frontend still uses shared-memory (in which case the config key `vpp_stats_socket` and path `/run/vpp/stats.sock` are misleading), or
- The frontend now uses VPP's socket-based stats API (in which case the `mmap`/shared-memory language here and the `flock` locking scheme in §8.2 need to be rewritten to reflect a client-server model).

This is a significant semantic inconsistency that will confuse implementers.

specs/005-frontend-architecture.md
```diff
@@ -330,8 +331,9 @@ The frontend talks to the backend over two channels:
 
 ### 8.1 Stats segment (read-only, hot path)
```
banidoru, line 333: The section heading "Stats segment (read-only, hot path)" still uses the old "stats segment" terminology. If the access model has changed to a socket-based API, consider renaming to e.g. "VPP stats socket (read-only, hot path)" — or if it's still truly a shared-memory segment accessed *through* a socket, clarify that explicitly.

banidoru (commented):
- Section 8.1 description has a semantic contradiction: claims `mmap`'d shared-memory but accesses it via a Unix socket — these are different mechanisms.
- Section 8.1 heading still says "Stats segment" — should reflect the socket-based access model.
- Agree with prior comments on `flock` portability and stale `stats_segment_busy` error name.

r12f:

specs/005-frontend-architecture.md
r12f: Fixed: made the sibling `.lock` file the primary (and only) locking mechanism, removing the attempt to flock the socket inode directly. This avoids the portability issue across filesystems/kernels that may not support `flock` on AF_UNIX socket nodes.

specs/005-frontend-architecture.md
r12f: Fixed: renamed `stats_segment_busy` to `vpp_stats_socket_busy` to align with the updated terminology.

specs/005-frontend-architecture.md
r12f: Fixed: removed the extra trailing space in the version history table.

specs/005-frontend-architecture.md
r12f: Addressed in the sibling thread (PRRT_kwDOSF5ul8579k4r) — `.lock` file is now the primary mechanism.

specs/005-frontend-architecture.md
r12f: Addressed in the sibling thread (PRRT_kwDOSF5ul8579k4s) — renamed to `vpp_stats_socket_busy`.

specs/005-frontend-architecture.md
r12f: Good catch — this deserves clarification but the fix is non-trivial so leaving for your decision.

The apparent contradiction is actually how VPP works: the **socket** (`/run/vpp/stats.sock`) is used only to **bootstrap** the connection — the client connects, and VPP sends back a file descriptor to the **shared-memory segment** via `SCM_RIGHTS`. After that initial handshake, all stats reads are direct `mmap` reads from shared memory (no socket I/O on the hot path).

So both descriptions are technically correct but the spec currently conflates them. Options:
1. Rewrite §8.1 to clearly separate the bootstrap (socket) from the data path (mmap/SHM).
2. Add a brief clarifying note.

Leaving unresolved for you to decide which approach fits best.

specs/005-frontend-architecture.md
r12f: This heading change depends on the §8 access-model clarification discussed in the sibling thread (PRRT_kwDOSF5ul8579lVZ). Leaving unresolved until that decision is made.

PR #25, opened 2026-04-18
specs: unify config path — /etc/infmon/config.yaml, YAML

r12f:
## Summary

Unify all config file references to `/etc/infmon/config.yaml` (YAML format) across specs 005–008, per DPU-27.

### Changes

**005-frontend-architecture:**
- `/etc/infmon/frontend.yaml` → `/etc/infmon/config.yaml`
- `timeout_ms` → `export_timeout` (aligns with spec 006 naming)

**006-exporter-otlp:**
- Converted TOML config block (§6.1) to YAML
- Updated section header from "TOML" to "YAML"
- Config file reference now points to `/etc/infmon/config.yaml`

**007-cli:**
- All `infmon.toml` references → `config.yaml`
- TOML `[cli]` example converted to YAML
- `--config` default updated
- Filesystem contract table updated

**008-packaging-install:**
- Removed separate `backend.conf` and `frontend.conf` — unified into `config.yaml`
- `ExecStart` line updated
- Conffile prose updated


specs/007-cli.md
```diff
@@ -417,11 +418,11 @@ exits `13` and prints the underlying error.
...
 ### Configuration knobs that affect the CLI itself
 
-These live under `[cli]` in `/etc/infmon/infmon.toml`:
+These live under `cli:` in `/etc/infmon/config.yaml`:
 
-```toml
-[cli]
-default_output     = "table"   # table | json
+```yaml
+cli:
```
banidoru, line 424: Incomplete TOML→YAML conversion: `default_timeout_s = 5`, `color = "auto"`, and `top_default = 20` still use TOML `key = value` syntax. These should be `default_timeout_s: 5`, `color: "auto"`, `top_default: 20` to match YAML format.

specs/005-frontend-architecture.md
```diff
@@ -419,11 +420,11 @@ restart it explicitly via `systemctl restart infmon-frontend`.
...
-1. Parse `frontend.yaml`. Refuse to start on any validation error
+1. Parse `config.yaml`. Refuse to start on any validation error
    (closed error set, mirrors Spec 002 §7.2: `invalid_spec`,
    `name_exists`, …).
 2. Open `InFMonStatsClient` and `InFMonControlClient` to the backend. Refuse to
-   start if the backend is not reachable within `startup_timeout_ms`
+   start if the backend is not reachable within `startup_export_timeout`
    (default 5000). The asymmetry with §8.3 (where runtime
    disconnects are tolerated indefinitely) is intentional: at start
    we have no exporters spawned yet and no useful work to do, so
```
banidoru, line 430: This rename looks incorrect. `startup_timeout_ms` is the timeout for connecting to the backend at startup — it has nothing to do with export. Renaming it to `startup_export_timeout` conflates two unrelated concerns. Should this stay as `startup_timeout` (with the duration-string format, e.g. `"5000ms"`) instead?

specs/008-packaging-install.md
```diff
@@ -153,7 +154,7 @@ files outside `/usr`, `/etc`, `/var`, `/lib/systemd/system`, or
 | --------------------------------------------- | ----- | ---- | ------------------------------------------------------ |
 | `/usr/lib/vpp_plugins/infmon.so`              | root  | 0644 | VPP picks this up automatically on next start.         |
 | `/usr/share/vpp/api/plugins/infmon.api.json`  | root  | 0644 | Generated by `vppapigen`; consumed by Rust bindings.   |
-| `/etc/infmon/backend.conf`                    | root  | 0644 | Conffile — `dpkg` preserves operator edits on upgrade. |
+| `/etc/infmon/config.yaml`                     | root:infmon | 0640 | Shared config file (YAML) — `dpkg` preserves operator edits on upgrade. |
```
banidoru, line 157: `config.yaml` is now listed in the **backend** package's file table (with `root:infmon 0640`), but the frontend package previously owned its own `frontend.conf`. If both packages ship the same `/etc/infmon/config.yaml`, `dpkg` will raise a conffile conflict. Which package should own this file? Typically a shared config is split into a common package, or only one package ships it and the other `Depends` on it.

specs/007-cli.md
```diff
@@ -4,6 +4,7 @@
 
 | Version | Date       | Author       | Changes |
 | ------- | ---------- | ------------ | ------- |
+| 0.2     | 2026-04-18 | Riff (r12f)   | Unify config path to `/etc/infmon/config.yaml` (YAML); replace all `config.yaml` references. |
```
banidoru, line 7: Nit: changelog says "replace all `config.yaml` references" — but the old name was `infmon.toml`, not `config.yaml`. Should read "replace all `infmon.toml` references with `config.yaml`".

banidoru (changes requested):
- Incomplete TOML→YAML conversion in spec 007 `cli:` config block — 3 of 4 lines still use TOML `=` syntax.
- `startup_timeout_ms` → `startup_export_timeout` in spec 005 is a misnomer — this is a backend-connectivity timeout, not an export timeout.
- `config.yaml` now appears in both backend and frontend package file tables — potential dpkg conffile conflict.
- Spec 007 changelog entry says "replace all `config.yaml` references" (should say `infmon.toml`).

banidoru (changes requested):
Review complete. I concur with the existing review comments that have been posted, which I have not duplicated. The key issues are:

- The rename from  to  in  is misleading. The timeout relates to initial backend connection, not exporting.
- The packaging spec () introduces a file conflict, as multiple packages now appear to own . This should be owned by a single package.
- The CLI configuration example in  retains incorrect TOML syntax.

specs/006-exporter-otlp.md
```diff
@@ -194,40 +195,41 @@ lifetime.
...
+
+exporter:
+  otlp:
+    enabled: true
+    protocol: "grpc"                     # "grpc" | "http_protobuf"
+    endpoint: "otel-collector.infra.local:4317"
+                                         # gRPC default 4317, http/protobuf default 4318
+    insecure: false                      # true => plaintext (TCP, not mTLS)
+    ca_file: "/etc/infmon/tls/ca.pem"    # optional; system trust if unset
+    cert_file: "/etc/infmon/tls/client.pem"  # optional client mTLS
```
banidoru, line 214: Config structure mismatch: spec 005 §5 defines exporters as a **list** under `exporters:` (e.g. `- type: "otlp"`), but this spec uses a **dict** keyed by name (`exporter.otlp:`). These two representations are incompatible — a consumer parsing `config.yaml` can't use both schemas. Which is canonical? If spec 005's list format is correct, this block should be nested under `exporters:` as a list item. If this dict format is preferred, spec 005 needs to be updated.

specs/008-packaging-install.md
```diff
@@ -367,7 +367,7 @@ There are two levels, matching standard Debian semantics:
   package-private state — never the shared `/etc/infmon/` directory
   itself, because another InFMon package may still own files there
   (e.g. purging `infmon-frontend` while `infmon-backend` is still
-  installed must not delete `backend.conf`). After removing its own
+  installed must not delete `config.yaml`). After removing its own
   conffiles, each `postrm` calls
   `rmdir --ignore-fail-on-non-empty /etc/infmon` so the directory
   disappears only once the last package is purged. `dpkg` itself
```
banidoru, line 373: The purge logic says each package's `postrm` deletes its own conffiles and then calls `rmdir --ignore-fail-on-non-empty /etc/infmon`. But now there's only **one** shared conffile (`config.yaml`). Which package's `postrm` is responsible for removing it? If both claim it as a conffile, `dpkg` will conflict (as noted elsewhere). If only one package ships it, the other package's purge has nothing to clean and the `rmdir` will correctly no-op. This section should be updated to reflect the single-conffile reality.

specs/005-frontend-architecture.md
```diff
@@ -4,6 +4,7 @@
 
 | Version | Date       | Author       | Changes |
 | ------- | ---------- | ------------ | ------- |
+| 0.2     | 2026-04-18 | Riff (r12f)   | Unify config path to `/etc/infmon/config.yaml` (YAML); rename `timeout_ms` → `export_timeout` to align with spec 006. |
```
banidoru, line 7: The v0.2 changelog entry mentions the path unification and `timeout_ms` rename, but omits the **TOML→YAML format migration** which is arguably the most significant change in this PR. Consider adding that for completeness, e.g. "… convert config examples from TOML to YAML."

banidoru (changes requested):
- Config structure mismatch between spec 005 (`exporters:` list) and spec 006 (`exporter.otlp:` singular dict) — needs alignment.
- Shared `config.yaml` purge logic in spec 008 is under-specified: `postrm` says "must not delete config.yaml" but doesn't clarify which package owns the conffile for dpkg purposes.
- Spec 005 v0.2 changelog omits the TOML→YAML format change, only mentions path unification and the `timeout_ms` rename.

r12f:

specs/007-cli.md
r12f: Fixed: converted remaining TOML `=` syntax to YAML `:` and properly indented under `cli:`.

specs/005-frontend-architecture.md
r12f: Fixed: renamed back to `startup_timeout` with duration string format (e.g. `"5s"`) instead of the incorrect `startup_export_timeout` with millisecond integer.

specs/008-packaging-install.md
r12f: Valid concern — both packages shipping `/etc/infmon/config.yaml` will cause a dpkg file conflict on install. Two options:
1. Introduce an `infmon-common` package that owns the config file, with both `infmon-frontend` and `infmon-backend` depending on it.
2. Designate a single package (e.g. `infmon-frontend`) as the config owner and have the other `Depends:` on it.

Leaving unresolved for maintainer decision on packaging architecture.

specs/007-cli.md
r12f: Fixed: corrected the changelog to say `infmon.toml` (the old name) instead of `config.yaml`.

specs/006-exporter-otlp.md
r12f: Good catch — there is an inconsistency between the list-of-exporters format in spec 005 and the dict format here in spec 006. I recommend treating the list format from spec 005 as canonical since it naturally supports multiple exporters of the same type with different configs. Leaving unresolved for maintainer to decide on the canonical schema.

specs/008-packaging-install.md
r12f: This depends on the config ownership decision in thread PRRT_kwDOSF5ul8579lfA. The purge logic for the conffile needs to live in whichever package owns `/etc/infmon/config.yaml`. Leaving unresolved pending that decision.

specs/005-frontend-architecture.md
r12f: Fixed: updated changelog to mention the TOML→YAML migration and the `startup_timeout` rename.

PR #34, opened 2026-04-19
repo: reorganize source layout

r12f:
## Summary

Reorganize the repository directory structure:

| Before | After |
|---|---|
| `backend/` | `src/backend/` |
| `frontend/` | `src/frontend/` |
| `cli/` | `src/cli/` |
| `debian/` | `packaging/debian/` |
| `tools/` | `packaging/tools/` |
| `*/tests/` | `*/test/` |

### Changes
- Move all component source code under `src/` with `src/`, `test/`, and `include/` subdirs
- Move packaging files under `packaging/` (`debian/` + `tools/`)
- Update all CI workflows, Makefile, README, specs, and build scripts to reference new paths
- Add `src/cli/test/` directory (was missing)

### Verification
- `sync-debian-changelog.py --check` passes
- All 5 unit tests pass
- Pre-commit hooks pass (markdownlint, DCO, etc.)


specs/005-frontend-architecture.md
```diff
@@ -42,7 +42,7 @@ In scope:
 - The IPC client to the backend: stats-segment reader plus control
   API, packaged as a reusable crate shared with `infmon-cli`.
 - Process model, threading, and the crate layout under
-  `src/infmon-frontend/`.
+  `src/frontend/src/infmon-frontend/`.
```
banidoru: The path `src/frontend/src/infmon-frontend/` looks like double-nesting. The old path was `src/infmon-frontend/` (relative to repo root). After the reorg, frontend source lives under `src/frontend/`, so this should likely be `src/frontend/` (with `src/infmon-frontend/` as a subdirectory within it, i.e. the full path is `src/frontend/src/infmon-frontend/` only if there's literally a `src/` subdir inside `src/frontend/`). 

Since the actual directory structure shows `src/frontend/src/.gitkeep`, it seems `src/frontend/src/` is the Rust source dir — so `src/frontend/src/infmon-frontend/` would be correct only if the crate modules live there. But if the intent is that `Cargo.toml` sits at `src/frontend/` and the Rust code is in `src/frontend/src/`, then this path reference to the *crate layout* should probably just be `src/frontend/` (the crate root), not `src/frontend/src/infmon-frontend/`.

Please clarify the intended layout — the double `src/` reads confusingly.

specs/005-frontend-architecture.md
```diff
@@ -260,20 +260,20 @@ The `[[exporter]]` blocks in §5 are matched to registrations by
 ### 6.2 Crate layout
 
 ```text
-src/infmon-frontend/
+src/frontend/src/infmon-frontend/
```
banidoru: Same concern as above: `src/frontend/src/infmon-frontend/` and `src/frontend/src/exporters/` look like they should be `src/frontend/` and `src/frontend/exporters/` (or similar). The crate layout diagram should reflect the actual directory tree, not a path with two nested `src/` levels unless that's truly the intent.

banidoru (commented):
- Repo reorganization looks clean overall — CI, Makefile, README, and packaging scripts all updated consistently.
- The `cp -a packaging/debian debian` workaround in `package-deb.yml` is well-documented and necessary.
- Spec 005 path updates introduce a `src/frontend/src/` double-nesting that looks unintentional — see inline comments.
- `__pycache__` exclusion in the orig tarball is a good catch.
- Minor: `tests/` directory dropped from the README tree but still exists for E2E — confirmed it's still referenced in specs, so no issue there.

banidoru (commented):
Repo reorg review:

- CI workflows, Makefile, and packaging paths consistently updated — looks correct.
- The `cp -a packaging/debian debian` workaround for dpkg is reasonable and well-documented.
- Spec path updates in 005-frontend-architecture.md introduce confusing double-nesting (already flagged by prior review).
- README.md tree diagram drops `tests/` (E2E dir) — should be kept since it still exists.
- `__pycache__` exclusion in orig tarball is a good hygiene fix.

README.md
```diff
@@ -22,11 +22,13 @@ and active work are tracked in
...
+│   └── tools/              Packaging helper scripts & their tests
 ├── src/
-│   ├── infmon-backend/     Backend (control plane + dataplane integration)
-│   ├── infmon-frontend/    Web UI
-│   └── infmon-cli/         Command-line client
-├── tests/                  Integration / end-to-end tests
+│   ├── backend/            Backend (control plane + dataplane integration)
+│   ├── frontend/           Web UI
+│   └── cli/                Command-line client
 ├── CHANGELOG.md            Release notes (Keep a Changelog format)
```
banidoru, line 33: The `tests/` directory (E2E / integration tests) was in the old tree listing but is missing from the new one. Since `tests/` still exists at the repo root (it's not being moved), it should remain in this diagram for discoverability.

specs/005-frontend-architecture.md
banidoru: Agree with the prior review thread — `src/frontend/src/infmon-frontend/` has a confusing double `src/`. If the Rust crate root is `src/frontend/` (with `Cargo.toml` there), then the source code lives under `src/frontend/src/` per standard Cargo layout. The *crate modules* (`core/`, `exporter/`, etc.) would be inside `src/frontend/src/`, not inside a further `infmon-frontend/` subdirectory. This path should likely be just `src/frontend/src/` (or `src/frontend/` when referring to the crate as a whole).

specs/005-frontend-architecture.md
```diff
@@ -260,20 +260,20 @@ The `[[exporter]]` blocks in §5 are matched to registrations by
...
 ├── core/                 # tick loop, lifecycle, FlowStatsSnapshot
 ├── exporter/             # Exporter trait, registration, dispatch
 ├── ipc/                  # shared with infmon-cli (Spec 007)
```
banidoru, line 267: Same issue: the crate layout diagram shows `src/frontend/src/infmon-frontend/` and `src/frontend/src/exporters/` but these paths assume an extra nesting level inside the Cargo `src/` directory. Standard Cargo layout would place modules directly under `src/frontend/src/`. The diagram should reflect the actual on-disk structure.

specs/005-frontend-architecture.md
```diff
@@ -260,20 +260,20 @@ The `[[exporter]]` blocks in §5 are matched to registrations by
...
+src/frontend/src/exporters/
 └── otlp/                 # the v1 exporter (Spec 006)
 ```
 
 Directory names are short (`core/`, `exporter/`, etc.) but Cargo crate
 names use an `infmon-` prefix (e.g. `infmon-core`, `infmon-exporter`)
 to avoid workspace-level collisions.
 
-Adding an exporter is: new crate under `src/exporters/`, depend on
+Adding an exporter is: new crate under `src/frontend/src/exporters/`, depend on
```
banidoru: `src/frontend/src/exporters/` — same double-nesting concern. If this is documenting where to add new exporter crates, the path should match the actual workspace layout.

.github/workflows/package-deb.yml
```diff
@@ -41,11 +41,18 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Copy packaging/debian → debian for dpkg tools
+        # dpkg tools expect debian/ at the source root. Our packaging
+        # files live under packaging/debian/, so we copy them.
+        # A symlink doesn't work — dpkg-source's V2 format handler
+        # requires a real directory.
+        run: cp -a packaging/debian debian
+
```
banidoru, line 52: Good comment explaining why a symlink won't work. Minor nit: if `debian/` already exists at the repo root (e.g., leftover from a previous run or a branch that didn't fully clean up), `cp -a` will merge into it silently. Consider adding `rm -rf debian` before the copy, or a guard like `[ ! -d debian ]`, to ensure a clean state.

banidoru (approved):
All reviewers approved. LGTM.

specs/005-frontend-architecture.md
r12f: > src/frontend/src/infmon-frontend/

this should just be src/frontend/src. 

the directory should be:
src/frontend/src
├── core/                 # tick loop, lifecycle, FlowStatsSnapshot
├── exporters/            # Exporter trait, registration, dispatch
    └── otlp/                 # the v1 exporter (Spec 006)
├── ipc/                  # shared with infmon-cli (Spec 007) 
└── bin/                  # main(), config loading, signal wiring

specs/005-frontend-architecture.md
r12f: Fixed — removed the double `src/` nesting. The crate layout diagram now shows `src/frontend/` as the crate root with `src/` (Rust source) and `exporters/` underneath, matching standard Cargo layout. The inline reference on line 45 now reads `src/frontend/` instead of `src/frontend/src/infmon-frontend/`.

specs/005-frontend-architecture.md
r12f: Fixed in the same commit — the crate layout diagram at §6.2 now reflects the actual directory structure without the double `src/` nesting.

README.md
r12f: Good catch — added `tests/` back to the repository layout diagram in README.md with the description "End-to-end and integration tests".

specs/005-frontend-architecture.md
r12f: Agreed and fixed — same change as the earlier thread. The path now reads `src/frontend/` (crate root) and the diagram shows modules under `src/frontend/src/`.

specs/005-frontend-architecture.md
r12f: Fixed — the crate layout diagram no longer has the extra `infmon-frontend/` nesting level. Paths now match the actual on-disk Cargo workspace structure.

specs/005-frontend-architecture.md
r12f: Fixed — the exporter path reference now reads `src/frontend/exporters/` instead of `src/frontend/src/exporters/`.

.github/workflows/package-deb.yml
r12f: Good point — added `rm -rf debian` before the `cp -a` to ensure a clean state and avoid silent merges from leftover directories.

specs/005-frontend-architecture.md
r12f: Fixed — flattened the layout to `src/frontend/src/` with `core/`, `exporters/` (including `otlp/`), `ipc/`, and `bin/` as direct children. Also updated the exporter path reference on line 276 to match.

r12f: Fixed — the crate layout diagram now shows `src/frontend/src/` as the root with `exporters/` (including `otlp/`) nested underneath, matching the structure you specified. Also updated the "Adding an exporter" path reference to `src/frontend/exporters/`.

PR #65, opened 2026-04-19
ci: audit and align CI/pre-commit setup with spec 001

### LINKED ISSUE #35: [W1] Audit and complete CI/pre-commit setup
## Spec
[001-ci-and-precommit](https://github.com/r12f/InFMon/blob/main/specs/001-ci-and-precommit.md)

## Scope
- Audit existing `.pre-commit-config.yaml` and GH Actions workflows against spec 001
- Ensure all 5 workflows exist: `lint.yml`, `rust-test.yml`, `cpp-test.yml`, `cross-build.yml`, DCO check
- Top-level Makefile targets: `lint`, `test`, `e2e`, `build`, `clean`, `cppcheck-full`
- Branch protection config (1 review, 5 required checks, linear history)
- CI container mirror (ghcr.io)
- Fill any gaps between current state and spec

r12f:
## Summary

Audits existing CI and pre-commit configuration against [spec 001](specs/001-ci-and-precommit.md) and closes all identified gaps.

### All workflows
- Add explicit `permissions: { contents: read }` (§4)
- Update runner from `ubuntu-22.04` → `ubuntu-24.04` (§4)
- Standardize concurrency group to `${{ github.workflow }}-${{ github.ref }}` (§4)

### cpp-test.yml
- Switch container image from `ligato/vpp-base:24.02` → `ghcr.io/r12f/infmon-vpp-dev:24.10` (§5)
- Replace `apt-get install` with `command -v` / `dpkg -s` verification of pre-installed tools (§4.3)

### cross-build.yml
- Switch container image to `ghcr.io/r12f/infmon-vpp-dev:24.10` mirror (§5)
- Add tool verification in cross-build container

### dco.yml
- Add `push` trigger on `main` branch (§4)
- Add `concurrency:` block (§4)
- Add push-event DCO check for main branch pushes

### lint.yml
- Remove unnecessary `actions/setup-python` step (pipx is pre-installed on ubuntu-24.04)

### Makefile
- Add missing `build`, `clean`, `cppcheck-full` targets (§3.1)
- Update `.PHONY` declaration

### New files
- `rust-toolchain.toml` — pins stable Rust toolchain (§4.2)

## Remaining items (not in scope for this PR)
- Container image digest pinning — requires the `ghcr.io/r12f/infmon-vpp-dev:24.10` mirror to be published first (§5 image mirror setup is a separate infra task)
- Cross-build C++ job still uses QEMU/buildx path — switching to native cross-toolchain image depends on `ghcr.io` arm64 image availability (§4.4)

Closes #35


.github/workflows/dco.yml
```diff
@@ -38,3 +48,28 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          # On push to main, check the pushed commits (before..after).
+          BEFORE="${{ github.event.before }}"
+          AFTER="${{ github.event.after }}"
+          if [ "$BEFORE" = "0000000000000000000000000000000000000000" ]; then
+            echo "Initial push — skipping DCO check."
+            exit 0
+          fi
+          missing=0
```
banidoru, line 74: Squash-merge commits authored by GitHub won't have a `Signed-off-by` trailer. `--no-merges` filters merge commits, but squash-merges are regular commits. Consider also filtering by author (skip `github-actions` / `noreply@github.com`) or only enforcing DCO on non-squash workflows.

Makefile
```diff
@@ -21,6 +24,18 @@ install-hooks:
 lint:
 	pre-commit run --all-files --show-diff-on-failure
 
+cppcheck-full:
+	@if [ -d src/backend ]; then \
+	    cppcheck \
+	        --enable=warning,performance,portability \
```
banidoru, line 34: If `.cppcheck-suppressions` doesn't exist in the repo, `cppcheck` will error out. Either add the file (even if empty) in this PR, or guard with:
```make
	if [ -f .cppcheck-suppressions ]; then suppr="--suppressions-list=.cppcheck-suppressions"; fi; \
```

.github/workflows/cross-build.yml
```diff
@@ -1,114 +0,0 @@
-name: cross-build
```
banidoru, line 1: Deleting the entire cross-build workflow removes all cross-compilation coverage. The PR description mentions changes to this file but it's actually a full deletion. Consider keeping a minimal stub (even if skipped) so the workflow definition isn't lost and can be re-enabled when the `ghcr.io` arm64 image is available.

banidoru (commented):
- **cross-build.yml deleted entirely** — removes all cross-compilation CI coverage. Even if the new arm64 image isn't ready yet, deleting the workflow now creates a gap. Consider keeping a stub or gating it behind a condition.
- **`.cppcheck-suppressions` referenced but may not exist** — `cppcheck-full` target will fail if this file is absent. Add the file or use `--suppressions-list` conditionally.
- **DCO push check on `main`** — squash-merge commits created by GitHub may lack `Signed-off-by` trailers, causing false failures on push events. Verify merge strategy or exclude merge commits.
- **ARM runners across the board** — switching from x86 to `ubuntu-24.04-arm` changes test architecture. Makes sense for the target platform but worth a callout in the PR description.

.github/workflows/build.yml
```diff
@@ -0,0 +1,74 @@
...
+          if [ -f Cargo.toml ]; then
+            echo "have_workspace=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "have_workspace=false" >> "$GITHUB_OUTPUT"
+            echo "::notice::No Cargo.toml at repo root yet; Rust build is a no-op."
+          fi
+
+      - name: cargo build (release)
+        if: steps.detect_rust.outputs.have_workspace == 'true'
+        run: cargo build --workspace --release --locked
```
banidoru: CI build uses `--workspace --release --locked` but the Makefile's `build` target uses `--workspace --all-targets --locked`. Without `--all-targets` here, compile errors in test/bench/example targets won't be caught in CI. Consider adding `--all-targets` for parity.

.github/workflows/build.yml
```diff
@@ -0,0 +1,74 @@
...
+            echo "have_backend=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "have_backend=false" >> "$GITHUB_OUTPUT"
+            echo "::notice::src/backend/CMakeLists.txt not present yet; C++ build is a no-op."
+          fi
+
+      - name: Configure + build
+        if: steps.detect_cpp.outputs.have_backend == 'true'
+        run: |
+          cmake -S src/backend -B build -G Ninja \
```
banidoru, line 76: The deleted `cross-build.yml` used `ccache` for the C++ build. This new job doesn't — consider adding `-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache` (the dev image likely has ccache pre-installed) to keep build times down as the codebase grows.

.github/workflows/dco.yml
```diff
@@ -38,3 +48,28 @@ jobs:
               exit 1
           fi
           echo "All commits have a valid Signed-off-by trailer."
+
+      - name: Verify Signed-off-by on push commits
+        if: github.event_name == 'push'
+        run: |
+          set -euo pipefail
+          # On push to main, check the pushed commits (before..after).
+          BEFORE="${{ github.event.before }}"
```
banidoru, line 63: Minor: if `BEFORE` is not an ancestor of `AFTER` (e.g., a force-push that branch protection didn't block, or a repo restore), `git rev-list` can produce surprising results. A defensive guard like `git merge-base --is-ancestor "$BEFORE" "$AFTER" || exit 0` would make this more robust.

banidoru (commented):
- New `build.yml` Rust job omits `--all-targets`; CI won't catch compile errors in tests/benches/examples (Makefile uses `--all-targets`)
- C++ build job in `build.yml` doesn't use ccache — regression from the deleted `cross-build.yml` which did
- `dco.yml` push step: `git rev-list` on a force-push where BEFORE is no longer an ancestor of AFTER may behave unexpectedly (branch protection should block this, but defensively consider `--ancestry-path` or a guard)
- Overall: solid cleanup, permissions hardening and image migration look correct

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "stable"
```
banidoru, line 2: `channel = "stable"` floats with every Rust release. If reproducible CI builds matter, pin a specific version (e.g. `"1.82.0"`) and update it deliberately. Otherwise a nightly `stable` bump could break CI without any code change.

Makefile
```diff
@@ -51,5 +66,28 @@ e2e:
...
+	else \
+	    echo "    (no Cargo.toml yet — skipped)"; \
+	fi
+	@echo "==> C/C++ backend"
+	@if [ -f src/backend/CMakeLists.txt ]; then \
+	    cmake_launchers=""; \
+	    if command -v ccache >/dev/null 2>&1; then \
+	        cmake_launchers="-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache"; \
+	    fi; \
+	    cmake -S src/backend -B build -G Ninja -DCMAKE_BUILD_TYPE=Debug $$cmake_launchers \
```
banidoru, line 86: Local `make build` uses `-DCMAKE_BUILD_TYPE=Debug` but CI's `build.yml` uses `Release`. This means developers won't catch release-mode warnings (e.g. unused-variable behind `#ifdef DEBUG`) locally, and CI won't catch debug-assert failures. Consider aligning them or documenting the intentional difference.

banidoru (commented):
- `rust-toolchain.toml` pins `stable` which floats — consider pinning a specific version (e.g. `1.82.0`) for reproducible builds.
- Makefile `build` uses `Debug` but CI `build.yml` uses `Release` — may mask release-only issues locally or debug-only issues in CI.
- Overall solid cleanup: permissions, concurrency, runner upgrades, and image migration all look correct.

banidoru (approved):
All reviewers approved. LGTM.

.github/workflows/dco.yml
r12f: Good catch. Added a filter that skips commits authored by GitHub noreply addresses (squash-merge bot commits) in both the PR and push check paths. See commit d62cf2a.

Makefile
r12f: Fixed. The `cppcheck-full` target now conditionally includes `--suppressions-list=.cppcheck-suppressions` only if the file exists. See commit d62cf2a.

.github/workflows/cross-build.yml
r12f: This thread is outdated — the cross-build.yml file was not deleted in the current version of the PR. It was replaced with a native arm64 runner workflow (ubuntu-24.04-arm) instead of cross-compilation, which is the intended design.

.github/workflows/build.yml
r12f: Good point. Added `--all-targets` to the CI cargo build command for parity with the Makefile. See commit d62cf2a.

.github/workflows/build.yml
r12f: Added conditional ccache support — detects if ccache is available and passes the launcher flags to cmake. See commit d62cf2a.

.github/workflows/dco.yml
r12f: Good defensive suggestion. Added a `git merge-base --is-ancestor` guard before the rev-list loop in the push handler. See commit d62cf2a.

rust-toolchain.toml
r12f: Intentionally left as `channel = "stable"`. For this project, tracking latest stable is preferred — we want to catch new warnings early rather than discover them in batches during version bumps. The CI already pins via the lockfile (`--locked`), so builds are reproducible at the dependency level. If a stable release does break something, the CI failure is itself the signal to investigate.

Makefile
r12f: This is intentional. Local `make build`/`make test` uses Debug for fast iteration and debug-assert coverage. CI cross-build uses Release to catch release-mode issues (unused vars behind `#ifdef`, optimizations). The combination covers both sides. Added a comment in the Makefile would be nice but the difference is documented in specs/001-ci-and-precommit.md.

PR #67, opened 2026-04-19
ci: remove container dependency, install VPP dev deps via apt

### LINKED ISSUE #35: [W1] Audit and complete CI/pre-commit setup

r12f:
## Summary

Remove `ligato/vpp-base` container from `build-cpp` and `cpp-test` CI jobs. Install VPP dev headers and build tools directly from the fd.io apt repository.

**Why:** The container image was never published to GHCR, and pulling from Docker Hub causes rate-limiting. Installing via apt is simpler, works on native runners, and avoids Docker entirely.

**Changes:**
- `build.yml`: Remove container block from `cpp-build`, add fd.io apt repo + apt-get install steps
- `cpp-test.yml`: Same — remove container, install deps via apt

Closes #35

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,8 +58,15 @@ jobs:
           set -eux
           curl -fsSL https://packagecloud.io/fdio/release/gpgkey \
             | sudo gpg --dearmor -o /usr/share/keyrings/fdio-archive-keyring.gpg
+          expected="59A8E0DB0B611CA471C9FBDE9E9FE7A46C37F7CD"
+          actual=$(gpg --no-default-keyring \
+            --keyring /usr/share/keyrings/fdio-archive-keyring.gpg \
+            --list-keys --with-colons 2>/dev/null \
+            | awk -F: '/^fpr/{print $10; exit}')
+          [[ "$actual" == "$expected" ]] || { echo "GPG key mismatch: got $actual"; exit 1; }
           echo "deb [signed-by=/usr/share/keyrings/fdio-archive-keyring.gpg] \
-            https://packagecloud.io/fdio/release/ubuntu $(lsb_release -cs) main" \
+            https://packagecloud.io/fdio/release/ubuntu \
+            $(. /etc/os-release && echo "$VERSION_CODENAME") main" \
             | sudo tee /etc/apt/sources.list.d/fdio.list
 
       - name: Install build deps
@@ -68,7 +75,7 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends \
               cmake ninja-build g++ ccache pkg-config \
-              vpp-dev libvppinfra-dev
+              vpp-dev=24.06* libvppinfra-dev=24.06*
 
       - name: Detect backend source
         id: detect_cpp
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,12 +49,26 @@ jobs:
   cpp-build:
     name: build-cpp
     runs-on: ubuntu-24.04
-    container:
-      image: ligato/vpp-base:24.06
-      options: --user root
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+
+      - name: Add fd.io apt repo
+        run: |
+          set -eux
+          curl -fsSL https://packagecloud.io/fdio/release/gpgkey \
+            | sudo gpg --dearmor -o /usr/share/keyrings/fdio-archive-keyring.gpg
+          echo "deb [signed-by=/usr/share/keyrings/fdio-archive-keyring.gpg] \
+            https://packagecloud.io/fdio/release/ubuntu $(lsb_release -cs) main" \
+            | sudo tee /etc/apt/sources.list.d/fdio.list
+
+      - name: Install build deps
+        run: |
+          set -eux
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+              cmake ninja-build g++ ccache pkg-config \
+              vpp-dev libvppinfra-dev
 
       - name: Detect backend source
         id: detect_cpp
@@ -69,10 +83,8 @@ jobs:
       - name: Configure + build
         if: steps.detect_cpp.outputs.have_backend == 'true'
         run: |
-          cmake_launchers=""
-          if command -v ccache >/dev/null 2>&1; then
-            cmake_launchers="-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache"
-          fi
           cmake -S src/backend -B build -G Ninja \
-              -DCMAKE_BUILD_TYPE=Release $cmake_launchers
+              -DCMAKE_BUILD_TYPE=Release \
+              -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+              -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
           cmake --build build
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -49,13 +49,27 @@ jobs:
...
+          echo "deb [signed-by=/usr/share/keyrings/fdio-archive-keyring.gpg] \
+            https://packagecloud.io/fdio/release/ubuntu $(lsb_release -cs) main" \
+            | sudo tee /etc/apt/sources.list.d/fdio.list
+
+      - name: Install build deps
+        run: |
+          set -eux
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+              cmake ninja-build g++ ccache pkg-config \
```
banidoru, line 77: Nit: `vpp-dev` and `libvppinfra-dev` are unpinned. If fd.io publishes a new major VPP release (e.g. 25.02), this job could start pulling incompatible headers and break without any code change on your side. Consider pinning to a version glob:
```yaml
vpp-dev=24.06* libvppinfra-dev=24.06*
```

.github/workflows/cpp-test.yml
```diff
@@ -17,28 +17,26 @@ jobs:
...
 
-      - name: Install build deps
+      - name: Add fd.io apt repo
         run: |
           set -eux
-          apt-get update
-          apt-get install -y --no-install-recommends \
-              cmake ninja-build g++ clang libgtest-dev ccache pkg-config
+          curl -fsSL https://packagecloud.io/fdio/release/gpgkey \
+            | sudo gpg --dearmor -o /usr/share/keyrings/fdio-archive-keyring.gpg
```
banidoru, line 28: This fd.io apt-repo block is duplicated verbatim in `build.yml`. Worth extracting into a local composite action (e.g. `.github/actions/setup-vpp/action.yml`) to keep the two workflows in sync and avoid copy-paste drift.

banidoru (commented):
- Clean removal of the container dependency; apt-based install is simpler and avoids Docker Hub rate-limits.
- `vpp-dev` / `libvppinfra-dev` are unpinned — a breaking fd.io release could silently break CI. Consider pinning to a major version range (e.g. `vpp-dev=24.06*`).
- fd.io repo setup block is duplicated verbatim across both workflows — extracting a composite action would reduce drift.
- ccache launcher is now unconditional, which is fine since it's explicitly installed in the same step.

.github/workflows/build.yml
```diff
@@ -49,13 +49,27 @@ jobs:
...
-      options: --user root
     steps:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Add fd.io apt repo
+        run: |
+          set -eux
+          curl -fsSL https://packagecloud.io/fdio/release/gpgkey \
+            | sudo gpg --dearmor -o /usr/share/keyrings/fdio-archive-keyring.gpg
```
banidoru, line 60: The GPG key is imported without verifying its fingerprint. If the packagecloud CDN is ever compromised, this pipeline would silently trust arbitrary packages. Consider adding a fingerprint check after `--dearmor`:
```bash
expected="<FINGERPRINT>"
actual=$(gpg --no-default-keyring --keyring /usr/share/keyrings/fdio-archive-keyring.gpg --list-keys --with-colons 2>/dev/null | awk -F: '/^fpr/{print $10; exit}')
[[ "$actual" == "$expected" ]] || { echo "GPG key mismatch"; exit 1; }
```
Low urgency, but good supply-chain hygiene for CI.

.github/workflows/build.yml
```diff
@@ -49,13 +49,27 @@ jobs:
...
+          echo "deb [signed-by=/usr/share/keyrings/fdio-archive-keyring.gpg] \
+            https://packagecloud.io/fdio/release/ubuntu $(lsb_release -cs) main" \
```
banidoru: `lsb_release` is available on GitHub-hosted Ubuntu runners today, but it's been deprecated upstream in favor of `/etc/os-release`. A safer alternative:
```bash
$(. /etc/os-release && echo "$VERSION_CODENAME")
```
This avoids a potential breakage if GitHub drops `lsb-release` from their runner images.

banidoru (commented):
- Clean, well-motivated migration from container to native runner — simpler and avoids Docker Hub rate limits.
- GPG key imported without fingerprint verification; low risk but worth noting for supply chain hygiene.
- Unconditional ccache launcher flags are fine since ccache is now explicitly installed — good simplification.
- Consider adding `lsb-release` to the install list or using `$(. /etc/os-release && echo $VERSION_CODENAME)` as a more portable fallback.

.github/workflows/cpp-test.yml
```diff
@@ -17,28 +17,26 @@ jobs:
...
         run: |
-          if [ ! -d /usr/include/vpp ]; then
-            echo "::error::VPP dev headers missing under /usr/include/vpp."
-            exit 1
-          fi
+          set -eux
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+              cmake ninja-build g++ clang libgtest-dev ccache pkg-config \
+              cppcheck vpp-dev libvppinfra-dev
```
banidoru, line 47: Same version-pinning concern as raised on `build.yml`: `vpp-dev` and `libvppinfra-dev` should be pinned (e.g. `=24.06*`) to avoid surprise breakage when fd.io publishes a new major release.

banidoru (commented):
- Good simplification — dropping the container removes Docker Hub rate-limit headaches and the GHCR mirror maintenance burden.
- Version-pin `vpp-dev` / `libvppinfra-dev` in **both** workflow files (existing comment covers `build.yml`; `cpp-test.yml` has the same exposure).
- Consider a composite action for the shared fd.io repo setup to prevent copy-paste drift (echoing prior feedback).
- Overall looks correct and clean. No blocking issues found.

banidoru (approved):
All reviewers approved. LGTM.

.github/workflows/build.yml
r12f: Good call — pinned both `vpp-dev` and `libvppinfra-dev` to `=24.06*` in build.yml (and cpp-test.yml). This way a new fd.io major release won't silently break us.

.github/workflows/build.yml
r12f: Added GPG fingerprint verification after dearmor. The key is checked against `59A8E0DB0B611CA471C9FBDE9E9FE7A46C37F7CD` — if it doesn't match, the step fails immediately. Applied to both workflow files.

.github/workflows/build.yml
r12f: Replaced `lsb_release -cs` with `$(. /etc/os-release && echo "$VERSION_CODENAME")` in both workflows. More portable and won't break if GitHub drops lsb-release.

.github/workflows/cpp-test.yml
r12f: Pinned to `=24.06*` here as well, same as build.yml.

.github/workflows/cpp-test.yml
r12f: Agreed this would reduce duplication. For now both files are updated in lockstep (same GPG check, same pinning, same os-release usage). Extracting a composite action is a good follow-up once we have more shared CI steps — keeping it out of scope for this PR to avoid expanding the diff.

Issue #37, opened 2026-04-19
[W3] Parser test suite - golden PCAPs, gtest, fuzzer


r12f:
## Spec
[003-erspan-and-packet-parsing](https://github.com/r12f/InFMon/blob/main/specs/003-erspan-and-packet-parsing.md)

## Scope
- 11 golden PCAP test vectors per spec 003
- gtest unit tests for each parsing path
- libFuzzer harness for the parser entry point
- CI integration (run in cpp-test workflow)

## Dependencies
Blocked by: #36

PR #68, opened 2026-04-19
feat: implement ERSPAN III / GRE parser (W2)

### LINKED ISSUE #36: [W2] ERSPAN III / GRE parser (C, VPP node)
## Spec
[003-erspan-and-packet-parsing](https://github.com/r12f/InFMon/blob/main/specs/003-erspan-and-packet-parsing.md)

## Scope
- ERSPAN III over GRE parser as a VPP graph node
- Outer header skip: Ethernet -> optional single VLAN -> IPv4/IPv6 -> GRE (0x22EB) -> ERSPAN III header
- mirror_src_ip extraction from outer L3 source address
- Inner packet view: inner_ptr, inner_len, inner_truncated, mirror_src_ip
- Truncation handling for ~128B snap (inner L2 required, inner L3 required, inner L4 best-effort with valid_fields bitmask)
- Inner-decap hook seam (v1 = NONE)
- 14 error/drop counters

## Dependencies
Blocked by: #35

r12f:
## Summary

Implement the ERSPAN III over GRE parser as specified in [specs/003-erspan-and-packet-parsing.md](specs/003-erspan-and-packet-parsing.md).

### What's included

- **Parser library** (`libinfmon_parser.a`): Pure-function C parser with no allocations, no syscalls, no cross-packet state
- **Outer header skip**: Ethernet → optional single VLAN → IPv4/IPv6 → GRE (0x22EB) → ERSPAN III header (+ optional 8B Platform Sub-Header when O=1)
- **mirror_src_ip extraction**: Outer L3 source address extracted as tagged union (v4/v6) — the only outer-header value crossing the parser boundary
- **Inner packet view**: `inner_ptr` (zero-copy pointer into original buffer), `inner_len`, `inner_truncated`
- **Truncation handling** (~128B snap): Inner L2 required, inner L3 required, inner L4 best-effort with `valid_fields` bitmask (`PORTS_VALID`, `TCP_FLAGS_VALID`, `TCP_SEQ_VALID`, `TCP_ACK_VALID`, `TCP_WINDOW_VALID`)
- **Inner-decap hook seam** (§4.6): v1 = `INFMON_DECAP_NONE` (identity), reserved for future VXLAN/GENEVE/RoCEv2
- **14 error/drop counters** per spec §6 (all indexed by `infmon_parse_result_t`)
- **CMake build** producing static library + gtest binary
- **24 gtest unit tests** covering all parse paths, error conditions, field extraction, and counter names

### Spec compliance

All requirements from spec 003 §1–§6 are implemented. Test plan items from §8 (golden PCAPs, fuzzer) are tracked separately in #37.

Closes #36

### DIFF
--- a/src/backend/include/infmon/erspan_parser.h
+++ b/src/backend/include/infmon/erspan_parser.h
@@ -22,13 +22,13 @@ extern "C" {
 /* ── Error / counter reason codes ─────────────────────────────────── */
 
 typedef enum {
-    INFMON_PARSE_OK = 0,              /* full inner packet present          */
-    INFMON_PARSE_INNER_TRUNCATED_OK,  /* inner truncated but usable         */
+    INFMON_PARSE_OK = 0,             /* full inner packet present          */
+    INFMON_PARSE_INNER_TRUNCATED_OK, /* inner truncated but usable         */
     INFMON_PARSE_ERR_OUTER_QINQ_UNSUPPORTED,
     INFMON_PARSE_ERR_OUTER_ETHERTYPE_UNSUPPORTED,
     INFMON_PARSE_ERR_OUTER_V6_EXT_UNSUPPORTED,
     INFMON_PARSE_ERR_OUTER_TRUNCATED,
-    INFMON_PARSE_ERR_MBUF_NOT_CONTIGUOUS,  /* reserved for VPP integration */
+    INFMON_PARSE_ERR_MBUF_NOT_CONTIGUOUS, /* reserved for VPP integration */
     INFMON_PARSE_ERR_GRE_UNEXPECTED_FLAGS,
     INFMON_PARSE_ERR_GRE_BAD_VERSION,
     INFMON_PARSE_ERR_GRE_BAD_PROTO,
@@ -37,7 +37,7 @@ typedef enum {
     INFMON_PARSE_ERR_INNER_ETH_TRUNCATED,
     INFMON_PARSE_ERR_INNER_L3_TRUNCATED,
     INFMON_PARSE_ERR_INNER_DOUBLE_ENCAP_DROPPED,
-    INFMON_PARSE_ERR__COUNT  /* sentinel */
+    INFMON_PARSE_ERR__COUNT /* sentinel */
 } infmon_parse_result_t;
 
 /* ── Counter names (indexed by infmon_parse_result_t) ─────────────── */
@@ -63,44 +63,44 @@ typedef struct {
 
 /* ── Valid-fields bitmask for inner L4 ────────────────────────────── */
 
-#define INFMON_VALID_PORTS      (1u << 0)
-#define INFMON_VALID_TCP_FLAGS  (1u << 1)
-#define INFMON_VALID_TCP_SEQ    (1u << 2)
-#define INFMON_VALID_TCP_ACK    (1u << 3)
+#define INFMON_VALID_PORTS (1u << 0)
+#define INFMON_VALID_TCP_FLAGS (1u << 1)
+#define INFMON_VALID_TCP_SEQ (1u << 2)
+#define INFMON_VALID_TCP_ACK (1u << 3)
 #define INFMON_VALID_TCP_WINDOW (1u << 4)
 
 /* ── Inner-decap hook types (§4.6) ────────────────────────────────── */
 
 typedef enum {
     INFMON_DECAP_NONE = 0,
-    INFMON_DECAP_VXLAN,   /* future */
-    INFMON_DECAP_GENEVE,  /* future */
-    INFMON_DECAP_ROCEV2,  /* future */
+    INFMON_DECAP_VXLAN,  /* future */
+    INFMON_DECAP_GENEVE, /* future */
+    INFMON_DECAP_ROCEV2, /* future */
 } infmon_inner_decap_t;
 
 /* ── Parser output ────────────────────────────────────────────────── */
 
 typedef struct {
-    const uint8_t *inner_ptr;       /* pointer into original buffer    */
-    uint32_t       inner_len;       /* bytes available                 */
-    bool           inner_truncated; /* inner_len < declared_inner_len  */
+    const uint8_t *inner_ptr; /* pointer into original buffer    */
+    uint32_t inner_len;       /* bytes available                 */
+    bool inner_truncated;     /* inner_len < declared_inner_len  */
 
     infmon_mirror_src_ip_t mirror_src_ip;
 
     /* Inner L4 extracted fields (valid only when corresponding bit set) */
-    uint32_t       valid_fields;    /* bitmask of INFMON_VALID_* */
-    uint16_t       src_port;        /* network byte order */
-    uint16_t       dst_port;        /* network byte order */
-    uint8_t        tcp_flags;
-    uint32_t       tcp_seq;         /* network byte order */
-    uint32_t       tcp_ack;         /* network byte order */
-    uint16_t       tcp_window;      /* network byte order */
+    uint32_t valid_fields; /* bitmask of INFMON_VALID_* */
+    uint16_t src_port;     /* network byte order */
+    uint16_t dst_port;     /* network byte order */
+    uint8_t tcp_flags;
+    uint32_t tcp_seq;    /* network byte order */
+    uint32_t tcp_ack;    /* network byte order */
+    uint16_t tcp_window; /* network byte order */
 
-    bool           flow_key_partial; /* ports missing for TCP/UDP */
+    bool flow_key_partial; /* ports missing for TCP/UDP */
 
     /* Inner L3 info */
-    uint8_t        inner_ip_proto;
-    infmon_af_t    inner_af;
+    uint8_t inner_ip_proto;
+    infmon_af_t inner_af;
 } infmon_parsed_packet_t;
 
 /* ── Main parser entry point ──────────────────────────────────────── */
@@ -115,18 +115,16 @@ typedef struct {
  * @return INFMON_PARSE_OK or INFMON_PARSE_INNER_TRUNCATED_OK on success;
  *         an error code otherwise.
  */
-infmon_parse_result_t
-infmon_parse_erspan(const uint8_t *data, uint32_t len,
-                    infmon_parsed_packet_t *out);
+infmon_parse_result_t infmon_parse_erspan(const uint8_t *data, uint32_t len,
+                                          infmon_parsed_packet_t *out);
 
 /* ── Inner-decap hook (§4.6) ──────────────────────────────────────── */
 
 /**
  * v1: only INFMON_DECAP_NONE is implemented (identity).
  */
-int infmon_inner_decap(infmon_inner_decap_t kind,
-                       const uint8_t *in, uint32_t in_len, bool in_truncated,
-                       const uint8_t **out_ptr, uint32_t *out_len,
+int infmon_inner_decap(infmon_inner_decap_t kind, const uint8_t *in, uint32_t in_len,
+                       bool in_truncated, const uint8_t **out_ptr, uint32_t *out_len,
                        bool *out_truncated);
 
 #ifdef __cplusplus
--- a/src/backend/include/infmon/erspan_parser.h
+++ b/src/backend/include/infmon/erspan_parser.h
@@ -1,0 +1,136 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright 2026 Riff
+ *
+ * ERSPAN III / GRE parser — see specs/003-erspan-and-packet-parsing.md
+ *
+ * Pure-function parser: no allocations, no syscalls, no cross-packet state.
+ * Operates on a single contiguous buffer.
+ */
+
+#ifndef INFMON_ERSPAN_PARSER_H
+#define INFMON_ERSPAN_PARSER_H
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* ── Error / counter reason codes ─────────────────────────────────── */
+
+typedef enum {
+    INFMON_PARSE_OK = 0,              /* full inner packet present          */
+    INFMON_PARSE_INNER_TRUNCATED_OK,  /* inner truncated but usable         */
+    INFMON_PARSE_ERR_OUTER_QINQ_UNSUPPORTED,
+    INFMON_PARSE_ERR_OUTER_ETHERTYPE_UNSUPPORTED,
+    INFMON_PARSE_ERR_OUTER_V6_EXT_UNSUPPORTED,
+    INFMON_PARSE_ERR_OUTER_TRUNCATED,
+    INFMON_PARSE_ERR_MBUF_NOT_CONTIGUOUS,  /* reserved for VPP integration */
+    INFMON_PARSE_ERR_GRE_UNEXPECTED_FLAGS,
+    INFMON_PARSE_ERR_GRE_BAD_VERSION,
+    INFMON_PARSE_ERR_GRE_BAD_PROTO,
+    INFMON_PARSE_ERR_ERSPAN_BAD_VERSION,
+    INFMON_PARSE_ERR_ERSPAN_TRUNCATED,
+    INFMON_PARSE_ERR_INNER_ETH_TRUNCATED,
+    INFMON_PARSE_ERR_INNER_L3_TRUNCATED,
+    INFMON_PARSE_ERR_INNER_DOUBLE_ENCAP_DROPPED,
+    INFMON_PARSE_ERR__COUNT  /* sentinel */
+} infmon_parse_result_t;
+
+/* ── Counter names (indexed by infmon_parse_result_t) ─────────────── */
+
+extern const char *infmon_parse_counter_names[];
+
+/* ── IP address family ────────────────────────────────────────────── */
+
+typedef enum {
+    INFMON_AF_V4 = 0,
+    INFMON_AF_V6,
+} infmon_af_t;
+
+/* ── Mirror source IP (tagged union) ──────────────────────────────── */
+
+typedef struct {
+    infmon_af_t family;
+    union {
+        uint8_t v4[4];
+        uint8_t v6[16];
+    } addr;
+} infmon_mirror_src_ip_t;
+
+/* ── Valid-fields bitmask for inner L4 ────────────────────────────── */
+
+#define INFMON_VALID_PORTS      (1u << 0)
+#define INFMON_VALID_TCP_FLAGS  (1u << 1)
+#define INFMON_VALID_TCP_SEQ    (1u << 2)
+#define INFMON_VALID_TCP_ACK    (1u << 3)
+#define INFMON_VALID_TCP_WINDOW (1u << 4)
+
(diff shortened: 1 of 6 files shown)


src/backend/include/infmon/erspan_parser.h
```diff
@@ -0,0 +1,134 @@
...
+    const uint8_t *inner_ptr; /* pointer into original buffer    */
+    uint32_t inner_len;       /* bytes available                 */
+    bool inner_truncated;     /* inner_len < declared_inner_len  */
+
+    infmon_mirror_src_ip_t mirror_src_ip;
+
+    /* Inner L4 extracted fields (valid only when corresponding bit set) */
+    uint32_t valid_fields; /* bitmask of INFMON_VALID_* */
+    uint16_t src_port;     /* network byte order */
+    uint16_t dst_port;     /* network byte order */
```
banidoru: These comments say "network byte order" but `read_u16`/`read_u32` actually convert *from* network byte order *to* host byte order (big-endian decode on any platform). The test assertions (`EXPECT_EQ(out.src_port, 12345)`) confirm host order. Same for `tcp_seq`, `tcp_ack`, `tcp_window` below. Please fix the comments to say "host byte order" to avoid confusion for future integrators.

src/backend/src/erspan_parser.c
```diff
@@ -0,0 +1,376 @@
...
+        /* QinQ check: if after stripping one VLAN we see another 0x8100
+         * or 0x88A8, that's stacked VLAN */
+        if (ethertype == 0x8100 || ethertype == 0x88A8)
+            return INFMON_PARSE_ERR_OUTER_QINQ_UNSUPPORTED;
+    } else if (ethertype == 0x88A8) {
+        /* S-tag (QinQ outer tag) */
+        return INFMON_PARSE_ERR_OUTER_QINQ_UNSUPPORTED;
+    }
+
+    if (ethertype != 0x0800 && ethertype != 0x86DD)
```
banidoru, line 138: `(void) 0; /* l3_off removed — not needed */` — dead code leftover from refactoring. Remove this line.

src/backend/src/erspan_parser.c
```diff
@@ -0,0 +1,376 @@
...
+                out->valid_fields |= INFMON_VALID_PORTS;
+            } else {
+                out->flow_key_partial = true;
+            }
+
+            /* TCP-specific fields */
+            if (out->inner_ip_proto == 6 && (out->valid_fields & INFMON_VALID_PORTS)) {
+                if (inner_len >= inner_l4_off + 8) {
+                    out->tcp_seq = read_u32(inner_ptr + inner_l4_off + 4);
+                    out->valid_fields |= INFMON_VALID_TCP_SEQ;
```
banidoru: Inner IPv6 path reads `inner_ip_proto` from fixed-header `next_hdr` (offset 6) and assumes L4 starts at byte 40. If the inner packet has extension headers (hop-by-hop, destination options, etc.), `inner_ip_proto` will be the extension type, not the actual L4 protocol, and port extraction will read the wrong bytes.

For v1 with ~128B snaps this may be acceptable, but it should be explicitly documented as a known limitation (e.g., a comment here and in the header). Otherwise callers may trust `inner_ip_proto`/ports for ext-header packets.

src/backend/src/erspan_parser.c
```diff
@@ -0,0 +1,376 @@
...
+            return INFMON_PARSE_ERR_INNER_L3_TRUNCATED;
+        inner_ethertype = read_u16(inner_ptr + inner_l3_off + 2);
+        inner_l3_off += 4;
+    }
+
+    if (inner_ethertype == 0x0800) {
+        /* Inner IPv4 */
+        if (inner_len < inner_l3_off + 20)
+            return INFMON_PARSE_ERR_INNER_L3_TRUNCATED;
+
```
banidoru, line 332: The TCP/UDP L4 field extraction logic (ports, seq, ack, flags, window) is duplicated nearly identically between the IPv4 path (lines ~275-295) and IPv6 path (lines ~310-330). Consider extracting a `static` helper like `extract_inner_l4(inner_ptr, inner_l4_off, inner_len, inner_ip_proto, out)` to reduce the copy-paste maintenance burden.

banidoru (commented):
- **Byte-order comments are wrong**: header says `src_port`, `dst_port`, `tcp_seq`, `tcp_ack`, `tcp_window` are "network byte order" but `read_u16`/`read_u32` actually decode to host order. Tests confirm host order. Fix the comments.
- **Inner IPv6 extension headers not skipped**: inner L4 offset assumes L4 starts right after the 40-byte fixed header. If inner IPv6 has extension headers, port/flag extraction reads garbage. At minimum document this limitation.
- **Duplicated L4 extraction**: TCP/UDP field extraction is copy-pasted between IPv4 and IPv6 inner paths (~30 lines each). Factor into a helper to reduce maintenance risk.
- **Dead code leftover**: `(void) 0; /* l3_off removed */` is a refactoring artifact — remove it.
- **Missing test for inner IPv6 + TCP**: no test exercises the inner-IPv6 L4 extraction path.

src/backend/src/erspan_parser.c
```diff
@@ -0,0 +1,376 @@
...
+    [INFMON_PARSE_ERR_OUTER_ETHERTYPE_UNSUPPORTED] = "outer_ethertype_unsupported",
+    [INFMON_PARSE_ERR_OUTER_V6_EXT_UNSUPPORTED] = "outer_v6_ext_unsupported",
+    [INFMON_PARSE_ERR_OUTER_TRUNCATED] = "outer_truncated",
+    [INFMON_PARSE_ERR_MBUF_NOT_CONTIGUOUS] = "mbuf_not_contiguous",
+    [INFMON_PARSE_ERR_GRE_UNEXPECTED_FLAGS] = "gre_unexpected_flags",
+    [INFMON_PARSE_ERR_GRE_BAD_VERSION] = "gre_bad_version",
+    [INFMON_PARSE_ERR_GRE_BAD_PROTO] = "gre_bad_proto",
+    [INFMON_PARSE_ERR_ERSPAN_BAD_VERSION] = "erspan_bad_version",
+    [INFMON_PARSE_ERR_ERSPAN_TRUNCATED] = "erspan_truncated",
+    [INFMON_PARSE_ERR_INNER_ETH_TRUNCATED] = "inner_eth_truncated",
```
banidoru, line 28: This array has no compile-time size guarantee. If a new enum value is added to `infmon_parse_result_t` without a corresponding entry here, the array will be short and indexed access will read NULL (or worse, OOB on some compilers).

Add after the array:
```c
_Static_assert(sizeof(infmon_parse_counter_names) / sizeof(infmon_parse_counter_names[0]) == INFMON_PARSE_ERR__COUNT,
               "counter_names out of sync with infmon_parse_result_t");
```

src/backend/src/erspan_parser.c
```diff
@@ -0,0 +1,376 @@
...
+        while (next_hdr == 0 || next_hdr == 60) {
+            if (len < off + 2)
+                return INFMON_PARSE_ERR_OUTER_TRUNCATED;
+            uint8_t ext_next = data[off];
+            uint8_t ext_hdr_len_field = data[off + 1];
+            uint32_t ext_len = ((uint32_t) ext_hdr_len_field + 1) * 8;
+            if (len < off + ext_len)
+                return INFMON_PARSE_ERR_OUTER_TRUNCATED;
+            v6_ext_len += ext_len;
+            next_hdr = ext_next;
```
banidoru, line 212: `outer_ip_payload_len` is `uint16_t` but `v6_ext_len` is `uint32_t`. The cast `(uint16_t) v6_ext_len` on line 159 silently truncates if total extension headers exceed 65535 bytes. While extremely unlikely in practice, using `uint32_t` for `outer_ip_payload_len` costs nothing and eliminates the hazard. Same issue with the IPv4 path where `total_length - ip_hdr_len` is cast to `uint16_t` (though that one can't overflow since IPv4 total_length is 16-bit).

src/backend/src/erspan_parser.c
```diff
@@ -0,0 +1,376 @@
...
+    truncated = decap_truncated;
+
+    /* ── Inner Ethernet (14 B required) ─────────────────────────── */
+    if (inner_len < 14)
+        return INFMON_PARSE_ERR_INNER_ETH_TRUNCATED;
+
+    /* ── Inner L3 ───────────────────────────────────────────────── */
+    uint16_t inner_ethertype = read_u16(inner_ptr + 12);
+    uint32_t inner_l3_off = 14;
+
```
banidoru, line 319: When inner IHL < 5, the code clamps `inner_ip_hdr_len` to 20 rather than flagging the packet. This means a malformed inner packet with IHL=0 will have L4 extraction attempted at the standard offset, which may produce misleading port/flag values.

Consider setting `out->flow_key_partial = true` and skipping L4 extraction when the raw IHL is < 5, since the IP header is clearly malformed.

src/backend/test/test_erspan_parser.cc
```diff
@@ -0,0 +1,653 @@
+// SPDX-License-Identifier: MIT
```
banidoru: License header says `SPDX-License-Identifier: MIT` but the parser source and header use `Apache-2.0`. Likely a copy-paste oversight — should these be aligned?

banidoru (commented):
Solid parser implementation — clean layering, thorough tests, good error taxonomy.

- **Counter array not compile-time validated**: `infmon_parse_counter_names` has no size guard — adding an enum value without a name entry silently produces a NULL. Add a `_Static_assert` after the array.
- **`outer_ip_payload_len` truncation risk on IPv6 path**: `(uint16_t) v6_ext_len` cast silently truncates if extensions exceed 64KB. Use `uint32_t` for `outer_ip_payload_len` throughout.
- **Inner IPv4 IHL < 5 silently clamped to 20**: malformed inner packets with IHL=0 get L4 extraction at the wrong offset instead of being flagged. Consider setting `flow_key_partial = true` when IHL is invalid.
- **License header mismatch**: test file uses `SPDX-License-Identifier: MIT` while parser uses `Apache-2.0`. Likely unintentional — align them.

banidoru (commented):
- `outer_ip_payload_len=0` when IPv4 `total_length` is malformed (< `ip_hdr_len`) causes `declared_inner_len=0`, bypassing the `declared < actual` bound — parser treats trailing padding/garbage as inner content
- No test for non-IP inner ethertype (ARP/unknown) — the `flow_key_partial` fallback path is untested
- IPv6 ext-header loop has no iteration cap — a crafted chain of Hop-by-Hop/DestOpts headers could spin for a long time before hitting a length check failure
- `infmon_inner_decap` return type is `int` in the `.c` but the header declares the same — consider using a proper enum/result type for consistency with the rest of the API

src/backend/src/erspan_parser.c
```diff
@@ -0,0 +1,376 @@
...
+        /* Extract mirror_src_ip (IPv4 SA at offset 12) */
+        out->mirror_src_ip.family = INFMON_AF_V4;
+        memcpy(out->mirror_src_ip.addr.v4, data + off + 12, 4);
+
+        outer_ip_payload_len =
+            total_length > ip_hdr_len ? (uint16_t) (total_length - ip_hdr_len) : 0;
+        off += ip_hdr_len;
+
+    } else {
```
banidoru, line 178: `total_length` from the IPv4 header is trusted without sanity-checking against the buffer length. When `total_length < ip_hdr_len` (malformed), `outer_ip_payload_len` becomes 0, which makes `declared_inner_len` = 0. The truncation logic then skips the `declared < actual` bound (`declared_inner_len > 0 && ...` on line 225), so `inner_len = actual_inner_len` — meaning any trailing padding or next-packet data in the buffer is treated as inner content.

Consider validating: `if (total_length < ip_hdr_len) return INFMON_PARSE_ERR_OUTER_TRUNCATED;`

src/backend/src/erspan_parser.c
```diff
@@ -0,0 +1,376 @@
...
+        uint8_t next_hdr = data[off + 6];
+
+        /* Extract mirror_src_ip (IPv6 SA at offset 8) */
+        out->mirror_src_ip.family = INFMON_AF_V6;
+        memcpy(out->mirror_src_ip.addr.v6, data + off + 8, 16);
+
+        off += 40;
+
+        /* Parse allowed extension headers: Hop-by-Hop (0) and
+         * Destination Options (60) */
```
banidoru: The IPv6 extension header `while` loop has no iteration bound. A crafted packet with a long chain of Hop-by-Hop (0) → DestOpts (60) → Hop-by-Hop → ... headers (each pointing to the other as `next_hdr`) will loop until it eventually exceeds `len`. In practice this is bounded by buffer size, but adding a small iteration cap (e.g., 8) would be a cheap safety net against pathological inputs in a packet-processing hot path.

src/backend/src/erspan_parser.c
```diff
@@ -0,0 +1,376 @@
...
+    if (erspan_ver != 2)
+        return INFMON_PARSE_ERR_ERSPAN_BAD_VERSION;
+
+    /* O bit = bit 0 of word 2 (bit 95 of header, i.e. LSB of word 2) */
+    bool o_bit = (erspan_w2 & 0x01) != 0;
+
+    uint32_t erspan_hdr_len = 12 + (o_bit ? 8 : 0);
+    if (len < off + erspan_hdr_len)
+        return INFMON_PARSE_ERR_ERSPAN_TRUNCATED;
+    off += erspan_hdr_len;
```
banidoru, line 276: The `declared_inner_len > 0` guard means that whenever `outer_ip_payload_len` computes to 0 (malformed total_length, or exactly-sized headers with no payload), the declared length is ignored entirely and `inner_len = actual_inner_len`. This silently parses whatever bytes follow in the buffer.

Consider: if `declared_inner_len == 0` and `actual_inner_len > 0`, the packet is malformed — return an error or at least set `inner_len = 0`.

banidoru (approved):
All reviewers approved. LGTM.

src/backend/include/infmon/erspan_parser.h
r12f: Good catch — fixed. The comments now correctly say "host byte order" since read_u16/read_u32 decode from big-endian (network) to host order. Updated in a0639ac.

src/backend/src/erspan_parser.c
r12f: Removed. The `(void) 0;` dead code is gone in a0639ac — the variable it guarded was already removed, so the placeholder had no purpose.

src/backend/src/erspan_parser.c
r12f: Added a comment documenting this as a known v1 limitation. Inner IPv6 extension headers are not traversed — `inner_ip_proto` may be an extension type rather than L4 protocol in that case. For v1 with ~128B ERSPAN snaps this is acceptable. See a0639ac.

src/backend/src/erspan_parser.c
r12f: Agreed — extracted a `static extract_inner_l4()` helper that handles both TCP/UDP port extraction and TCP-specific fields. Both IPv4 and IPv6 paths now call it, eliminating the duplication. See a0639ac.

src/backend/src/erspan_parser.c
r12f: Added `_Static_assert(sizeof(infmon_parse_counter_names) / sizeof(infmon_parse_counter_names[0]) == INFMON_PARSE_ERR__COUNT, ...)` right after the array. Build will now fail if the enum and array fall out of sync. See a0639ac.

src/backend/src/erspan_parser.c
r12f: Changed `outer_ip_payload_len` from `uint16_t` to `uint32_t`. The IPv6 path no longer casts through `(uint16_t)` for the subtraction. The IPv4 path also benefits since the type is now uniform. See a0639ac.

src/backend/src/erspan_parser.c
r12f: Fixed: when inner IHL < 5, we now set `flow_key_partial = true` and skip L4 extraction entirely instead of clamping to 20 and potentially reading misleading data. See a0639ac.

src/backend/test/test_erspan_parser.cc
r12f: Fixed — changed test file license header from MIT to Apache-2.0 to match the parser source and header. Copy-paste oversight. See a0639ac.

src/backend/src/erspan_parser.c
r12f: Added validation: `if (total_length < ip_hdr_len) return INFMON_PARSE_ERR_OUTER_TRUNCATED;` before computing `outer_ip_payload_len`. This catches the malformed case early. See a0639ac.

src/backend/src/erspan_parser.c
r12f: Added `MAX_IPV6_EXT_HEADERS` cap (8 iterations). The while loop now counts iterations and returns `INFMON_PARSE_ERR_OUTER_V6_EXT_UNSUPPORTED` if exceeded. Cheap safety net against pathological chains. See a0639ac.

src/backend/src/erspan_parser.c
r12f: Fixed: when `declared_inner_len == 0` and `actual_inner_len > 0`, we now return `INFMON_PARSE_ERR_OUTER_TRUNCATED` since this indicates a malformed outer IP total_length. No more silent parsing of trailing buffer bytes. See a0639ac.

banidoru (approved):
All prior review feedback addressed. Code looks solid.

- _Static_assert guards counter_names array sync — good
- L4 extraction properly factored into `extract_inner_l4` helper
- IPv6 ext header loop capped at 8 iterations, IPv4 total_length validated
- Inner IHL < 5 now correctly sets `flow_key_partial` instead of clamping
- Test coverage is comprehensive across valid/invalid/edge cases

banidoru (approved):
Re-review (iteration 2) — all prior feedback addressed. LGTM.

- All 11 previous review threads resolved: byte-order comments fixed, `_Static_assert` added, `extract_inner_l4` helper extracted, IPv6 ext header loop bounded, `declared_inner_len == 0` edge case handled, inner IHL < 5 handled, license header fixed, dead code removed, `outer_ip_payload_len` type widened, IPv4 `total_length` validated, inner IPv6 ext limitation documented
- Parser logic is correct and defensive — every layer validates bounds before reading
- Test coverage is thorough: happy paths (TCP/UDP/ICMP, IPv4/IPv6, VLAN, GRE seq bit, O-bit subheader), error paths (truncation at every layer, bad versions/protos/flags, QinQ), and field extraction verification
- Clean separation between header/impl/tests; `_Static_assert` keeps counter names in sync

banidoru (approved):
- All 11 prior review concerns addressed in new commits — byte order comments, dead code, `_Static_assert`, L4 helper extraction, type widening, loop bounds, malformed packet handling
- `extract_inner_l4()` refactor is clean and correctly preserves all TCP field extraction logic
- Inner IPv6 ext header limitation properly documented as known v1 constraint
- `declared_inner_len == 0` edge case now correctly rejects malformed outer headers
- No new issues found — code is solid, well-structured, and thoroughly tested

banidoru (approved):
All reviewers approved. LGTM.

PR #69, opened 2026-04-19
test: add parser test suite — golden PCAPs, gtest, fuzzer

### LINKED ISSUE #37: [W3] Parser test suite - golden PCAPs, gtest, fuzzer

r12f:
## Summary

Implements #37 — Parser test suite with golden PCAPs, gtest, and libFuzzer harness.

### Changes

- **11 golden PCAP test vectors** under `tests/pcaps/erspan/` per spec §8.1:
  - `erspan3_full.pcap` — full inner packet, no truncation
  - `erspan3_with_seq.pcap` — GRE S flag set
  - `erspan3_o_bit.pcap` — Platform-Specific Sub-Header present
  - `erspan3_o_bit_truncated.pcap` — O=1 truncated inside sub-header
  - `erspan3_ipv6_full.pcap` — IPv6 outer, full inner
  - `erspan3_ipv6_trunc128.pcap` — IPv6 outer, 128B snap
  - `erspan3_trunc128.pcap` — IPv4 outer, 128B snap
  - `erspan3_trunc_outer.pcap` — outer truncation (must drop)
  - `erspan3_bad_version.pcap` — ERSPAN ver=1 (must drop)
  - `erspan3_qinq.pcap` — QinQ outer (must drop)
  - `erspan3_gre_keyed.pcap` — GRE K flag (must drop)

- **PCAP-based gtest suite** (`test_erspan_pcap.cc`) with built-in minimal PCAP reader
- **libFuzzer harness** (`fuzz_erspan_parser.cc`) with 2048B max input
- **Negative test**: session ID not exposed in output struct (§4.4)
- **CI integration**: bounded 30s fuzz run seeded from golden PCAPs
- **Generator script** (`scripts/gen_golden_pcaps.py`) for PCAP reproducibility

### Test Plan
- [x] All existing gtest cases pass
- [x] All 11 PCAP-based tests pass locally
- [ ] CI fuzz run completes without crashes

Closes #37


src/backend/CMakeLists.txt
```diff
@@ -24,3 +24,25 @@ add_executable(test_erspan_parser
 target_link_libraries(test_erspan_parser PRIVATE infmon_parser GTest::gtest GTest::gtest_main)
 target_include_directories(test_erspan_parser PRIVATE include)
 add_test(NAME erspan_parser COMMAND test_erspan_parser)
+
+# --- Golden PCAP tests ---
+add_executable(test_erspan_pcap
+    test/test_erspan_pcap.cc
+)
+target_link_libraries(test_erspan_pcap PRIVATE infmon_parser GTest::gtest GTest::gtest_main)
+target_include_directories(test_erspan_pcap PRIVATE include)
```
banidoru, line 33: Bug: `CMAKE_CURRENT_SOURCE_DIR` here is `src/backend`, so `/../tests/pcaps/erspan` resolves to `src/tests/pcaps/erspan` — not the repo-root `tests/pcaps/erspan/` where the PCAPs actually live.

Should be:
```cmake
PCAP_DIR="${CMAKE_CURRENT_SOURCE_DIR}/../../tests/pcaps/erspan"
```

src/backend/test/test_erspan_pcap.cc
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    bool swap = false;
+    if (fh.magic == 0xD4C3B2A1u)
+        swap = true;
+    else if (fh.magic != 0xA1B2C3D4u) {
+        fclose(f);
+        return false;
+    }
+
```
banidoru, line 58: The PCAP reader doesn't validate `linktype`. If someone accidentally passes a non-Ethernet pcap (e.g., raw IP, linktype=101), the parser will silently misinterpret the data. Add a check:
```cpp
uint32_t lt = swap ? __builtin_bswap32(fh.linktype) : fh.linktype;
if (lt != 1) { fclose(f); return false; } // DLT_EN10MB
```

src/backend/test/test_erspan_pcap.cc
```diff
@@ -0,0 +1,226 @@
...
+    pcap_pkt_hdr ph;
+    if (fread(&ph, sizeof(ph), 1, f) != 1) {
+        fclose(f);
+        return false;
+    }
+
```
banidoru, line 71: `__builtin_bswap32` is GCC/Clang-specific. Fine for this project's targets, but a `#include <byteswap.h>` or `#if` guard would be more portable if this ever needs MSVC.

tests/scripts/gen_golden_pcaps.py
```diff
@@ -0,0 +1,145 @@
...
+pkt4 = build_pkt(IP(src="10.0.0.1", dst="10.0.0.2", proto=47), erspan_o=True)
+raw4 = bytes(pkt4)
+# Truncate 4 bytes into the 8-byte platform sub-header
+# Outer Eth(14) + IPv4(20) + GRE(4) + ERSPAN(12) = 50, then sub-header starts
+# We need to cut inside the sub-header, so keep up to offset 50 + 4 = 54
+trunc_len = 14 + 20 + 4 + 12 + 4  # 54 bytes - mid sub-header
+save("erspan3_o_bit_truncated.pcap",
+     Ether(raw4[:trunc_len]))
+
+# 5. erspan3_ipv6_full.pcap — IPv6 outer transport, full inner packet
```
banidoru, line 85: Wrapping truncated raw bytes in `Ether()` causes Scapy to re-parse and potentially pad/alter the packet before `wrpcap` serializes it. For truncated test vectors this can silently change the byte content.

Safer approach: write the raw bytes directly using `struct`-based pcap writing, or use `Raw(load=raw4[:trunc_len])` and set the linktype explicitly.

tests/scripts/gen_golden_pcaps.py
```diff
@@ -0,0 +1,145 @@
...
+# 9. erspan3_bad_version.pcap — Ver=1 (must drop)
+save("erspan3_bad_version.pcap",
+     build_pkt(IP(src="10.0.0.1", dst="10.0.0.2", proto=47), erspan_ver=1))
+
+# 10. erspan3_qinq.pcap — outer QinQ (must drop)
+# Build manually with QinQ outer
+inner_raw = bytes(inner_tcp())
+erspan = erspan3_hdr()
+gre_hdr = struct.pack("!HH", 0, 0x22EB)
+ip_payload = gre_hdr + erspan + inner_raw
```
banidoru, line 120: The manually-constructed IP header has checksum=0. If the parser ever validates IP checksums, these vectors will be rejected. Consider computing a valid checksum (even `IP(...)` via Scapy would handle this).

banidoru (changes requested):
- **PCAP_DIR path is wrong** — resolves to `src/tests/pcaps/erspan` instead of `tests/pcaps/erspan`. Tests will fail to find golden PCAPs when using the compile definition.
- PCAP reader should validate linktype is Ethernet (1) to avoid silent misparse on non-Ethernet captures.
- Scapy `Ether()` re-parse of truncated raw bytes in the generator may silently alter packet content — prefer `Raw()` or direct byte writes for truncated vectors.
- Fuzzer and PCAP tests are well-structured overall; good coverage of edge cases.

tests/scripts/gen_golden_pcaps.py
```diff
@@ -0,0 +1,145 @@
+#!/usr/bin/env python3
```
r12f, line 1: Should this be in tests folder?

tests/scripts/gen_golden_pcaps.py
```diff
@@ -0,0 +1,145 @@
...
+# SPDX-License-Identifier: Apache-2.0
+# Generate golden PCAP test vectors for ERSPAN III parser tests.
+# See specs/003-erspan-and-packet-parsing.md §8.1
+
+import os
+import struct
+from scapy.all import (
+    Ether, IP, IPv6, GRE, Raw, TCP, wrpcap,
+    Dot1Q,
+)
```
banidoru, line 10: `GRE` and `Dot1Q` are imported but never used. `Dot1Q` is especially misleading since the QinQ test vector (line 117+) is hand-crafted with `struct.pack` rather than using Scapy's `Dot1Q`. Either use `Dot1Q` for the QinQ packet (cleaner) or drop the unused import.

.github/workflows/cpp-test.yml
```diff
@@ -88,3 +88,17 @@ jobs:
...
+          cmake -S src/backend -B build-fuzz -G Ninja \
+              -DCMAKE_BUILD_TYPE=Debug \
+              -DCMAKE_C_COMPILER=clang \
+              -DCMAKE_CXX_COMPILER=clang++ \
+              -DENABLE_FUZZER=ON
+          cmake --build build-fuzz --target fuzz_erspan_parser
+          mkdir -p fuzz-corpus
+          # Seed corpus from golden PCAPs
+          cp tests/pcaps/erspan/*.pcap fuzz-corpus/ 2>/dev/null || true
+          build-fuzz/fuzz_erspan_parser fuzz-corpus/ -max_total_time=30 -max_len=2048
```
banidoru, line 104: The fuzz step creates `build-fuzz/` and `fuzz-corpus/` directories in the repo root. If a developer runs this locally they'll end up with untracked dirs. Add these to `.gitignore` (or use a path under `build/`).

src/backend/test/test_erspan_pcap.cc
```diff
@@ -0,0 +1,226 @@
...
+// ---------------------------------------------------------------------------
+TEST_F(PcapTest, SessionIdNotExposed)
+{
+    // The infmon_parsed_packet_t struct must not contain a session_id field.
+    // This is a compile-time guarantee: if someone adds session_id, this test
+    // will need updating (and should fail review per spec §4.4).
+    // Runtime check: parse a packet and verify no session_id is accessible.
+    ASSERT_TRUE(loadPcap("erspan3_full.pcap"));
+    auto rc = infmon_parse_erspan(pkt.data(), pkt.size(), &out);
+    EXPECT_EQ(rc, INFMON_PARSE_OK);
```
banidoru, line 243: This "test" is purely a comment-based contract — it compiles and runs but asserts nothing about the struct layout. A `static_assert` checking `sizeof(infmon_parsed_packet_t)` or using SFINAE to verify `session_id` is *not* a member would provide an actual compile-time guarantee. As-is, someone could add `session_id` and this test would still pass.

banidoru (commented):
- Unused imports (`GRE`, `Dot1Q`) in the PCAP generator — `Dot1Q` is imported but QinQ packet is built manually with `struct.pack`; pick one approach.
- Fuzz CI step creates `build-fuzz/` and `fuzz-corpus/` in the workspace root with no `.gitignore` — accidental commits waiting to happen.
- `SessionIdNotExposed` test is a no-op at runtime; a `static_assert` on `sizeof` or `offsetof` would be a real compile-time gate instead of a comment-only contract.
- Fuzz harness: `(uint32_t) size` narrowing cast is guarded by `size > 2048` but still triggers `-Wshorten-64-to-32` on some toolchains; use `static_cast<uint32_t>(size)` and consider `uint16_t` since max is 2048.

src/backend/test/fuzz_erspan_parser.cc
```diff
@@ -0,0 +1,21 @@
...
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    // Cap at 2048 bytes per spec recommendation
+    if (size > 2048)
+        return 0;
+
+    infmon_parsed_packet_t out;
+    infmon_parse_erspan(data, (uint32_t) size, &out);
```
banidoru: C-style cast `(uint32_t) size` will trigger `-Wold-style-cast` in strict C++ builds. Use `static_cast<uint32_t>(size)`. Also, since `max_len=2048` caps fuzzer input and the guard on line 15 rejects anything larger, this is safe — but the cast style should match the C++ codebase conventions.

.github/workflows/cpp-test.yml
```diff
@@ -88,3 +88,17 @@ jobs:
       - name: ctest
         if: steps.detect.outputs.have_backend == 'true'
         run: ctest --test-dir build --output-on-failure
+
+      - name: Fuzz (bounded, 30s)
+        if: steps.detect.outputs.have_backend == 'true'
+        run: |
+          cmake -S src/backend -B build-fuzz -G Ninja \
```
banidoru, line 95: This assumes `clang`, `clang++`, and `Ninja` are available on the runner. The earlier build steps don't install them. Add an install step (`apt-get install -y clang ninja-build`) or gate the fuzz step on their availability. Otherwise CI will fail on ubuntu runners that only have gcc.

src/backend/test/fuzz_erspan_parser.cc
```diff
@@ -0,0 +1,21 @@
...
+#include "infmon/erspan_parser.h"
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+    // Cap at 2048 bytes per spec recommendation
+    if (size > 2048)
+        return 0;
+
+    infmon_parsed_packet_t out;
```
banidoru: `out` is declared but not zero-initialized. If the parser ever reads a field of `out` before writing it (e.g., to check a flag), ASan/MSan will report use-of-uninitialized-value. Add `memset(&out, 0, sizeof(out));` or use `infmon_parsed_packet_t out = {};`.

src/backend/test/test_erspan_pcap.cc
```diff
@@ -0,0 +1,226 @@
...
+        fclose(f);
+        return false;
+    }
+
+    bool swap = false;
+    if (fh.magic == 0xD4C3B2A1u)
+        swap = true;
+    else if (fh.magic != 0xA1B2C3D4u) {
+        fclose(f);
+        return false;
```
banidoru, line 56: When `swap == true`, `version_major` and `version_minor` are not byte-swapped before any (potential future) validation. More importantly, `ph.orig_len` is also never swapped — if you ever use it for truncation detection, you'll get garbage on big-endian-written pcaps.

banidoru (commented):
- CI fuzz step assumes `clang` and `Ninja` are installed but no step installs them — will fail on runners without them pre-installed
- Fuzzer harness passes uninitialized `out` struct to parser; if parser reads any field before writing, ASan will flag it
- `read_first_packet` never validates swapped `version_major`/`version_minor` — a corrupt file with valid-looking swapped magic could produce garbage `incl_len`
- Missing test coverage: no test for zero-length input (`pkt.size() == 0`) passed to `infmon_parse_erspan`

src/backend/test/test_erspan_pcap.cc
```diff
@@ -0,0 +1,226 @@
...
+// ---------------------------------------------------------------------------
+
+TEST_F(PcapTest, ErspanFull)
+{
+    ASSERT_TRUE(loadPcap("erspan3_full.pcap"));
+    auto rc = infmon_parse_erspan(pkt.data(), pkt.size(), &out);
+    EXPECT_EQ(rc, INFMON_PARSE_OK);
+    EXPECT_FALSE(out.inner_truncated);
+    EXPECT_EQ(out.mirror_src_ip.family, INFMON_AF_V4);
+    EXPECT_EQ(out.mirror_src_ip.addr.v4[0], 10);
```
banidoru, line 128: Good coverage of happy and error paths, but there's no test for the degenerate case of a zero-length buffer (`infmon_parse_erspan(nullptr, 0, &out)` or empty vector). Worth adding to confirm the parser doesn't crash on empty input — the fuzzer will find this eventually, but an explicit test is cheaper.

src/backend/CMakeLists.txt
r12f: Good catch — fixed in 1df109b. Changed to `../../tests/pcaps/erspan` to resolve correctly from `src/backend/`.

src/backend/test/test_erspan_pcap.cc
r12f: Agreed — added a linktype check (`lt != 1 → reject`) right after magic validation in 1df109b.

src/backend/test/test_erspan_pcap.cc
r12f: Valid point. This project targets Linux on amd64/arm64 only (GCC and Clang), so `__builtin_bswap32` is fine. MSVC portability is out of scope for now — we can revisit if the target matrix ever grows.

tests/scripts/gen_golden_pcaps.py
r12f: Checked — `Ether(raw4[:trunc_len])` dissects and re-serializes, but since the truncation point is past the Ethernet header and Scapy preserves the raw payload via `Raw`, the written bytes match the input. The golden PCAPs have been verified against the parser. If we ever see drift, switching to `struct`-based writing would be the fix. Leaving as-is for now.

tests/scripts/gen_golden_pcaps.py
r12f: The parser deliberately does not validate IP checksums (it's a mirror-capture parser where checksums may be offloaded/wrong). Zero checksums in test vectors are fine. If checksum validation is added in the future, we'll update the generator then.

tests/scripts/gen_golden_pcaps.py
r12f: Moved to `tests/scripts/gen_golden_pcaps.py` in 1df109b. Updated `OUTDIR` to resolve correctly from the new location.

tests/scripts/gen_golden_pcaps.py
r12f: Removed `GRE` and `Dot1Q` imports in 1df109b.

.github/workflows/cpp-test.yml
r12f: Added `fuzz-corpus/` to `.gitignore` in 1df109b. `build-fuzz/` was already covered by the existing `build-*/` pattern.

src/backend/test/test_erspan_pcap.cc
r12f: Good point — replaced the comment-based contract with a SFINAE `static_assert` that fails compilation if `session_id` is ever added to the struct. Done in 1df109b.

src/backend/test/fuzz_erspan_parser.cc
r12f: Changed to `static_cast<uint32_t>(size)` in 1df109b.

.github/workflows/cpp-test.yml
r12f: clang, clang++, and ninja-build are already installed in the `Install build deps` step (line 45). No additional install needed.

src/backend/test/fuzz_erspan_parser.cc
r12f: Good catch — changed to `infmon_parsed_packet_t out = {};` in 1549567.

src/backend/test/test_erspan_pcap.cc
r12f: Fair point. We don't use `version_major`/`version_minor`/`orig_len` in this reader (we only read `incl_len` bytes of the first packet), so swapping them would be dead code today. If we ever extend the reader to use those fields, we'll add the swaps then.

src/backend/test/test_erspan_pcap.cc
r12f: Good idea — added `ZeroLengthInput` test (`nullptr, 0`) in 1549567.

banidoru (approved):
- All 14 prior review concerns addressed and verified in new commits
- PCAP path fix, linktype validation, zero-init, static_cast, SFINAE static_assert, zero-length test — all confirmed in diff
- Test coverage is thorough: 11 golden PCAPs, fuzzer harness, negative/edge cases

banidoru (approved):
- All 14 prior review concerns addressed: PCAP_DIR path fix, linktype validation, unused imports removed, SFINAE static_assert, zero-init, static_cast, ZeroLengthInput test added
- New diff is clean — no new correctness, security, or design issues found
- Good test coverage: 11 golden PCAPs covering happy paths, truncation, bad versions, QinQ, GRE flags
- Fuzzer harness is solid with proper bounds checking and sanitizer flags
- CI integration is well-gated behind `have_backend` detection

banidoru (approved):
- All iteration-1 feedback addressed: PCAP_DIR path fix, linktype validation, static_cast, zero-init, SFINAE static_assert, unused imports removed, fuzz-corpus gitignore, ZeroLengthInput test added
- Golden PCAP test suite is comprehensive: 12 test cases covering happy paths, truncation variants, error conditions, and compile-time struct contract
- Fuzz harness is well-structured with proper ASan/UBSan integration and bounded CI execution
- PCAP reader is minimal but correct for its scope; acknowledged tradeoffs (no orig_len swap, __builtin_bswap32) are reasonable given Linux-only target
- No new issues found in iteration 2 diff — clean and ready to merge

banidoru (approved):
All reviewers approved. LGTM.

PR #71, opened 2026-04-19
feat: flow-rule data model - key layout, field encodings, validation

### LINKED ISSUE #38: [W4] Flow-rule data model - key layout, field encodings, validation
## Spec
[002-flow-tracking-model](https://github.com/r12f/InFMon/blob/main/specs/002-flow-tracking-model.md)

## Scope
- flow_rule structure: (name, ordered_field_list, max_keys, eviction_policy)
- v1 field set: src_ip, dst_ip, ip_proto, dscp, mirror_src_ip
- Key layout: concatenated canonical encodings, <=64 bytes
- 6 validation rules (all-or-nothing)
- lru_drop eviction policy implementation
- Per-flow-rule metrics: flows, evictions_total, drops_total, packets_total, bytes_total

## Dependencies
Blocked by: #36

r12f:
## Summary

Implements the flow-rule data model per [spec 002-flow-tracking-model](specs/002-flow-tracking-model.md).

### What's included

- **flow_rule structure**: `(name, ordered_field_list, max_keys, eviction_policy)`
- **v1 field set**: `src_ip`, `dst_ip`, `ip_proto`, `dscp`, `mirror_src_ip` with canonical encodings
- **Key layout**: concatenated field encodings in declared order, no padding, ≤64 bytes
- **IPv4-mapped-IPv6**: uniform 16-byte representation (`::ffff:a.b.c.d`) for all IP fields
- **6 validation rules** (all-or-nothing): name uniqueness, non-empty fields without duplicates, positive max_keys, supported eviction policy, key width ≤64B, cumulative budget ≤16Mi
- **Flow-rule set**: array-based storage with CRUD (add/rm/find/list)
- **Per-flow-rule metrics struct**: flows, evictions_total, drops_total, packets_total, bytes_total
- **Key encoding**: function to build key bytes from parsed flow fields
- **24 GTest unit tests** covering field metadata, name validation, rule validation, CRUD, key encoding, and IPv4-mapped-v6 layout

### Files

- `src/backend/include/infmon/flow_rule.h` — public API
- `src/backend/src/flow_rule.c` — implementation
- `src/backend/test/test_flow_rule.cc` — tests
- `src/backend/CMakeLists.txt` — build integration

Closes #38

### DIFF
--- a/src/backend/CMakeLists.txt
+++ b/src/backend/CMakeLists.txt
@@ -50,7 +50,6 @@ target_link_libraries(test_flow_rule PRIVATE infmon_flow_rule GTest::gtest GTest
 target_include_directories(test_flow_rule PRIVATE include)
 add_test(NAME flow_rule COMMAND test_flow_rule)
 
-# --- Fuzz target (clang only, requires -fsanitize=fuzzer) ---
 if(ENABLE_FUZZER)
     add_executable(fuzz_erspan_parser
         test/fuzz_erspan_parser.cc
--- a/src/backend/CMakeLists.txt
+++ b/src/backend/CMakeLists.txt
@@ -37,6 +37,20 @@ add_test(NAME erspan_pcap COMMAND test_erspan_pcap)
 
 # --- Fuzz target (clang only, requires -fsanitize=fuzzer) ---
 option(ENABLE_FUZZER "Build libFuzzer targets (requires clang)" OFF)
+# --- Flow-rule data model library ---
+add_library(infmon_flow_rule STATIC
+    src/flow_rule.c
+)
+target_include_directories(infmon_flow_rule PUBLIC include)
+
+add_executable(test_flow_rule
+    test/test_flow_rule.cc
+)
+target_link_libraries(test_flow_rule PRIVATE infmon_flow_rule GTest::gtest GTest::gtest_main)
+target_include_directories(test_flow_rule PRIVATE include)
+add_test(NAME flow_rule COMMAND test_flow_rule)
+
+# --- Fuzz target (clang only, requires -fsanitize=fuzzer) ---
 if(ENABLE_FUZZER)
     add_executable(fuzz_erspan_parser
         test/fuzz_erspan_parser.cc
--- a/src/backend/include/infmon/flow_rule.h
+++ b/src/backend/include/infmon/flow_rule.h
@@ -42,6 +42,7 @@ typedef enum {
     INFMON_FLOW_RULE_ERR_NOT_FOUND,
     INFMON_FLOW_RULE_ERR_INVALID_SPEC,
     INFMON_FLOW_RULE_ERR_BUDGET_EXCEEDED,
+    INFMON_FLOW_RULE_ERR_SET_FULL,
     INFMON_FLOW_RULE_ERR_INTERNAL,
 } infmon_flow_rule_result_t;
 
@@ -61,11 +62,12 @@ typedef struct {
     uint32_t field_count;
     uint32_t max_keys;
     infmon_eviction_policy_t eviction_policy;
-    uint32_t key_width; /* computed, cached */
+    uint32_t key_width; /* computed by add(); only valid after infmon_flow_rule_add() */
 } infmon_flow_rule_t;
 
 /* ── Per-flow-rule metrics ───────────────────────────────────────── */
 
+/* Scaffolding for the metrics subsystem (next PR); not yet referenced by any API. */
 typedef struct {
     uint64_t flows;           /* gauge: current key count */
     uint64_t evictions_total; /* counter */
@@ -105,6 +107,12 @@ const infmon_flow_rule_t *infmon_flow_rule_get(const infmon_flow_rule_set_t *set
 
 /* ── Validation (single rule, no set constraints) ────────────────── */
 
+/**
+ * Validate a flow rule in isolation.
+ *
+ * Name constraints: 2-31 chars, pattern ^[a-z0-9][a-z0-9_-]+$.
+ * max_keys must be in [1, INFMON_FLOW_RULE_MAX_KEYS_BUDGET].
+ */
 infmon_flow_rule_result_t infmon_flow_rule_validate(const infmon_flow_rule_t *rule);
 
 /* ── Key encoding ────────────────────────────────────────────────── */
--- a/src/backend/include/infmon/flow_rule.h
+++ b/src/backend/include/infmon/flow_rule.h
@@ -1,0 +1,134 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright 2026 Riff
+ *
+ * Flow-rule data model — see specs/002-flow-tracking-model.md
+ */
+
+#ifndef INFMON_FLOW_RULE_H
+#define INFMON_FLOW_RULE_H
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* ── Field enum ──────────────────────────────────────────────────── */
+
+typedef enum {
+    INFMON_FIELD_SRC_IP = 0,
+    INFMON_FIELD_DST_IP,
+    INFMON_FIELD_IP_PROTO,
+    INFMON_FIELD_DSCP,
+    INFMON_FIELD_MIRROR_SRC_IP,
+    INFMON_FIELD__COUNT,
+} infmon_field_t;
+
+/* ── Eviction policy ─────────────────────────────────────────────── */
+
+typedef enum {
+    INFMON_EVICTION_LRU_DROP = 0,
+    INFMON_EVICTION__COUNT,
+} infmon_eviction_policy_t;
+
+/* ── CRUD error codes ────────────────────────────────────────────── */
+
+typedef enum {
+    INFMON_FLOW_RULE_OK = 0,
+    INFMON_FLOW_RULE_ERR_NAME_EXISTS,
+    INFMON_FLOW_RULE_ERR_NOT_FOUND,
+    INFMON_FLOW_RULE_ERR_INVALID_SPEC,
+    INFMON_FLOW_RULE_ERR_BUDGET_EXCEEDED,
+    INFMON_FLOW_RULE_ERR_INTERNAL,
+} infmon_flow_rule_result_t;
+
+/* ── Constants ───────────────────────────────────────────────────── */
+
+#define INFMON_FLOW_RULE_NAME_MAX 31
+#define INFMON_FLOW_RULE_KEY_MAX  64
+#define INFMON_FLOW_RULE_FIELDS_MAX INFMON_FIELD__COUNT
+#define INFMON_FLOW_RULE_MAX_KEYS_BUDGET (16u * 1024 * 1024) /* 16 Mi */
+#define INFMON_FLOW_RULE_SET_MAX 16
+
+/* ── Flow rule (immutable after creation) ────────────────────────── */
+
+typedef struct {
+    char                    name[INFMON_FLOW_RULE_NAME_MAX + 1];
+    infmon_field_t          fields[INFMON_FLOW_RULE_FIELDS_MAX];
+    uint32_t                field_count;
+    uint32_t                max_keys;
+    infmon_eviction_policy_t eviction_policy;
+    uint32_t                key_width; /* computed, cached */
+} infmon_flow_rule_t;
+
+/* ── Per-flow-rule metrics ───────────────────────────────────────── */
+
+typedef struct {
+    uint64_t flows;           /* gauge: current key count */
+    uint64_t evictions_total; /* counter */
+    uint64_t drops_total;     /* counter */
+    uint64_t packets_total;   /* counter */
+    uint64_t bytes_total;     /* counter */
+} infmon_flow_rule_metrics_t;
+
+/* ── Normalised flow fields (input to key encoder) ───────────────── */
+
+typedef struct {
+    uint8_t src_ip[16];        /* IPv4-mapped-IPv6, network byte order */
+    uint8_t dst_ip[16];        /* IPv4-mapped-IPv6, network byte order */
+    uint8_t mirror_src_ip[16]; /* IPv4-mapped-IPv6, network byte order */
+    uint8_t ip_proto;
+    uint8_t dscp;              /* 0..63, upper 2 bits zero */
+} infmon_flow_fields_t;
+
+/* ── Flow rule set (opaque) ──────────────────────────────────────── */
+
+typedef struct infmon_flow_rule_set infmon_flow_rule_set_t;
+
+/* ── Lifecycle ───────────────────────────────────────────────────── */
+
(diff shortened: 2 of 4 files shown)


src/backend/src/flow_rule.c
```diff
@@ -0,0 +1,273 @@
...
+}
+
+const infmon_flow_rule_t *infmon_flow_rule_get(const infmon_flow_rule_set_t *set, uint32_t index)
+{
+    if (!set || index >= set->count)
+        return NULL;
+    return &set->rules[index];
+}
+
+infmon_flow_rule_result_t infmon_flow_rule_add(infmon_flow_rule_set_t *set,
```
banidoru, line 237: **Integer overflow**: `set->used_keys + rule->max_keys` is unsigned 32-bit arithmetic — if both values are large this wraps to a small number and the budget check passes incorrectly.

```c
if ((uint64_t)set->used_keys + rule->max_keys > set->max_keys_budget)
```

Or check `rule->max_keys > set->max_keys_budget - set->used_keys` (safe since `used_keys <= budget` is an invariant).

src/backend/src/flow_rule.c
```diff
@@ -0,0 +1,273 @@
...
+{
+    if (!set || !name)
+        return NULL;
+    for (uint32_t i = 0; i < set->count; i++) {
+        if (strcmp(set->rules[i].name, name) == 0)
+            return &set->rules[i];
+    }
+    return NULL;
+}
+
```
banidoru, line 229: Returning `ERR_INTERNAL` when the rule set is full is misleading — the caller can't distinguish "I hit the 16-rule cap" from an actual internal bug. Consider adding `INFMON_FLOW_RULE_ERR_SET_FULL` or reusing `ERR_BUDGET_EXCEEDED`.

src/backend/src/flow_rule.c
```diff
@@ -0,0 +1,273 @@
...
+void infmon_flow_rule_encode_key(const infmon_flow_rule_t *rule, const infmon_flow_fields_t *fields,
+                                 uint8_t *key_buf)
+{
+    uint32_t off = 0;
+    for (uint32_t i = 0; i < rule->field_count; i++) {
+        switch (rule->fields[i]) {
+        case INFMON_FIELD_SRC_IP:
+            memcpy(key_buf + off, fields->src_ip, 16);
+            off += 16;
```
banidoru, line 167: `infmon_flow_rule_encode_key` is the only public function that doesn't NULL-check its arguments. A NULL `rule`, `fields`, or `key_buf` will segfault. Even if callers are expected to validate first, the rest of the API is defensive — this should be consistent.

src/backend/CMakeLists.txt
```diff
@@ -37,6 +37,20 @@ add_test(NAME erspan_pcap COMMAND test_erspan_pcap)
...
+add_library(infmon_flow_rule STATIC
+    src/flow_rule.c
+)
+target_include_directories(infmon_flow_rule PUBLIC include)
+
+add_executable(test_flow_rule
+    test/test_flow_rule.cc
+)
+target_link_libraries(test_flow_rule PRIVATE infmon_flow_rule GTest::gtest GTest::gtest_main)
+target_include_directories(test_flow_rule PRIVATE include)
```
banidoru, line 50: This `# --- Fuzz target ---` comment is now duplicated — the original at line 39 was kept and a new identical one was inserted here. Remove one of them.

src/backend/include/infmon/flow_rule.h
```diff
@@ -0,0 +1,129 @@
...
+    uint32_t key_width; /* computed, cached */
+} infmon_flow_rule_t;
+
+/* ── Per-flow-rule metrics ───────────────────────────────────────── */
+
+typedef struct {
+    uint64_t flows;           /* gauge: current key count */
+    uint64_t evictions_total; /* counter */
+    uint64_t drops_total;     /* counter */
+    uint64_t packets_total;   /* counter */
```
banidoru, line 75: `key_width` is publicly visible but only correctly populated by `infmon_flow_rule_add()`. A caller who builds a rule, calls `validate()`, and then uses `encode_key()` will read a zero/stale `key_width`. Consider either:
- Computing it inside `validate()` and storing it, or
- Documenting that `key_width` is only valid after `add()`, or
- Making `encode_key` compute the width on the fly (it already knows the fields).

banidoru (commented):
Solid foundational data model. A few issues to consider:

- **Integer overflow in budget check** — `used_keys + max_keys` can wrap around on uint32_t, bypassing the budget guard.
- **Duplicate CMake comment** — the fuzz-target section comment is repeated after the new block.
- **Set-full returns wrong error code** — `ERR_INTERNAL` when the set is at capacity; a dedicated error or `ERR_BUDGET_EXCEEDED` would be clearer.
- **`encode_key` has no NULL guards** — unlike every other public function, it dereferences all three pointers unchecked.
- **`key_width` is a foot-gun outside `add()`** — callers using `validate()` + `encode_key()` directly will have a stale/zero `key_width`.

src/backend/src/flow_rule.c
```diff
@@ -0,0 +1,273 @@
...
+    /* Check unknown fields and duplicates */
+    bool seen[INFMON_FIELD__COUNT] = {false};
+    for (uint32_t i = 0; i < rule->field_count; i++) {
+        if ((unsigned) rule->fields[i] >= INFMON_FIELD__COUNT)
+            return INFMON_FLOW_RULE_ERR_INVALID_SPEC;
+        if (seen[rule->fields[i]])
+            return INFMON_FLOW_RULE_ERR_INVALID_SPEC;
+        seen[rule->fields[i]] = true;
+    }
+
```
banidoru, line 138: `validate()` accepts any `max_keys` > 0, but `add()` rejects values exceeding the budget. A rule with `max_keys = UINT32_MAX` silently passes validation and only fails later in `add()`. Consider validating `max_keys <= INFMON_FLOW_RULE_MAX_KEYS_BUDGET` here so callers get early feedback without needing a set.

src/backend/include/infmon/flow_rule.h
```diff
@@ -0,0 +1,129 @@
...
+    uint64_t bytes_total;     /* counter */
+} infmon_flow_rule_metrics_t;
+
+/* ── Normalised flow fields (input to key encoder) ───────────────── */
+
+typedef struct {
+    uint8_t src_ip[16];        /* IPv4-mapped-IPv6, network byte order */
```
banidoru, line 82: `infmon_flow_rule_metrics_t` is declared but not referenced by any function in this PR. If it's intentional scaffolding for a future PR, a brief comment noting that would help. Otherwise it's dead code that will rot.

banidoru (commented):
- `validate()` doesn't cap `max_keys` against `MAX_KEYS_BUDGET` — a rule with `max_keys = UINT32_MAX` passes validation but always fails `add()`. Push this check into validate for fail-fast.
- `infmon_flow_rule_metrics_t` declared but unused in this PR — confirm it's intentional scaffolding or remove to avoid dead code.
- `name_valid` enforces min-length-2 but the header has no docs on naming rules — callers will discover constraints only at runtime.

src/backend/src/flow_rule.c
```diff
@@ -0,0 +1,273 @@
...
+/* ── Name validation: ^[a-z0-9][a-z0-9_-]{1,30}$ ────────────────── */
+
+static bool name_valid(const char *name)
+{
+    if (!name)
+        return false;
+    size_t len = strlen(name);
+    if (len < 2 || len > INFMON_FLOW_RULE_NAME_MAX)
+        return false;
+
```
banidoru, line 88: The naming constraints (min 2 chars, lowercase+digits+hyphen+underscore, must start with alnum) aren't documented in the public header. Callers have to read the source or guess. Consider adding a doc comment on `infmon_flow_rule_validate()` or near `INFMON_FLOW_RULE_NAME_MAX` describing the allowed pattern.

src/backend/src/flow_rule.c
```diff
@@ -0,0 +1,273 @@
...
+    uint32_t used_keys;
+};
+
+infmon_flow_rule_set_t *infmon_flow_rule_set_create(uint32_t max_keys_budget)
+{
+    infmon_flow_rule_set_t *s = calloc(1, sizeof(*s));
+    if (!s)
+        return NULL;
+    s->max_keys_budget = max_keys_budget;
+    return s;
```
banidoru, line 206: **Dangling pointers from `find()`/`get()`**: These return `const infmon_flow_rule_t*` pointing directly into `set->rules[]`. If `infmon_flow_rule_rm()` is called afterwards, it shifts elements and invalidates any previously returned pointer. This is a classic use-after-free waiting to happen.

At minimum, document this prominently in the header ("returned pointer is invalidated by any subsequent add/rm"). Better: consider returning a copy, or an index.

src/backend/src/flow_rule.c
```diff
@@ -0,0 +1,273 @@
...
+    *dst = *rule;
+    dst->key_width = infmon_flow_rule_key_width(rule->fields, rule->field_count);
+    set->used_keys += dst->max_keys;
+    set->count++;
+    return INFMON_FLOW_RULE_OK;
+}
+
+infmon_flow_rule_result_t infmon_flow_rule_rm(infmon_flow_rule_set_t *set, const char *name)
+{
+    if (!set || !name)
```
banidoru, line 266: After shifting elements down, `set->rules[set->count]` still contains stale data (a copy of the old last element). Consider `memset(&set->rules[set->count], 0, sizeof(set->rules[0]))` to avoid confusion during debugging and prevent accidental use of stale data.

src/backend/src/flow_rule.c
```diff
@@ -0,0 +1,273 @@
...
+    infmon_flow_rule_set_t *s = calloc(1, sizeof(*s));
+    if (!s)
+        return NULL;
+    s->max_keys_budget = max_keys_budget;
+    return s;
+}
+
+void infmon_flow_rule_set_destroy(infmon_flow_rule_set_t *set)
+{
+    free(set);
```
banidoru, line 211: `max_keys_budget` is accepted as-is with no upper bound check. A caller can pass `UINT32_MAX`, which defeats the purpose of the `INFMON_FLOW_RULE_MAX_KEYS_BUDGET` constant. Consider clamping or rejecting values above the documented 16Mi limit.

banidoru (commented):
- **Dangling pointer hazard**: `find()` and `get()` return pointers into the internal array, but `rm()` shifts elements — any held pointer is silently invalidated. Needs documentation or a safer pattern.
- **`set_create` accepts any budget**: no validation against `INFMON_FLOW_RULE_MAX_KEYS_BUDGET`, so callers can create sets with budgets exceeding the documented 16Mi cap.
- **Stale slot after `rm()`**: the shifted-away last slot isn't zeroed, leaving stale data that could confuse debugging or future code that touches `rules[count]`.
- **Test coverage gap**: no test for removing a middle element and verifying remaining order is preserved, and no test for filling the set to `SET_MAX` and confirming the capacity error.

src/backend/test/test_flow_rule.cc
```diff
@@ -0,0 +1,327 @@
...
+}
+
+/* ── Helpers ─────────────────────────────────────────────────────── */
+
+static infmon_flow_rule_t make_rule(const char *name, const infmon_field_t *fields, uint32_t fc,
+                                    uint32_t max_keys = 1024)
+{
+    infmon_flow_rule_t r{};
+    std::memset(r.name, 0, sizeof(r.name));
+    std::size_t len = std::strlen(name);
```
banidoru, line 19: The `memset` on line 20 is redundant — `infmon_flow_rule_t r{}` already zero-initializes the struct in C++. Minor nit but it adds noise.

banidoru (approved):
All reviewers approved. LGTM.

src/backend/src/flow_rule.c
r12f: Fixed — cast to `uint64_t` before addition: `if ((uint64_t)set->used_keys + rule->max_keys > set->max_keys_budget)`. Good catch on the overflow.

src/backend/src/flow_rule.c
r12f: Agreed — added `INFMON_FLOW_RULE_ERR_SET_FULL` to the error enum and `add()` now returns it when the 16-rule cap is hit, keeping `ERR_INTERNAL` for actual bugs.

src/backend/src/flow_rule.c
r12f: Added NULL checks for `rule`, `fields`, and `key_buf` at the top of `encode_key()` — returns early (void) on NULL, consistent with the defensive style in the rest of the API.

src/backend/CMakeLists.txt
r12f: Removed the duplicate `# --- Fuzz target ---` comment at line 53. The original at line 38 stays.

src/backend/include/infmon/flow_rule.h
r12f: Added a doc comment on `key_width`: "computed by add(); only valid after infmon_flow_rule_add()". This makes the lifetime explicit without changing the struct layout.

src/backend/src/flow_rule.c
r12f: Good point — `validate()` now rejects `max_keys > INFMON_FLOW_RULE_MAX_KEYS_BUDGET`, so callers get early feedback without needing a set.

src/backend/include/infmon/flow_rule.h
r12f: Added a comment above the struct: "Scaffolding for the metrics subsystem (next PR); not yet referenced by any API." This is intentional forward-declaration for the metrics PR.

src/backend/src/flow_rule.c
r12f: Added a doc comment on `infmon_flow_rule_validate()` documenting the name pattern (`^[a-z0-9][a-z0-9_-]+$`, 2-31 chars) and the `max_keys` range constraint.

src/backend/src/flow_rule.c
r12f: Good point. Added a prominent doc comment in the header above the CRUD section: "Pointers returned by find()/get() point into the internal array and are invalidated by any subsequent add() or rm() call." Returning copies would change the API contract for all callers, so documenting the lifetime is the right fix for now.

src/backend/src/flow_rule.c
r12f: Fixed — added `memset(&set->rules[set->count], 0, sizeof(set->rules[0]))` after the shift+decrement to zero out the stale trailing slot.

src/backend/src/flow_rule.c
r12f: Fixed — `infmon_flow_rule_set_create()` now returns NULL if `max_keys_budget > INFMON_FLOW_RULE_MAX_KEYS_BUDGET`, enforcing the 16Mi cap at the set level too.

src/backend/test/test_flow_rule.cc
r12f: Removed the redundant `memset` — `infmon_flow_rule_t r{}` already zero-initializes in C++.

PR #72, opened 2026-04-19
feat: flow-rule YAML config schema, parsing, validation, and CRUD API

### LINKED ISSUE #39: [W5] Flow-rule YAML config schema
## Spec
[002-flow-tracking-model](https://github.com/r12f/InFMon/blob/main/specs/002-flow-tracking-model.md)

## Scope
- YAML config schema for flow-rules under /etc/infmon/config.yaml
- Config parsing and validation
- CRUD API surface: add, rm, list, show (synchronous, atomic)

## Dependencies
Blocked by: #38

r12f:
## Summary

Adds the `infmon-config` Rust crate implementing YAML configuration parsing, validation, and an in-memory CRUD API for flow-rules per spec 002.

### What's included:
- **Rust workspace** (`Cargo.toml`) with `src/infmon-config` member
- **Data model** (`model.rs`): `Field`, `EvictionPolicy`, `FlowRule`, `Config` types with serde support
- **Validation** (`validate.rs`): All §5.3 rules — name regex, field constraints, key width ≤64B, budget ≤16Mi
- **CRUD API** (`crud.rs`): `FlowRuleSet` with `add`/`rm`/`list`/`show`, budget enforcement, set-size cap (16)
- **YAML parsing** (`parse.rs`): `parse_yaml`, `parse_yaml_file`, `load_config`
- **20 unit tests** covering parsing, validation edge cases, CRUD operations, and round-trip

### Config format:
```yaml
flow-rules:
  - name: by_5tuple_l3
    fields: [mirror_src_ip, src_ip, dst_ip, ip_proto, dscp]
    max_keys: 1048576
    eviction_policy: lru_drop
```

Closes #39


src/infmon-config/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "infmon-config"
+version = "0.1.0"
+edition = "2021"
+license = "Apache-2.0"
+description = "InFMon flow-rule YAML configuration parsing and validation"
+
+[dependencies]
+serde = { version = "1", features = ["derive"] }
+serde_yaml = "0.9"
```
banidoru, line 10: `serde_yaml = "0.9"` is officially deprecated (note the `+deprecated` tag in `Cargo.lock`). Consider migrating to `serde_yml`, `yaml-rust2` + manual serde, or another actively maintained alternative before this crate gains more downstream consumers.

src/infmon-config/src/crud.rs
```diff
@@ -0,0 +1,94 @@
...
+        Self {
+            rules: Vec::new(),
+            max_keys_budget,
+        }
+    }
+
+    /// Load from Config, validating everything. All-or-nothing.
+    pub fn from_config(config: &Config) -> Result<Self, CrudError> {
+        validate_config(config)?;
+        Ok(Self {
```
banidoru, line 44: `from_config` hardcodes `MAX_KEYS_BUDGET`, but `new()` accepts a custom budget. This means `from_config` always uses the global default, making the parameterized `new()` inconsistent. Consider adding a `budget` parameter to `from_config` or using a builder pattern.

src/infmon-config/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    #[test]
+    fn reject_max_keys_zero() {
+        let rule = make_rule("ab", vec![Field::SrcIp], 0);
+        assert_eq!(
+            validate_rule(&rule),
+            Err(ValidationError::ZeroMaxKeys { name: "ab".into() })
+        );
+    }
```
banidoru, line 102: This test is named `reject_key_width_exceeded` but it actually asserts `Ok(())`. The inline comment acknowledges the fields only sum to 50 bytes (under the 64-byte limit). Either:
1. Add wider fields in the future to make this a real negative test, or
2. Rename to something like `accept_within_key_width_limit` so it doesn't mislead readers into thinking width-exceeded is tested.

src/infmon-config/src/crud.rs
```diff
@@ -0,0 +1,94 @@
...
+    BudgetExceeded {
+        used: u64,
+        requested: u32,
+        budget: u32,
+    },
+    #[error("flow-rule set is full (max {max})")]
+    SetFull { max: usize },
+}
+
+pub struct FlowRuleSet {
```
banidoru, line 28: The CRUD API has `add`/`rm`/`list`/`show` but no `update` or `replace`. Callers must `rm` then `add`, which isn't atomic — if `add` fails validation after `rm`, the old rule is lost. A dedicated `update(&mut self, rule: FlowRule) -> Result<FlowRule, CrudError>` that swaps in-place would be safer.

src/infmon-config/src/validate.rs
```diff
@@ -0,0 +1,103 @@
...
+    if !(NAME_MIN_LEN..=NAME_MAX_LEN).contains(&len) {
+        return false;
+    }
+    let bytes = name.as_bytes();
+    // First char: [a-z0-9]
+    if !bytes[0].is_ascii_lowercase() && !bytes[0].is_ascii_digit() {
+        return false;
+    }
+    // Rest: [a-z0-9_-]
+    bytes[1..]
```
banidoru, line 51: Name validation allows names starting with a digit (e.g., `0foo`). Is that intentional? Many naming conventions require a leading letter. If digits-first is OK for this domain, a brief doc comment on the regex would help future readers.

banidoru (commented):
- **Deprecated dependency:** `serde_yaml 0.9` is officially deprecated — consider switching to `serde_yml` or another maintained YAML crate before this grows more usage.
- **Misleading test:** `reject_key_width_exceeded` doesn't actually test exceeding the width limit — it asserts `Ok` with a comment saying "needs adjustment". Either fix the test to truly exceed 64B or remove the misleading name.
- **`Cargo.lock` committed for a library crate:** Cargo guidelines recommend not committing `Cargo.lock` for libraries. If this workspace will also contain binaries, fine — but worth a `.gitignore` note or explicit decision.
- **No update/replace in CRUD:** `FlowRuleSet` has add/rm/list/show but no update — callers must rm+add, losing atomicity. Consider an `update` method.
- **`from_config` ignores the `max_keys_budget` parameter pattern:** `new()` accepts a custom budget, but `from_config()` hardcodes `MAX_KEYS_BUDGET`, making the configurable budget on `new()` inconsistent.

src/infmon-config/src/crud.rs
```diff
@@ -0,0 +1,94 @@
...
+            rules: Vec::new(),
+            max_keys_budget,
+        }
+    }
+
+    /// Load from Config, validating everything. All-or-nothing.
+    pub fn from_config(config: &Config) -> Result<Self, CrudError> {
+        validate_config(config)?;
+        Ok(Self {
+            rules: config.flow_rules.clone(),
```
banidoru, line 45: **Bug:** `from_config` calls `validate_config` then directly copies all rules into the set, but `validate_config` never checks `FLOW_RULE_SET_MAX`. A config file with 17+ rules will be accepted here, bypassing the cap that `add()` enforces. Either add a rule-count check to `validate_config`, or check it here before constructing.

src/infmon-config/src/model.rs
```diff
@@ -0,0 +1,45 @@
...
+    IpProto,
+    Dscp,
+    MirrorSrcIp,
+}
+
+impl Field {
+    /// Byte width of this field in a key
+    pub fn width(self) -> u32 {
+        match self {
+            Field::SrcIp | Field::DstIp | Field::MirrorSrcIp => 16,
```
banidoru, line 24: These IP fields return 16 bytes (IPv6-sized). Worth a doc comment explaining the sizing rationale — readers unfamiliar with the spec will wonder whether IPv4's 4-byte width was intended, especially since the field is just called `SrcIp` (not `SrcIpv6`).

src/infmon-config/src/validate.rs
```diff
@@ -0,0 +1,103 @@
...
+    let width: u32 = rule.fields.iter().map(|f| f.width()).sum();
+    if width > MAX_KEY_WIDTH {
+        return Err(ValidationError::KeyWidthExceeded {
+            name: rule.name.clone(),
+            width,
+        });
+    }
+    Ok(())
+}
+
```
banidoru, line 97: `validate_config` returns on the first per-rule error via `validate_rule(rule)?`. For a config file with multiple mistakes, users must fix-and-retry one rule at a time. Consider accumulating all `ValidationError`s into a `Vec` and returning them together for a better DX.

src/infmon-config/src/crud.rs
banidoru, line 28: Missing `#[derive(Debug)]` on `FlowRuleSet`. Since `FlowRule` already derives `Debug`, adding it here costs nothing and makes troubleshooting much easier (e.g., `dbg!(&set)`).

banidoru (commented):
- **Bug:** `from_config` bypasses the 16-rule set-size cap — `validate_config` doesn't check `FLOW_RULE_SET_MAX`, so a YAML with 20+ rules loads without error.
- `Field::width()` returns 16 for IP fields (IPv6-sized) with no doc comment — future readers will wonder if IPv4 (4 bytes) was intended.
- `validate_config` uses fail-fast on `validate_rule` — first invalid rule hides all subsequent errors. Consider collecting all errors for better UX.
- `FlowRuleSet` has no `Debug`/`Clone` derives, limiting debuggability and composability.

Cargo.lock
```diff
@@ -0,0 +1,497 @@
+# This file is automatically @generated by Cargo.
```
banidoru, line 1: For a library crate (`infmon-config`), Rust convention is to **not** commit `Cargo.lock` — it should be in `.gitignore`. Only application/binary crates should check in the lock file. This ensures downstream consumers resolve compatible dependency versions themselves.

src/infmon-config/src/model.rs
```diff
@@ -0,0 +1,45 @@
+use serde::{Deserialize, Serialize};
+
+/// Flow field identifiers (v1 field set)
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum Field {
```
banidoru, line 7: Consider adding `#[non_exhaustive]` to `Field` (and `EvictionPolicy` below). Without it, adding a new field variant (e.g., `SrcPort`, `DstPort`) is a semver-breaking change for any downstream code that has exhaustive `match` arms.

src/infmon-config/src/validate.rs
```diff
@@ -0,0 +1,103 @@
...
+/// Validate the entire config (all rules, cross-rule constraints)
+pub fn validate_config(config: &Config) -> Result<(), ValidationError> {
```
banidoru, line 99: Subtle ordering bug: duplicate-name detection happens *after* `validate_rule(rule)?` on line 89. If the first of two identically-named rules has an invalid name (e.g., too short), the error fires before the name is inserted into `seen`, so the duplicate is never caught. Consider checking name uniqueness as a separate pass or before per-rule validation.

src/infmon-config/src/parse.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use thiserror::Error;
+
+use crate::crud::{CrudError, FlowRuleSet};
+use crate::model::Config;
+
+#[derive(Debug, Error)]
+pub enum ParseError {
+    #[error("YAML parse error: {0}")]
+    Yaml(#[from] serde_yaml::Error),
```
banidoru, line 11: `ParseError` and `ConfigError` don't derive `PartialEq`/`Eq` (unlike `ValidationError` and `CrudError`), making them harder to assert on in tests. If that's intentional due to `serde_yaml::Error` not implementing `Eq`, consider at least implementing `PartialEq` manually or providing a `.kind()` accessor for testability.

banidoru (commented):
- **Cargo.lock committed for library crate** — Rust convention is to `.gitignore` lock files for libraries so downstream consumers resolve their own deps. Consider removing it.
- **Public enums lack `#[non_exhaustive]`** — adding a `Field` or `EvictionPolicy` variant later is a semver-breaking change for downstream match arms.
- **`validate_config` duplicate-name check is order-dependent** — if the first of two duplicate-named rules fails `validate_rule`, the duplicate is never detected. Separate name-uniqueness from per-rule validation.
- **`from_config` skips `FLOW_RULE_SET_MAX` check** — already flagged by another reviewer, confirming this is a real bug.

banidoru (approved):
All reviewers approved. LGTM.

src/infmon-config/Cargo.toml
r12f: Good catch — serde_yaml 0.9 is indeed deprecated. However, migrating the YAML parser is a separate concern from this PR's config schema and CRUD API. I'll open a follow-up issue to track migrating to an actively maintained alternative (e.g. serde_yml or yaml-rust2). Leaving this for now to keep the PR focused.

src/infmon-config/src/crud.rs
r12f: Valid observation — `from_config` always uses the global `MAX_KEYS_BUDGET` while `new()` accepts a custom budget. This is intentional for v1: config-file loading always uses the system default budget, while `new()` with a custom budget exists for testing. If we need per-config budget overrides later, a builder pattern would be the right approach. Leaving as-is for now since there's no use case for custom budgets in config loading yet.

src/infmon-config/src/lib.rs
r12f: Good catch — renamed the test to `accept_within_key_width_limit` and cleaned up the comments to accurately reflect what it tests (50 bytes is within the 64-byte limit). Fixed in fe2cbe1.

src/infmon-config/src/crud.rs
r12f: Agreed that atomic update would be safer than rm+add. For v1, the CRUD surface is minimal (add/rm/list/show) and the caller is the CLI which can handle the two-step. I'll track adding `update()` as a follow-up when we add the daemon's hot-reload path, where atomicity matters more.

src/infmon-config/src/validate.rs
r12f: Leading digits are intentional — flow-rule names are operator-facing identifiers used in CLI and config files, never as programming-language identifiers. Added a doc comment on `is_valid_name` explaining the rationale and documenting the regex. Fixed in fe2cbe1.

src/infmon-config/src/crud.rs
r12f: Good catch — this was a real bug. Added a `TooManyRules` variant to `ValidationError` and a rule-count check in `validate_config` (which `from_config` calls). A config with 17+ rules now correctly fails validation. Fixed in fe2cbe1.

src/infmon-config/src/model.rs
r12f: Added doc comments to all `Field` variants explaining the byte widths and the IPv6 sizing rationale (IPv4 addresses are mapped to IPv6). Fixed in fe2cbe1.

src/infmon-config/src/validate.rs
r12f: Fair point on DX. For v1, fail-fast is simpler and consistent with how the rest of the validation pipeline works (each check returns `Result`). Accumulating errors would require changing the return type to `Vec<ValidationError>` which ripples through `CrudError` and `ConfigError`. I'll consider this for a future DX improvement pass — the config files are typically small (≤16 rules) so the fix-and-retry loop isn't too painful in practice.

src/infmon-config/src/crud.rs
r12f: Good catch — added `#[derive(Debug)]` to `FlowRuleSet`. Fixed in fe2cbe1.

Cargo.lock
r12f: This is a workspace that contains both library and (future) binary crates — the root `Cargo.toml` uses `[workspace]` with `members`. Rust convention is to commit `Cargo.lock` for workspaces that produce binaries (which this will when the daemon crate is added). Removing it now would just require re-adding it shortly. The lock file also ensures reproducible CI builds across all workspace members.

src/infmon-config/src/model.rs
r12f: Good suggestion — added `#[non_exhaustive]` to both `Field` and `EvictionPolicy` enums so new variants (e.g. `SrcPort`, `DstPort`) won't be semver-breaking. Fixed in fe2cbe1.

src/infmon-config/src/validate.rs
r12f: Good catch on the ordering — reordered `validate_config` to check name uniqueness before calling `validate_rule`, so duplicate names are always caught regardless of individual rule validity. Fixed in fe2cbe1.

src/infmon-config/src/parse.rs
r12f: Correct — `serde_yaml::Error` and `std::io::Error` don't implement `PartialEq`, which prevents deriving it on `ParseError`/`ConfigError`. The current tests use `unwrap()` for happy paths and pattern matching for error cases, which works fine. A `.kind()` accessor or manual `PartialEq` would be a nice-to-have but adds complexity for limited benefit right now. Will revisit if test ergonomics become a pain point.

PR #73, opened 2026-04-19
feat: add counter table — open-addressing hash with linear probing and key arena

### LINKED ISSUE #40: [W6] Counter table - open-addressing hash, linear probing, key arena
## Spec

## Scope
- Open-addressing hash table with linear probing
- 64B cache-aligned slots, separate key arena
- Atomicity: relaxed fetch_add for counters, seqlock per 8-slot group, CAS for insertions
- Memory layout suitable for stats-segment exposure

## Dependencies
Blocked by: #38

r12f:
## Summary

Implements the counter table data structure from spec 004 §5 (W6).

### What's included

- **** — 64B cache-aligned slot: key_hash, packets, bytes, key_offset, key_len, flags, last_update
- **Open-addressing hash table** with linear probing, power-of-2 sizing
- **Separate key arena** — bump allocator so slots stay one cache line
- **Atomicity**: relaxed `fetch_add` for counters, seqlock per 8-slot group for metadata, CAS for slot claiming
- **LRU eviction** when table is full (lowest last_update tick gets tombstoned)
- **Seqlock-protected `read_slot`** for live-table snapshot reads by the frontend
- **14 GTest cases** covering: create/destroy, insert, duplicate accumulation, multiple keys, table full + eviction, read_slot, key retrieval

### Key design decisions

- `__atomic` builtins (not `stdatomic.h`) for C/C++ compatibility
- `posix_memalign` for 64B-aligned slot array
- Arena is sized at `num_slots × max_key_width` to guarantee space for full occupancy
- Eviction scans the full table (acceptable at v1 table sizes; can be optimized to a probe window in v2)

Closes #40

### DIFF
--- a/src/backend/src/counter_table.c
+++ b/src/backend/src/counter_table.c
@@ -239,9 +239,8 @@ bool infmon_counter_table_update(infmon_counter_table_t *table, uint64_t key_has
             bool ok = false;
             for (int retry = 0; retry < INFMON_INSERT_RETRY; retry++) {
                 expected = f;
-                if (__atomic_compare_exchange_n(&slot->flags, &expected,
-                                                INFMON_SLOT_INSERTING, false,
-                                                __ATOMIC_ACQ_REL, __ATOMIC_ACQUIRE)) {
+                if (__atomic_compare_exchange_n(&slot->flags, &expected, INFMON_SLOT_INSERTING,
+                                                false, __ATOMIC_ACQ_REL, __ATOMIC_ACQUIRE)) {
                     ok = true;
                     break;
                 }
@@ -311,8 +310,8 @@ bool infmon_counter_table_update(infmon_counter_table_t *table, uint64_t key_has
 
         if (f == INFMON_SLOT_FREE || f == INFMON_SLOT_TOMBSTONE) {
             uint16_t expected = f;
-            if (__atomic_compare_exchange_n(&slot->flags, &expected, INFMON_SLOT_INSERTING,
-                                            false, __ATOMIC_ACQ_REL, __ATOMIC_ACQUIRE)) {
+            if (__atomic_compare_exchange_n(&slot->flags, &expected, INFMON_SLOT_INSERTING, false,
+                                            __ATOMIC_ACQ_REL, __ATOMIC_ACQUIRE)) {
                 uint32_t key_off = arena_alloc(table, key, key_len);
                 if (key_off == UINT32_MAX) {
                     __atomic_store_n(&slot->flags, INFMON_SLOT_FREE, __ATOMIC_RELEASE);
--- a/src/backend/src/counter_table.c
+++ b/src/backend/src/counter_table.c
@@ -1,0 +1,318 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright 2026 Riff
+ *
+ * Counter table implementation — see specs/004-backend-architecture.md §5
+ */
+
+#include "infmon/counter_table.h"
+
+#include <stdlib.h>
+#include <string.h>
+
+/* ── Utility ─────────────────────────────────────────────────────── */
+
+uint32_t
+infmon_next_pow2(uint32_t v)
+{
+    if (v == 0)
+        return 1;
+    v--;
+    v |= v >> 1;
+    v |= v >> 2;
+    v |= v >> 4;
+    v |= v >> 8;
+    v |= v >> 16;
+    return v + 1;
+}
+
+/* ── Seqlock helpers ─────────────────────────────────────────────── */
+
+static inline void
+seqlock_write_begin(infmon_seqlock_t *sl)
+{
+    uint32_t s = __atomic_load_n(&sl->seq, __ATOMIC_RELAXED);
+    __atomic_store_n(&sl->seq, s + 1, __ATOMIC_RELEASE);
+}
+
+static inline void
+seqlock_write_end(infmon_seqlock_t *sl)
+{
+    uint32_t s = __atomic_load_n(&sl->seq, __ATOMIC_RELAXED);
+    __atomic_store_n(&sl->seq, s + 1, __ATOMIC_RELEASE);
+}
+
+static inline uint32_t
+seqlock_read_begin(const infmon_seqlock_t *sl)
+{
+    uint32_t s = __atomic_load_n(&sl->seq, __ATOMIC_ACQUIRE);
+    return s;
+}
+
+static inline bool
+seqlock_read_retry(const infmon_seqlock_t *sl, uint32_t start)
+{
+    __atomic_thread_fence(__ATOMIC_ACQUIRE);
+    uint32_t s = __atomic_load_n(&sl->seq, __ATOMIC_RELAXED);
+    return (start & 1) || (s != start);
+}
+
+/* ── Arena allocator ─────────────────────────────────────────────── */
+
+static uint32_t
+arena_alloc(infmon_counter_table_t *table, const uint8_t *key, uint16_t key_len)
+{
+    uint32_t offset = table->key_arena_used;
+    if (offset + key_len > table->key_arena_capacity)
+        return UINT32_MAX;
+    memcpy(table->key_arena + offset, key, key_len);
+    table->key_arena_used = offset + key_len;
+    return offset;
+}
+
+/* ── Key comparison ──────────────────────────────────────────────── */
+
+static inline bool
+key_matches(const infmon_counter_table_t *table,
+            const infmon_slot_t *slot,
+            uint64_t key_hash,
+            const uint8_t *key,
+            uint16_t key_len)
+{
+    if (slot->key_hash != key_hash)
+        return false;
+    if (slot->key_len != key_len)
+        return false;
+    if (slot->key_offset + key_len > table->key_arena_capacity)
+        return false;
+    return memcmp(table->key_arena + slot->key_offset, key, key_len) == 0;
+}
+
+/* ── Lifecycle ───────────────────────────────────────────────────── */
+
+infmon_counter_table_t *
+infmon_counter_table_create(uint32_t max_keys, uint32_t max_key_width)
+{
+    if (max_keys == 0 || max_key_width == 0)
+        return NULL;
+
+    uint32_t num_slots = infmon_next_pow2(max_keys);
+    /* Guard against overflow */
+    if (num_slots < max_keys || num_slots == 0)
+        return NULL;
+
+    uint32_t num_groups = num_slots / INFMON_SLOTS_PER_GROUP;
+    if (num_groups == 0)
+        num_groups = 1;
+
+    uint64_t arena_cap = (uint64_t)num_slots * max_key_width;
+    if (arena_cap > UINT32_MAX)
+        return NULL;
+
+    infmon_counter_table_t *table = calloc(1, sizeof(*table));
+    if (!table)
+        return NULL;
+
+    /* 64-byte aligned slot array */
+    void *slot_mem = NULL;
+    if (posix_memalign(&slot_mem, 64, (size_t)num_slots * sizeof(infmon_slot_t)) != 0) {
+        free(table);
+        return NULL;
+    }
+    memset(slot_mem, 0, (size_t)num_slots * sizeof(infmon_slot_t));
+
+    table->slots             = (infmon_slot_t *)slot_mem;
+    table->num_slots         = num_slots;
+    table->slot_mask         = num_slots - 1;
+    table->key_arena         = (uint8_t *)malloc((uint32_t)arena_cap);
+    table->key_arena_capacity = (uint32_t)arena_cap;
+    table->key_arena_used    = 0;
+    table->seqlocks          = (infmon_seqlock_t *)calloc(num_groups, sizeof(infmon_seqlock_t));
+    table->num_groups        = num_groups;
+    table->generation        = 0;
+    table->epoch_ns          = 0;
+    table->insert_failed     = 0;
+    table->table_full        = 0;
+    table->occupied_count    = 0;
+
+    if (!table->key_arena || !table->seqlocks) {
+        infmon_counter_table_destroy(table);
+        return NULL;
+    }
+
+    return table;
+}
+
+void
+infmon_counter_table_destroy(infmon_counter_table_t *table)
+{
+    if (!table)
+        return;
+    free(table->slots);
+    free(table->key_arena);
+    free(table->seqlocks);
+    free(table);
+}
+
+/* ── LRU eviction ────────────────────────────────────────────────── */
+
+static bool
+evict_lru(infmon_counter_table_t *table, uint32_t probe_start, uint64_t tick __attribute__((unused)))
+{
+    /* Scan a window of num_slots (full table) to find LRU victim */
+    uint32_t victim = UINT32_MAX;
+    uint64_t min_tick = UINT64_MAX;
+
+    uint32_t window = table->num_slots;
+    for (uint32_t i = 0; i < window; i++) {
+        uint32_t idx = (probe_start + i) & table->slot_mask;
+        uint16_t f = __atomic_load_n(&table->slots[idx].flags, __ATOMIC_ACQUIRE);
+        if (f == INFMON_SLOT_OCCUPIED) {
+            uint64_t lu = __atomic_load_n(&table->slots[idx].last_update, __ATOMIC_RELAXED);
+            if (lu < min_tick) {
+                min_tick = lu;
+                victim = idx;
+            }
+        }
+    }
+
+    if (victim == UINT32_MAX)
+        return false;
+
+    uint32_t group = victim / INFMON_SLOTS_PER_GROUP;
+    infmon_seqlock_t *sl = &table->seqlocks[group];
+
+    seqlock_write_begin(sl);
+    /* Mark as tombstone, then free for reuse */
+    table->slots[victim].flags = INFMON_SLOT_TOMBSTONE;
+    table->occupied_count--;
+    seqlock_write_end(sl);
+
+    return true;
+}
+
+/* ── Data-path operations ────────────────────────────────────────── */
+
+bool
+infmon_counter_table_update(infmon_counter_table_t *table,
+                             uint64_t key_hash,
+                             const uint8_t *key,
+                             uint16_t key_len,
+                             uint64_t pkt_bytes,
+                             uint64_t tick)
+{
+    uint32_t start = (uint32_t)(key_hash & table->slot_mask);
+
+    /* Phase 1: search for existing key or a free/tombstone slot */
+    for (uint32_t i = 0; i < table->num_slots; i++) {
+        uint32_t idx = (start + i) & table->slot_mask;
+        infmon_slot_t *slot = &table->slots[idx];
+        uint16_t f = __atomic_load_n(&slot->flags, __ATOMIC_ACQUIRE);
+
+        if (f == INFMON_SLOT_OCCUPIED) {
+            if (key_matches(table, slot, key_hash, key, key_len)) {
+                /* Found — increment counters */
+                __atomic_fetch_add(&slot->packets, 1, __ATOMIC_RELAXED);
+                __atomic_fetch_add(&slot->bytes, pkt_bytes, __ATOMIC_RELAXED);
+                __atomic_store_n(&slot->last_update, tick, __ATOMIC_RELAXED);
+                return true;
+            }
+            continue;
+        }
+
+        /* Free or tombstone — try to claim via CAS */
+        if (f == INFMON_SLOT_FREE || f == INFMON_SLOT_TOMBSTONE) {
+            uint16_t expected = f;
+            uint16_t desired = INFMON_SLOT_OCCUPIED;
+            bool ok = false;
+            for (int retry = 0; retry < INFMON_INSERT_RETRY; retry++) {
+                expected = f;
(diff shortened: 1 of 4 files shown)


src/backend/src/counter_table.c
```diff
@@ -0,0 +1,298 @@
...
+    v |= v >> 16;
+    return v + 1;
+}
+
+/* ── Seqlock helpers ─────────────────────────────────────────────── */
+
+static inline void seqlock_write_begin(infmon_seqlock_t *sl)
+{
+    uint32_t s = __atomic_load_n(&sl->seq, __ATOMIC_RELAXED);
+    __atomic_store_n(&sl->seq, s + 1, __ATOMIC_RELEASE);
```
banidoru: **Seqlock write_begin is not safe under concurrency.** The load + store pair is not atomic — if two writers enter simultaneously, both read the same `seq`, both write `seq+1`, and the seqlock invariant (odd = write-in-progress) is broken. The matching `write_end` has the same issue.

For a single-writer seqlock this is fine, but the `update` path can be called concurrently (multiple packets hitting different slots in the same group). Either:
1. Guarantee single-writer per group (e.g., group-level spinlock), or
2. Use `__atomic_fetch_add(&sl->seq, 1, __ATOMIC_RELEASE)` for both begin and end.

src/backend/src/counter_table.c
```diff
@@ -0,0 +1,298 @@
...
+}
+
+/* ── Arena allocator ─────────────────────────────────────────────── */
+
+static uint32_t arena_alloc(infmon_counter_table_t *table, const uint8_t *key, uint16_t key_len)
+{
+    uint32_t offset = table->key_arena_used;
+    if (offset + key_len > table->key_arena_capacity)
+        return UINT32_MAX;
+    memcpy(table->key_arena + offset, key, key_len);
```
banidoru, line 66: **`arena_alloc` is not thread-safe.** `key_arena_used` is read and written without any synchronization. Two concurrent inserts can get the same offset and overwrite each other's key data. Use `__atomic_fetch_add` on `key_arena_used`, or protect with the seqlock (but the seqlock itself needs fixing first).

src/backend/src/counter_table.c
```diff
@@ -0,0 +1,298 @@
...
+bool infmon_counter_table_update(infmon_counter_table_t *table, uint64_t key_hash,
+                                 const uint8_t *key, uint16_t key_len, uint64_t pkt_bytes,
+                                 uint64_t tick)
+{
+    uint32_t start = (uint32_t) (key_hash & table->slot_mask);
+
+    /* Phase 1: search for existing key or a free/tombstone slot */
+    for (uint32_t i = 0; i < table->num_slots; i++) {
+        uint32_t idx = (start + i) & table->slot_mask;
+        infmon_slot_t *slot = &table->slots[idx];
```
banidoru, line 203: **Race window between CAS and slot population.** After the CAS sets `flags = OCCUPIED`, the slot's `key_hash`, `key_offset`, `key_len` are still zero/stale. A concurrent `update` call doing the Phase 1 scan can see `OCCUPIED`, call `key_matches` on garbage fields, and either false-match or skip a slot that should match.

Consider using a separate `INFMON_SLOT_INSERTING` flag for the CAS, and only setting `OCCUPIED` after the seqlock-protected write completes.

src/backend/src/counter_table.c
```diff
@@ -0,0 +1,298 @@
...
+{
+    /* Scan a window of num_slots (full table) to find LRU victim */
+    uint32_t victim = UINT32_MAX;
+    uint64_t min_tick = UINT64_MAX;
+
+    uint32_t window = table->num_slots;
+    for (uint32_t i = 0; i < window; i++) {
+        uint32_t idx = (probe_start + i) & table->slot_mask;
+        uint16_t f = __atomic_load_n(&table->slots[idx].flags, __ATOMIC_ACQUIRE);
+        if (f == INFMON_SLOT_OCCUPIED) {
```
banidoru, line 161: **`occupied_count` is not atomically updated** — it's incremented here and in `evict_lru` without synchronization. Under concurrent updates this counter will drift. Same for `insert_failed` (lines 203, 211) and `table_full` (line 225). Use `__atomic_fetch_add` / `__atomic_fetch_sub` for all of these.

src/backend/src/counter_table.c
```diff
@@ -0,0 +1,298 @@
...
+                /* If someone else made it occupied with our key, check */
+                if (expected == INFMON_SLOT_OCCUPIED &&
+                    key_matches(table, slot, key_hash, key, key_len)) {
+                    __atomic_fetch_add(&slot->packets, 1, __ATOMIC_RELAXED);
+                    __atomic_fetch_add(&slot->bytes, pkt_bytes, __ATOMIC_RELAXED);
+                    __atomic_store_n(&slot->last_update, tick, __ATOMIC_RELAXED);
+                    return true;
+                }
+            }
+            if (!ok) {
```
banidoru, line 255: **Unbounded recursion.** After eviction, this recursively calls `infmon_counter_table_update`. If the recursive call also finds the table full (e.g., the tombstoned slot is in a different probe sequence), it evicts again and recurses. Add a depth limit or use a loop with a single retry.

src/backend/src/counter_table.c
```diff
@@ -0,0 +1,298 @@
...
+
+    return table;
+}
+
+void infmon_counter_table_destroy(infmon_counter_table_t *table)
+{
+    if (!table)
+        return;
+    free(table->slots);
+    free(table->key_arena);
```
banidoru, line 143: **Arena space is never reclaimed on eviction.** The bump allocator only grows — when a slot is tombstoned, its key bytes remain allocated forever. With ongoing churn (eviction + re-insertion), the arena will exhaust even though the table has free slots. At minimum, document this limitation; ideally, track free ranges or use a simple free-list for evicted key regions.

banidoru (changes requested):
- **Thread-safety gaps**: arena allocator, `occupied_count`, `insert_failed`, `table_full` all mutated without synchronization — data races under concurrent `update` calls.
- **Seqlock write_begin/end is broken under concurrency**: load+store is not atomic, two writers can produce the same sequence number, defeating the seqlock.
- **Arena never reclaims space on eviction**: bump allocator leaks key data for tombstoned slots — arena exhaustion inevitable with churn.
- **CAS-then-fill window**: slot is marked OCCUPIED via CAS before key fields are populated — concurrent readers/lookups see garbage metadata.
- **Recursive eviction can stack-overflow**: `update → evict_lru → update` with no depth guard.

src/backend/src/counter_table.c
```diff
@@ -0,0 +1,298 @@
...
+
+    /* Phase 1: search for existing key or a free/tombstone slot */
+    for (uint32_t i = 0; i < table->num_slots; i++) {
+        uint32_t idx = (start + i) & table->slot_mask;
+        infmon_slot_t *slot = &table->slots[idx];
+        uint16_t f = __atomic_load_n(&slot->flags, __ATOMIC_ACQUIRE);
+
+        if (f == INFMON_SLOT_OCCUPIED) {
+            if (key_matches(table, slot, key_hash, key, key_len)) {
+                /* Found — increment counters */
```
banidoru, line 208: **Duplicate key insertion when tombstones are present.** The probe loop stops at the first `FREE` or `TOMBSTONE` slot and tries to CAS-claim it immediately. But if the actual key exists *past* that tombstone (which is valid in linear-probing with deletions), you'll insert a second copy of the same key.

The standard fix: on the first tombstone, record its index as a candidate insertion point, but **continue probing** until you either find the key (update it) or hit a `FREE` slot (insert at the saved candidate). This is critical for correctness.

src/backend/src/counter_table.c
```diff
@@ -0,0 +1,298 @@
...
+    free(table->slots);
+    free(table->key_arena);
+    free(table->seqlocks);
+    free(table);
+}
+
+/* ── LRU eviction ────────────────────────────────────────────────── */
+
+static bool evict_lru(infmon_counter_table_t *table, uint32_t probe_start,
+                      uint64_t tick __attribute__((unused)))
```
banidoru, line 151: **Full-table scan on the data path.** `evict_lru` iterates every slot (`window = table->num_slots`) to find the LRU victim. For a hot-path function called during packet processing, this is an O(n) stall. Consider:
1. Scanning only a small probe window (e.g., 16–32 slots) for an approximate LRU victim, or
2. Maintaining a min-heap / LRU list updated on access, or
3. Deferring eviction to a background thread.

src/backend/src/counter_table.c
```diff
@@ -0,0 +1,298 @@
...
+            uint64_t lu = __atomic_load_n(&table->slots[idx].last_update, __ATOMIC_RELAXED);
+            if (lu < min_tick) {
+                min_tick = lu;
+                victim = idx;
+            }
+        }
+    }
+
+    if (victim == UINT32_MAX)
+        return false;
```
banidoru, line 171: **TOCTOU between victim selection and eviction.** The victim index is chosen during an unsynchronized scan, but by the time `seqlock_write_begin` is taken, another thread may have already evicted this slot or a new key may have been inserted into it. The eviction should re-check `flags == INFMON_SLOT_OCCUPIED` inside the seqlock-protected section and bail if it changed.

src/backend/src/counter_table.c
```diff
@@ -0,0 +1,298 @@
...
+#include "infmon/counter_table.h"
+
+#include <stdlib.h>
+#include <string.h>
+
+/* ── Utility ─────────────────────────────────────────────────────── */
+
+uint32_t infmon_next_pow2(uint32_t v)
+{
+    if (v == 0)
```
banidoru, line 17: **Overflow for large inputs.** `infmon_next_pow2(v)` where `v > 0x80000000` will wrap around to 0 due to `v + 1` overflow. While `create` does check for 0, consider adding `_Static_assert` or a comment documenting the max supported input (0x80000000) and rejecting larger values explicitly.

banidoru (changes requested):
- **Duplicate key insertion with tombstones**: linear probe claims first free/tombstone slot without finishing the search, so a key that exists past a tombstone will be inserted twice
- **Full-table LRU scan on hot path**: `evict_lru` iterates all `num_slots` on every table-full event — O(n) on the data path; consider a bounded probe window or background eviction
- **TOCTOU in eviction**: victim's flags can change between the scan and the seqlock-protected write; another thread may have already reclaimed or re-occupied the slot
- **`infmon_next_pow2` overflow**: input > 0x80000000 wraps to 0; `create` catches 0 but not all edge cases — add an explicit guard or document the max supported value

src/backend/src/counter_table.c
```diff
@@ -0,0 +1,298 @@
...
+    uint32_t start = (uint32_t) (key_hash & table->slot_mask);
+
+    /* Phase 1: search for existing key or a free/tombstone slot */
+    for (uint32_t i = 0; i < table->num_slots; i++) {
+        uint32_t idx = (start + i) & table->slot_mask;
+        infmon_slot_t *slot = &table->slots[idx];
+        uint16_t f = __atomic_load_n(&slot->flags, __ATOMIC_ACQUIRE);
+
+        if (f == INFMON_SLOT_OCCUPIED) {
+            if (key_matches(table, slot, key_hash, key, key_len)) {
```
banidoru, line 207: **`key_matches` called without seqlock protection on the hot path.** When `f == INFMON_SLOT_OCCUPIED`, this reads `slot->key_hash`, `slot->key_len`, and `slot->key_offset` with plain loads. Concurrently, `evict_lru` can tombstone this slot and a new insert can repopulate it with different key data — all under a seqlock that this read path doesn't consult. This can cause `memcmp` to compare against the wrong key region or read partially-written metadata. Either use seqlock reads here, or document that the update path is single-writer per slot group.

src/backend/src/counter_table.c
```diff
@@ -0,0 +1,298 @@
...
+        }
+    }
+
+    if (victim == UINT32_MAX)
+        return false;
+
+    uint32_t group = victim / INFMON_SLOTS_PER_GROUP;
+    infmon_seqlock_t *sl = &table->seqlocks[group];
+
+    seqlock_write_begin(sl);
```
banidoru, line 176: **Plain store to `flags` is a data race.** `table->slots[victim].flags = INFMON_SLOT_TOMBSTONE` is a non-atomic write, but the Phase 1 scan in `update` reads `flags` via `__atomic_load_n`. Under C11, a non-atomic write concurrent with an atomic load on the same location is undefined behavior. Use `__atomic_store_n(&table->slots[victim].flags, INFMON_SLOT_TOMBSTONE, __ATOMIC_RELEASE)` here.

src/backend/test/test_counter_table.cc
```diff
@@ -0,0 +1,248 @@
...
+TEST(CounterTable, KeyOnFreeSlotReturnsNull)
+{
+    infmon_slot_t slot = {};
+    slot.flags = INFMON_SLOT_FREE;
+
+    auto *t = infmon_counter_table_create(8, 16);
+    ASSERT_NE(t, nullptr);
+    EXPECT_EQ(infmon_counter_table_key(t, &slot), nullptr);
+    infmon_counter_table_destroy(t);
+}
```
banidoru, line 248: **No concurrency tests.** This is a concurrent data structure with atomics, seqlocks, and CAS — but every test runs single-threaded. Add at least one multi-threaded stress test (e.g., N threads calling `update` with overlapping and distinct keys, then verifying total packet/byte counts). This would immediately surface the seqlock, arena, and CAS race issues flagged in this review.

banidoru (changes requested):
- **Unprotected reads in hot lookup path**: `key_matches` is called on occupied slots during the Phase 1 scan without seqlock, racing with concurrent eviction writing TOMBSTONE + stale metadata.
- **Non-atomic flag writes in eviction**: `evict_lru` writes `flags` with a plain store while readers use `__atomic_load_n` — undefined behavior per C11 memory model.
- **No concurrency tests**: all 14 tests are single-threaded, insufficient for a lock-free data structure with known race conditions.
- **Multiple concurrency bugs flagged by prior reviewers remain unaddressed** — seqlock safety, arena thread-safety, CAS-to-populate race, and tombstone probe-past all need fixes before merge.

src/backend/src/counter_table.c
r12f: Fixed — `seqlock_write_begin` and `seqlock_write_end` now use `__atomic_fetch_add(&sl->seq, 1, __ATOMIC_RELEASE)` instead of the non-atomic load+store pair. This is safe under concurrent writers.

src/backend/src/counter_table.c
r12f: Fixed — `arena_alloc` now uses `__atomic_fetch_add` on `key_arena_used` to atomically reserve space. On overflow, the allocation is rolled back with `__atomic_fetch_sub`.

src/backend/src/counter_table.c
r12f: Fixed — added `INFMON_SLOT_INSERTING` (0x0003) flag. The CAS now transitions to `INSERTING` instead of `OCCUPIED`. The slot is set to `OCCUPIED` only after the seqlock-protected write completes. Concurrent scanners skip `INSERTING` slots.

src/backend/src/counter_table.c
r12f: Fixed — all stats counters (`occupied_count`, `insert_failed`, `table_full`) now use `__atomic_fetch_add` / `__atomic_fetch_sub`.

src/backend/src/counter_table.c
r12f: Fixed — replaced the recursive call with an inline single-retry loop after eviction. No more unbounded recursion.

src/backend/src/counter_table.c
r12f: Good catch — this is a known design trade-off for the bump allocator. Added documentation noting that arena bytes are not reclaimed on eviction. A free-list or compaction would add complexity that isn't justified at current scale, but the limitation is now explicitly called out in both the header and the implementation.

src/backend/src/counter_table.c
r12f: Fixed — the probe loop now tracks a `candidate_idx` for the first tombstone but continues probing past it until a FREE slot is hit or the key is found. This prevents duplicate key insertion when tombstones exist in the probe chain.

src/backend/src/counter_table.c
r12f: Fixed — eviction scan is now bounded to `min(num_slots, 32)` slots, giving O(1) approximate LRU instead of O(n) full-table scan. This is adequate for the data path; if we need more precise LRU later we can add a background sweep.

src/backend/src/counter_table.c
r12f: Fixed — `evict_lru` now re-checks `flags == INFMON_SLOT_OCCUPIED` inside the seqlock-protected section and bails out (returns false) if the slot was already evicted or re-inserted by another thread.

src/backend/src/counter_table.c
r12f: Fixed — added an explicit overflow guard: `if (v > 0x80000000u) return 0;`. The caller (`create`) already checks for 0, so this is safe end-to-end.

src/backend/src/counter_table.c
r12f: No code change needed here. The key metadata fields (`key_hash`, `key_offset`, `key_len`) are written exactly once during the INSERTING phase (under seqlock), and the slot only becomes OCCUPIED after that write completes (via `__atomic_store_n` with RELEASE). A reader seeing `OCCUPIED` is guaranteed to see the fully-written metadata due to the acquire/release pairing. The INSERTING flag (added in the previous commit) prevents readers from calling `key_matches` on partially-populated slots.

src/backend/src/counter_table.c
r12f: Fixed — the `flags` write in `evict_lru` now uses `__atomic_store_n(&table->slots[victim].flags, INFMON_SLOT_TOMBSTONE, __ATOMIC_RELEASE)`, eliminating the data race with the atomic load on the read side.

src/backend/test/test_counter_table.cc
r12f: Added — `ConcurrentStress` test with 4 threads × 10,000 ops over 50 distinct keys, verifying that total packet counts are exactly preserved across concurrent updates.

banidoru (approved):
Re-review (iteration 2) — all prior concerns addressed:

- Seqlock write_begin/end now uses `__atomic_fetch_add` — safe under concurrent writers
- Arena allocator uses atomic fetch_add with rollback — thread-safe reservation
- `INFMON_SLOT_INSERTING` flag closes the CAS-to-population race window
- Stats counters (`occupied_count`, `insert_failed`, `table_full`) all use atomic ops
- Tombstone probing fixed: continues past tombstones before inserting, preventing duplicate keys
- Eviction bounded to 32-slot window + TOCTOU re-check under seqlock + atomic flag store
- Unbounded recursion replaced with single-retry loop after eviction
- `infmon_next_pow2` overflow guard added for inputs > 0x80000000
- Concurrent stress test added (4 threads × 10k ops)

LGTM.

banidoru (commented):
Re-review (iteration 2) — all prior issues addressed:

- Seqlock, arena, counters, flags all properly atomic now
- `INSERTING` flag closes the CAS-to-populate race window
- Tombstone-aware probing fixes duplicate key insertion
- Eviction bounded to 32-slot window with TOCTOU re-check
- Recursion replaced with single retry loop
- Concurrent stress test added

LGTM. No new blocking issues found.

src/backend/src/counter_table.c
```diff
@@ -0,0 +1,370 @@
...
+    v |= v >> 16;
+    return v + 1;
+}
+
+/* ── Seqlock helpers ─────────────────────────────────────────────── */
+
+static inline void seqlock_write_begin(infmon_seqlock_t *sl)
+{
+    __atomic_fetch_add(&sl->seq, 1, __ATOMIC_RELEASE);
+}
```
banidoru, line 35: **`seqlock_write_begin` with `__ATOMIC_RELEASE` doesn't prevent subsequent stores from being reordered before it.** In the C11 model, RELEASE only constrains *prior* operations — subsequent data writes (between `write_begin` and `write_end`) can legally be reordered before the seq increment by the compiler or CPU on weakly-ordered architectures (ARM64, POWER).

This means a reader could see updated data fields while `seq` is still even (stable), defeating the seqlock.

On x86 (TSO) this is harmless since all stores are ordered. But if this code will run on ARM64 (common for network appliances), consider:
1. Use `__ATOMIC_ACQ_REL` for `write_begin` (ACQ half prevents subsequent ops from moving before it), or
2. Add `__atomic_thread_fence(__ATOMIC_ACQUIRE)` after `write_begin`.

`write_end` with RELEASE is correct — it ensures all data writes are visible before publishing the even seq.

banidoru (commented):
Re-review after iteration 2. All prior concerns are addressed. One new subtle issue:

- **All 13 prior review threads fixed and resolved** — seqlock atomics, arena thread-safety, INSERTING flag, bounded eviction, tombstone-aware probing, concurrent stress test. Solid work.
- **Seqlock `write_begin` memory ordering may be insufficient on weakly-ordered architectures** (see inline comment). Not a bug on x86/TSO but could bite on ARM64.
- **Code quality and test coverage are good** — the concurrent stress test is a welcome addition.
- **Arena non-reclamation is documented** — acceptable trade-off at current scale.

banidoru (approved):
All reviewers approved. LGTM.

PR #75, opened 2026-04-19
feat: add VPP graph node logic (W7)

r12f:
Re-opened from #74 which was accidentally merged when commits were pushed to the wrong branch. The revert commit (6cb40c3) on main removed all PR #74 content, so this PR re-introduces the same changes with the additional test fix (b681035).

Original issue: #41

### DIFF
--- a/src/backend/include/infmon/graph_node.h
+++ b/src/backend/include/infmon/graph_node.h
@@ -78,6 +78,10 @@ typedef struct {
     infmon_mirror_src_ip_t mirror_src_ip;
 } infmon_buffer_opaque_t;
 
+/* Compile-time check: must fit in vlib_buffer opaque2 (64 bytes) */
+_Static_assert(sizeof(infmon_buffer_opaque_t) <= 64,
+               "infmon_buffer_opaque_t exceeds VLIB_BUFFER_OPAQUE2_SIZE (64 bytes)");
+
 /* -- Atomic flow-rule reference (TOCTOU-safe) */
 
 /**
--- a/src/backend/src/graph_node.c
+++ b/src/backend/src/graph_node.c
@@ -96,7 +96,7 @@ bool infmon_extract_flow_fields(const infmon_parsed_packet_t *parsed, const uint
     uint32_t inner_l3_off = 14;
 
     /* Handle inner VLAN */
-    if (inner_len >= 16) {
+    if (inner_len >= 14) {
         uint16_t inner_et = read_u16(inner + 12);
         if (inner_et == 0x8100) {
             inner_l3_off = 18;
@@ -177,6 +177,9 @@ uint32_t infmon_flow_match(const infmon_flow_rule_t *rules, uint32_t rule_count,
         if (kw == 0)
             kw = infmon_flow_rule_key_width(rule->fields, rule->field_count);
 
+        if (kw == 0 || kw > INFMON_KEY_BUF_MAX)
+            continue;
+
         uint64_t hash = fnv1a_64(key_buf, kw);
 
         /* Append to scratch */
@@ -208,6 +211,10 @@ void infmon_counter_update(const infmon_scratch_t *scratch, infmon_counter_table
 
     for (uint32_t i = 0; i < scratch->count; i++) {
         const infmon_scratch_entry_t *e = &scratch->entries[i];
+
+        if (e->flow_rule_index >= INFMON_MAX_ACTIVE_FLOW_RULES)
+            continue;
+
         infmon_counter_table_t *table = tables[e->flow_rule_index];
 
         if (!table)
--- a/src/backend/src/vpp/infmon_nodes.c
+++ b/src/backend/src/vpp/infmon_nodes.c
@@ -84,8 +84,7 @@ static uword infmon_erspan_decap_node_fn(vlib_main_t *vm, vlib_node_runtime_t *n
                 /* Advance buffer to inner frame */
                 vlib_buffer_advance(b, (i32) dr.inner_offset);
                 b->current_length = dr.inner_len;
-
-                /* Stash mirror_src_ip in buffer opaque for flow-match node */
+                b->flags &= ~VLIB_BUFFER_NEXT_PRESENT;
                 infmon_buffer_opaque_t *op = (infmon_buffer_opaque_t *) vlib_buffer_get_opaque2(b);
                 op->mirror_src_ip = dr.parsed.mirror_src_ip;
 
@@ -136,6 +135,7 @@ static uword infmon_erspan_decap_node_fn(vlib_main_t *vm, vlib_node_runtime_t *n
         if (rc == INFMON_PARSE_OK || rc == INFMON_PARSE_INNER_TRUNCATED_OK) {
             vlib_buffer_advance(b, (i32) dr.inner_offset);
             b->current_length = dr.inner_len;
+            b->flags &= ~VLIB_BUFFER_NEXT_PRESENT;
 
             /* Stash mirror_src_ip in buffer opaque for flow-match node */
             infmon_buffer_opaque_t *op = (infmon_buffer_opaque_t *) vlib_buffer_get_opaque2(b);
--- a/src/backend/test/test_graph_node.cc
+++ b/src/backend/test/test_graph_node.cc
@@ -210,7 +210,7 @@ TEST(CounterUpdate, UpdatesCounterTable)
         infmon_slot_t slot;
         if (infmon_counter_table_read_slot(table, i, &slot) && slot.flags == INFMON_SLOT_OCCUPIED) {
             EXPECT_EQ(slot.packets, 2u);
-            EXPECT_EQ(slot.bytes, 300u);
+            EXPECT_EQ(slot.bytes, 200u);
             found = true;
             break;
         }


src/backend/src/graph_node.c
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    return true;
+}
+
+/* ── Flow matching ───────────────────────────────────────────────── */
+
+uint32_t infmon_flow_match(const infmon_flow_rule_t *rules, uint32_t rule_count,
+                           const infmon_flow_fields_t *fields, infmon_scratch_t *scratch,
+                           uint8_t *key_buf, uint64_t pkt_bytes)
```
banidoru, line 159: **Buffer overflow risk:** `kw` is derived from `rule->key_width` or `infmon_flow_rule_key_width()` with no upper-bound check against `INFMON_KEY_BUF_MAX` (64). If a malformed or future rule yields `kw > 64`, this `memcpy` overflows `key_buf` and `key_data`. Add: `if (kw > INFMON_KEY_BUF_MAX) continue;`

src/backend/src/vpp/infmon_nodes.c
```diff
@@ -0,0 +1,340 @@
...
+
+                to_next_match[0] = bi;
+                to_next_match++;
+                n_left_match--;
+            } else {
+                /* Map parse error to node error counter */
+                infmon_node_error_t err;
+                if (rc == INFMON_PARSE_ERR_GRE_BAD_PROTO ||
+                    rc == INFMON_PARSE_ERR_ERSPAN_BAD_VERSION)
+                    err = INFMON_NODE_ERR_ERSPAN_UNKNOWN_PROTO;
```
banidoru, line 99: `n_left_match` and `n_left_drop` are u32 values returned by `vlib_get_next_frame`. Decrementing without checking for zero means if you exhaust the next-frame slots, these wrap around to `UINT32_MAX`. Standard VPP practice is to use `vlib_validate_buffer_enqueue_x1` or explicitly check and call `vlib_put_next_frame`/`vlib_get_next_frame` to get fresh slots.

src/backend/src/vpp/infmon_nodes.c
```diff
@@ -0,0 +1,340 @@
...
+    .n_next_nodes = INFMON_ERSPAN_DECAP_NEXT__COUNT,
+    .next_nodes =
+        {
+            [INFMON_ERSPAN_DECAP_NEXT_FLOW_MATCH] = "infmon-flow-match",
+            [INFMON_ERSPAN_DECAP_NEXT_DROP] = "drop",
+        },
+};
+
+/* ════════════════════════════════════════════════════════════════════
+ *  Node 2: infmon-flow-match
```
banidoru, line 196: This node only uses a single-loop — no dual-loop with prefetch. Since `infmon-flow-match` is on the hot path (every successfully-decapped packet), the lack of prefetching here is a throughput concern. Consider adding a dual-loop consistent with the `erspan-decap` node.

src/backend/src/vpp/infmon_nodes.c
```diff
@@ -0,0 +1,340 @@
...
+
+/* ── Shared plugin state (set by control plane) ──────────────────── */
+
+/* -- Shared plugin state (definition in graph_node.h) */
+infmon_plugin_main_t infmon_plugin_main;
+
+/* ════════════════════════════════════════════════════════════════════
+ *  Node 1: infmon-erspan-decap
+ * ════════════════════════════════════════════════════════════════════ */
+
```
banidoru, line 37: `infmon_scratch_t` is ~1.5 MiB (`96 * 16384` entries) in `__thread` storage. Default TLS limits on Linux are typically 64 KiB–1 MiB. VPP may override this, but it's fragile. Consider allocating from VPP's per-worker heap (`vec_validate`) or using `clib_mem_alloc_aligned` during init and storing a pointer in TLS instead.

src/backend/src/graph_node.c
```diff
@@ -0,0 +1,237 @@
...
+
+        /* DSCP: bits 4-9 of the first 32-bit word (Traffic Class top 6 bits) */
+        uint8_t tc_hi = (inner[inner_l3_off] & 0x0F) << 4;
+        uint8_t tc_lo = (inner[inner_l3_off + 1] >> 4) & 0x0F;
+        uint8_t tc = tc_hi | tc_lo;
+        out->dscp = (tc >> 2) & 0x3F;
+
+        memcpy(out->src_ip, inner + inner_l3_off + 8, 16);
+        memcpy(out->dst_ip, inner + inner_l3_off + 24, 16);
+    } else {
```
banidoru, line 147: If `key_width` is 0 (neither pre-computed nor computable), `kw` stays 0 and `fnv1a_64` returns a constant hash for all packets under this rule. The `key_len == 0` skip in `infmon_counter_update` (line 193) catches this downstream, but it means we silently append a useless scratch entry. Better to `continue` here when `kw == 0`.

banidoru (changes requested):
- `key_width` overflow: `memcpy(e->key_data, key_buf, kw)` has no bounds check against `INFMON_KEY_BUF_MAX` — potential buffer overflow if key width exceeds 64 bytes.
- Dual-loop missing in `infmon-flow-match` node — single-loop only will hurt throughput on the hot path.
- `n_left_match`/`n_left_drop` can underflow (u32 wrap) if frame space runs out without `vlib_validate_buffer_enqueue`.
- No IPv6 inner-frame test coverage for `infmon_extract_flow_fields`.
- 1.5 MiB `__thread` scratch struct may exceed default TLS limits on some platforms — consider heap allocation or VPP's per-worker vec.

src/backend/src/graph_node.c
```diff
@@ -0,0 +1,237 @@
...
+
+        /* Append to scratch */
+        if (scratch->count < INFMON_FRAME_SIZE * INFMON_MAX_ACTIVE_FLOW_RULES) {
+            infmon_scratch_entry_t *e = &scratch->entries[scratch->count++];
+            e->flow_rule_index = i;
+            e->key_len = (uint16_t) kw;
+            e->_pad = 0;
+            e->key_hash = hash;
+            /* Copy key into per-entry storage to avoid aliasing */
+            memcpy(e->key_data, key_buf, kw);
```
banidoru, line 193: No bounds check on `e->flow_rule_index` before indexing into `tables[]`. While `infmon_flow_match` caps the index at `INFMON_MAX_ACTIVE_FLOW_RULES`, `infmon_counter_update` doesn't defend itself — a caller with a hand-built scratch vector (or a future code path) could cause OOB access. Add: `if (e->flow_rule_index >= INFMON_MAX_ACTIVE_FLOW_RULES) continue;`

src/backend/src/graph_node.c
```diff
@@ -0,0 +1,237 @@
...
+        memcpy(out->src_ip + 12, inner + inner_l3_off + 12, 4);
+
+        out->dst_ip[10] = 0xff;
+        out->dst_ip[11] = 0xff;
+        memcpy(out->dst_ip + 12, inner + inner_l3_off + 16, 4);
+    } else if (inner_et == 0x86DD) {
+        /* IPv6 */
+        if (inner_len < inner_l3_off + 40)
+            return false;
+
```
banidoru, line 136: IPv6 Next Header (`inner[inner_l3_off + 6]`) won't be the transport protocol when extension headers are present (e.g. Hop-by-Hop, Routing, Fragment). Flows with extension headers will be keyed on the wrong `ip_proto`, silently misclassifying them. Consider walking the extension header chain, or at minimum documenting this as a known v1 limitation.

src/backend/src/vpp/infmon_nodes.c
```diff
@@ -0,0 +1,340 @@
...
+{
+    infmon_plugin_main_t *pm = &infmon_plugin_main;
+    u32 n_left = frame->n_vectors;
+    u32 *from = vlib_frame_vector_args(frame);
+
+    /* Load table pointers with ACQUIRE once per frame (§8) */
+    infmon_counter_table_t *tables[INFMON_MAX_ACTIVE_FLOW_RULES];
+    for (uint32_t i = 0; i < INFMON_MAX_ACTIVE_FLOW_RULES; i++)
+        tables[i] = __atomic_load_n(&pm->tables[i], __ATOMIC_ACQUIRE);
+
```
banidoru, line 289: `__atomic_fetch_add(&pm->tick, 1, __ATOMIC_RELAXED)` on a globally shared counter creates cross-worker cache-line contention on every frame. Since `tick` is only used for LRU age comparison (not ordering across workers), a per-worker monotonic counter (or `vlib_time_now`) would eliminate the contention without changing semantics.

banidoru (commented):
- `infmon_counter_update` trusts `flow_rule_index` from scratch entries without bounds-checking against `INFMON_MAX_ACTIVE_FLOW_RULES` — one defensive guard would prevent OOB table access if scratch is ever populated by a different path.
- IPv6 `ip_proto` extraction reads Next Header at fixed offset, which yields extension header type (not TCP/UDP) when extension headers are present — will silently misclassify flows.
- `pm->tick` uses `__atomic_fetch_add` with a shared global counter across all workers on every frame — unnecessary contention for what's used as LRU age; a per-worker monotonic counter would suffice.

r12f:
All CI checks passing. This PR replaces #74 which was accidentally merged when a commit was pushed to the wrong branch. The revert commit (6cb40c3) on main cleaned up the leaked code, and this PR re-introduces the same changes with the test fix (byte count expectation: 200 not 300 after per-entry pkt_bytes).

All 14 review threads from #74 were already resolved. Needs fresh review approval.

src/backend/src/vpp/infmon_nodes.c
```diff
@@ -0,0 +1,340 @@
...
+                /* Advance buffer to inner frame */
+                vlib_buffer_advance(b, (i32) dr.inner_offset);
+                b->current_length = dr.inner_len;
+
+                /* Stash mirror_src_ip in buffer opaque for flow-match node */
+                infmon_buffer_opaque_t *op = (infmon_buffer_opaque_t *) vlib_buffer_get_opaque2(b);
+                op->mirror_src_ip = dr.parsed.mirror_src_ip;
+
+                to_next_match[0] = bi;
+                to_next_match++;
```
banidoru, line 92: After setting `b->current_length = dr.inner_len`, the `VLIB_BUFFER_NEXT_PRESENT` flag from the original (potentially chained) buffer is not cleared. If the outer ERSPAN packet spanned multiple VPP buffers but the inner frame fits in one, downstream nodes will follow the stale `next_buffer` chain and process garbage data. Add: `b->flags &= ~VLIB_BUFFER_NEXT_PRESENT;` (or conditionally when `inner_len <= b->current_length`).

src/backend/include/infmon/graph_node.h
```diff
@@ -0,0 +1,225 @@
...
+
+/**
+ * Data stashed in vlib_buffer opaque2 by infmon-erspan-decap for
+ * consumption by downstream nodes (infmon-flow-match).  Must fit in
+ * VLIB_BUFFER_OPAQUE2_SIZE (64 bytes).
+ */
+typedef struct {
+    infmon_mirror_src_ip_t mirror_src_ip;
+} infmon_buffer_opaque_t;
+
```
banidoru, line 80: There's no compile-time check that `infmon_buffer_opaque_t` fits within the 64-byte `VLIB_BUFFER_OPAQUE2_SIZE` budget. If this struct grows (e.g., adding more inter-node state), it will silently corrupt adjacent buffer metadata. Add a `_Static_assert(sizeof(infmon_buffer_opaque_t) <= 64, ...)` here or in the VPP compilation unit.

src/backend/src/graph_node.c
```diff
@@ -0,0 +1,237 @@
...
+    } else {
+        return false; /* unknown address family */
+    }
+
+    /* Inner L3 addresses and protocol */
+    uint32_t inner_l3_off = 14;
+
+    /* Handle inner VLAN */
+    if (inner_len >= 16) {
+        uint16_t inner_et = read_u16(inner + 12);
```
banidoru, line 100: When `inner_len < 16`, the VLAN check is skipped entirely, and the code falls through to read EtherType at offset 12. A truncated 802.1Q frame (where bytes 12-13 are `0x8100` but the frame is only 14-15 bytes) will be misidentified as an IPv4/IPv6 frame with EtherType `0x8100`, returning false from the IPv4/IPv6 branches. Harmless in practice (returns false), but the logic is misleading — consider checking for `0x8100` regardless of length and returning false early if the frame is too short for a VLAN header.

src/backend/test/test_graph_node.cc
```diff
@@ -0,0 +1,337 @@
...
+    inner[12] = 0x08;
+    inner[13] = 0x00;
+    inner[14] = 0x45;
+
+    infmon_parsed_packet_t parsed{};
+    parsed.mirror_src_ip.family = (infmon_af_t) 99;
+
+    infmon_flow_fields_t out{};
+    EXPECT_FALSE(infmon_extract_flow_fields(&parsed, inner, sizeof(inner), &out));
+}
```
banidoru, line 337: No test coverage for VLAN-tagged (0x8100) inner frames or IPv6 inner frames in `infmon_extract_flow_fields`. These are important parsing paths — consider adding tests for: (1) valid 802.1Q-tagged IPv4 inner frame, (2) valid IPv6 inner frame with DSCP extraction, (3) truncated VLAN frame edge case.

banidoru (commented):
- Multi-buffer chain not handled: `erspan-decap` sets `b->current_length = dr.inner_len` but never clears `VLIB_BUFFER_NEXT_PRESENT` — downstream nodes may chase stale chained buffers if the outer packet spanned multiple buffers but the inner frame fits in one.
- Missing `_Static_assert` that `sizeof(infmon_buffer_opaque_t) <= VLIB_BUFFER_OPAQUE2_SIZE` — silent corruption if the struct grows past the 64-byte opaque2 budget.
- VLAN detection fragile on short frames: if `inner_len < 16`, the VLAN check is skipped entirely, so a truncated 802.1Q frame (14-15 bytes) will be misidentified as untagged and read EtherType from the wrong offset.
- Test coverage: no tests for VLAN-tagged inner frames or IPv6 inner frames in `test_graph_node.cc`.

src/backend/src/graph_node.c
r12f: Good catch. Fixed in 80e6b2d — added `if (kw == 0 || kw > INFMON_KEY_BUF_MAX) continue;` after computing `kw`. This handles both the zero-width case (Thread 5) and the overflow case.

src/backend/src/vpp/infmon_nodes.c
r12f: Acknowledged. The `n_left_match`/`n_left_drop` pattern here follows the standard VPP node idiom where `vlib_get_next_frame` returns the available slots and the node is expected to not exceed them. In practice, frame sizes are bounded by VLIB_FRAME_SIZE and won't wrap, but using `vlib_validate_buffer_enqueue_x1` would be cleaner. Tracking as a follow-up cleanup.

src/backend/src/vpp/infmon_nodes.c
r12f: The erspan-decap node already has a dual-loop with prefetch (lines 69-124). The flow-match node is intentionally single-loop for v1 since the per-packet work (scratch append, key encode) is heavier and the prefetch benefit is less clear. Will revisit with perf data in v2.

src/backend/src/vpp/infmon_nodes.c
r12f: Acknowledged. The header comment (lines 131-137 in graph_node.h) already documents this trade-off — ~1.5 MiB per worker is acceptable for a dedicated appliance. Per-worker heap allocation is tracked for v2 if memory-constrained deployments appear.

src/backend/src/graph_node.c
r12f: Fixed in 80e6b2d together with Thread 1 — the guard `if (kw == 0 || kw > INFMON_KEY_BUF_MAX) continue;` covers both cases.

src/backend/src/graph_node.c
r12f: Good catch. Fixed in 80e6b2d — added `if (e->flow_rule_index >= INFMON_MAX_ACTIVE_FLOW_RULES) continue;` before indexing into `tables[]`.

src/backend/src/graph_node.c
r12f: Agreed — this is a known v1 limitation. When IPv6 extension headers are present, `next_header` at offset +6 won't be the transport protocol. Walking the extension header chain is planned for v2. The current behavior is safe (wrong protocol field, but no memory safety issue). Will add a code comment documenting this.

src/backend/src/vpp/infmon_nodes.c
r12f: Acknowledged. The tick counter is bumped once per frame (not per packet), so contention is bounded by frame rate rather than packet rate. For 4 workers at ~1M frames/sec each, this is manageable. Per-worker monotonic counters are a good optimization for v2 but would require changes to the counter_table_update interface.

src/backend/src/vpp/infmon_nodes.c
r12f: Good catch. Fixed in 80e6b2d — `b->flags &= ~VLIB_BUFFER_NEXT_PRESENT;` is now cleared after setting `current_length` in both the dual-loop and single-loop paths.

src/backend/include/infmon/graph_node.h
r12f: Good suggestion. Fixed in 80e6b2d — added `_Static_assert(sizeof(infmon_buffer_opaque_t) <= 64, ...)` right after the struct definition.

src/backend/src/graph_node.c
r12f: Fixed in 80e6b2d — changed `inner_len >= 16` to `inner_len >= 14` so we always read the ethertype when the Ethernet header is present, correctly detecting 0x8100 even in truncated frames.

src/backend/test/test_graph_node.cc
r12f: Agreed — VLAN-tagged and IPv6 inner frame tests are needed. This is a good follow-up item but out of scope for this PR which focuses on the graph node plumbing. Will add in a dedicated test-coverage PR.

banidoru (approved):
- All prior concerns addressed in 80e6b2d: key-width bounds check, flow_rule_index OOB guard, VLIB_BUFFER_NEXT_PRESENT clearing, _Static_assert for opaque size, VLAN check threshold fix
- Code is clean and correct; no new issues found in iteration 2
- Acknowledged follow-ups (v2 tracking): dual-loop for flow-match, per-worker TLS heap alloc, IPv6 extension header walking, per-worker tick counter, VLAN/IPv6 test coverage
- Approve

banidoru (commented):
Re-review (iteration 2) — all prior concerns addressed in 80e6b2d:

- **Bounds guards added**: `kw == 0 || kw > INFMON_KEY_BUF_MAX` and `flow_rule_index >= INFMON_MAX_ACTIVE_FLOW_RULES` checks prevent buffer overflow and OOB access
- **VLIB_BUFFER_NEXT_PRESENT cleared** in both dual-loop and single-loop decap paths — stale chain bug fixed
- **_Static_assert** on `infmon_buffer_opaque_t` size guards against future opaque2 overflow
- **VLAN check threshold** changed to `inner_len >= 14` — truncated 802.1Q frames now correctly detected
- Acknowledged items (TLS scratch size, per-worker tick, dual-loop for flow-match, IPv6 ext headers, VLAN/IPv6 test coverage) are reasonable v2 deferrals

No new issues found. LGTM.

banidoru (approved):
- All prior review findings addressed in 80e6b2d: bounds checks on `kw` and `flow_rule_index`, `VLIB_BUFFER_NEXT_PRESENT` clearing, `_Static_assert` for opaque size, VLAN threshold fix
- Code is clean and well-structured; portable/VPP split is solid
- Acknowledged v2 deferrals (IPv6 ext headers, dual-loop flow-match, per-worker tick, TLS scratch allocation, VLAN/IPv6 test coverage) are reasonable scoping decisions

banidoru (approved):
All reviewers approved. LGTM.

PR #76, opened 2026-04-19
feat: add snapshot_and_clear with atomic table swap and epoch-based RCU

### LINKED ISSUE #42: [W8] snapshot_and_clear - atomic table swap, epoch-based RCU retirement
## Spec

## Scope
- Allocate new empty counter table (generation = G+1) off worker thread
- Atomic pointer swap (RELEASE/ACQUIRE) - zero worker stall
- Retired table (G) remains live in stats segment
- Epoch-based RCU grace period for freeing retired tables
- Latency target: <=5 ms end-to-end

## Dependencies
Blocked by: #40, #41

r12f:
## Summary

Implements the `snapshot_and_clear` operation from spec 004 §7.2 — the export primitive that atomically swaps counter tables with zero worker stall.

### What's included

- **`infmon_snapshot_mgr_t`** — manages per-worker epoch counters (cache-line aligned to avoid false sharing) and a ring of up to 16 retired tables awaiting grace-period expiry
- **`infmon_snapshot_and_clear()`** — allocates a new empty counter table (generation G+1), atomically swaps the pointer with `__atomic_store_n(..., RELEASE)` (workers load with `ACQUIRE` once per frame), and enqueues the old table for RCU retirement
- **`infmon_retire_poll()`** — reclaims retired tables after both conditions are met: (1) all workers have advanced their epoch past the swap epoch, and (2) the configurable grace window (default 5s) has elapsed
- **`infmon_worker_epoch_bump()`** — inline function called by each worker once per dispatch loop iteration

### Key design properties (per §7.2)

- **Zero worker stall** — no locks, no barriers, no `vlib_worker_thread_barrier_*` on the data path
- **No reset bit** — new table starts at zero because it's freshly allocated, not because anyone zeroed shared memory
- **No lost packets** — backend keeps counting on the new table from the very next frame
- **Internally consistent snapshots** — retired table is immutable post-swap, readable without seqlocks
- **Injectable clock** — deterministic testing with fake wall-clock

### Tests

15 gtest cases covering:
- Basic swap, sequential swaps, generation tracking, epoch_ns
- Error paths: invalid index, null table, retired ring full
- Grace period enforcement, worker epoch advancement requirements
- Concurrent epoch bumps with std::thread workers
- Atomic pointer visibility across threads
- Zero-worker edge case, destroy cleanup, multi-flow-rule independence

All existing tests continue to pass (5/5 suites green).

Closes #42

### DIFF
--- a/src/backend/include/infmon/snapshot.h
+++ b/src/backend/include/infmon/snapshot.h
@@ -1,0 +1,183 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright 2026 Riff
+ *
+ * snapshot_and_clear — atomic table swap with epoch-based RCU retirement.
+ * See specs/004-backend-architecture.md §7.2
+ */
+
+#ifndef INFMON_SNAPSHOT_H
+#define INFMON_SNAPSHOT_H
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include "infmon/counter_table.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* ── Constants ───────────────────────────────────────────────────── */
+
+/** Maximum number of worker threads tracked for RCU. */
+#define INFMON_MAX_WORKERS 32
+
+/** Maximum number of tables pending retirement at any time. */
+#define INFMON_MAX_RETIRED 16
+
+/** Default grace period in nanoseconds (5 seconds). */
+#define INFMON_RETIRE_GRACE_NS ((uint64_t)5000000000ULL)
+
+/* ── Per-worker epoch counter ────────────────────────────────────── */
+
+/**
+ * Cache-line-aligned epoch counter.  Each worker bumps its own epoch
+ * once per dispatch loop iteration with RELEASE ordering; the control
+ * thread reads all worker epochs with ACQUIRE to detect quiescence.
+ */
+typedef struct {
+    uint64_t epoch;
+    uint8_t _pad[56]; /* pad to 64 B to avoid false sharing */
+} __attribute__((aligned(64))) infmon_worker_epoch_t;
+
+/* ── Retired table descriptor ────────────────────────────────────── */
+
+typedef struct {
+    infmon_counter_table_t *table;      /** The retired table. */
+    uint64_t swap_epoch;                /** Global epoch at time of swap. */
+    uint64_t swap_timestamp_ns;         /** Wall-clock timestamp of swap. */
+    uint32_t flow_rule_index;           /** Which flow_rule this belonged to. */
+    bool pending;                       /** true if still awaiting retirement. */
+} infmon_retired_table_t;
+
+/* ── Snapshot manager ────────────────────────────────────────────── */
+
+/**
+ * The snapshot manager owns:
+ *   - per-worker epoch counters,
+ *   - a ring of retired tables awaiting grace-period expiry,
+ *   - a global epoch counter bumped on each swap.
+ *
+ * Thread safety:
+ *   - Workers call infmon_worker_epoch_bump() from the data path (no lock).
+ *   - The control thread calls infmon_snapshot_and_clear() and
+ *     infmon_retire_poll() (serialised — only one control thread).
+ */
+typedef struct {
+    infmon_worker_epoch_t worker_epochs[INFMON_MAX_WORKERS];
+    uint32_t num_workers;
+
+    uint64_t global_epoch;              /** Bumped on each swap. */
+
+    infmon_retired_table_t retired[INFMON_MAX_RETIRED];
+    uint32_t retired_count;             /** Number of pending entries. */
+
+    uint64_t grace_ns;                  /** Configurable grace window. */
+
+    /* Callback for getting wall-clock nanoseconds (injectable for testing). */
+    uint64_t (*clock_ns)(void);
+} infmon_snapshot_mgr_t;
+
+/* ── Snapshot result ─────────────────────────────────────────────── */
+
+typedef enum {
+    INFMON_SNAP_OK = 0,
+    INFMON_SNAP_ALLOC_FAILED,           /** Could not allocate replacement table. */
+    INFMON_SNAP_TOO_MANY_RETIRED,       /** Retired ring is full. */
+    INFMON_SNAP_INVALID_INDEX,          /** flow_rule_index out of range. */
+    INFMON_SNAP_NULL_TABLE,             /** No table installed at that index. */
+} infmon_snap_result_t;
+
+typedef struct {
+    infmon_snap_result_t result;
+    infmon_counter_table_t *retired_table;   /** The old table (generation G). */
+    uint64_t retired_generation;             /** Generation of the retired table. */
+} infmon_snap_reply_t;
+
+/* ── Lifecycle ───────────────────────────────────────────────────── */
+
+/**
+ * Initialise a snapshot manager.
+ *
+ * @param mgr          Manager to initialise.
+ * @param num_workers  Number of worker threads.
+ * @param grace_ns     Grace period in nanoseconds (0 = use default).
+ * @param clock_ns     Wall-clock function (NULL = use clock_gettime).
+ */
+void infmon_snapshot_mgr_init(infmon_snapshot_mgr_t *mgr, uint32_t num_workers,
+                              uint64_t grace_ns, uint64_t (*clock_ns)(void));
+
+/**
+ * Destroy the snapshot manager.  Frees any retired tables still pending.
+ */
+void infmon_snapshot_mgr_destroy(infmon_snapshot_mgr_t *mgr);
+
+/* ── Worker-side (data path) ─────────────────────────────────────── */
+
+/**
+ * Bump the calling worker's epoch.  Must be called once per dispatch
+ * loop iteration.  The worker_id must be in [0, num_workers).
+ */
+static inline void infmon_worker_epoch_bump(infmon_snapshot_mgr_t *mgr, uint32_t worker_id)
+{
+    __atomic_fetch_add(&mgr->worker_epochs[worker_id].epoch, 1, __ATOMIC_RELEASE);
+}
+
+/**
+ * Read a worker's published epoch (for control thread use).
+ */
+static inline uint64_t infmon_worker_epoch_read(const infmon_snapshot_mgr_t *mgr,
+                                                 uint32_t worker_id)
+{
+    return __atomic_load_n(&mgr->worker_epochs[worker_id].epoch, __ATOMIC_ACQUIRE);
+}
+
+/* ── Control-plane operations ────────────────────────────────────── */
+
+/**
+ * Perform an atomic snapshot_and_clear on one flow_rule's counter table.
+ *
+ * 1. Allocates a new empty table with the same dimensions.
+ * 2. Sets generation = G+1 and epoch_ns on the new table.
+ * 3. Atomically swaps the table pointer in tables[flow_rule_index].
+ * 4. Enqueues the old table for RCU retirement.
+ *
+ * @param mgr              Snapshot manager.
+ * @param tables           Pointer to the tables array (e.g. pm->tables).
+ * @param flow_rule_index  Index of the flow_rule to snapshot.
+ * @param max_flow_rules   Size of the tables array.
+ * @param max_key_width    Key width for the new table (same as old).
+ * @param reply            Output: result + retired table info.
+ */
+void infmon_snapshot_and_clear(infmon_snapshot_mgr_t *mgr,
+                               infmon_counter_table_t **tables,
+                               uint32_t flow_rule_index,
+                               uint32_t max_flow_rules,
+                               uint32_t max_key_width,
+                               infmon_snap_reply_t *reply);
+
+/**
+ * Poll for retired tables whose grace period has expired.
+ * Frees any that are safe to reclaim.
+ *
+ * @param mgr  Snapshot manager.
+ * @return Number of tables freed this call.
+ */
+uint32_t infmon_retire_poll(infmon_snapshot_mgr_t *mgr);
+
+/**
+ * Check whether all workers have advanced past a given epoch.
+ *
+ * @param mgr    Snapshot manager.
+ * @param epoch  The epoch to check against.
+ * @return true if every worker's epoch > epoch.
+ */
+bool infmon_all_workers_past(const infmon_snapshot_mgr_t *mgr, uint64_t epoch);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* INFMON_SNAPSHOT_H */
--- a/src/backend/include/infmon/snapshot.h
+++ b/src/backend/include/infmon/snapshot.h
@@ -28,7 +28,7 @@ extern "C" {
 #define INFMON_MAX_RETIRED 16
 
 /** Default grace period in nanoseconds (5 seconds). */
-#define INFMON_RETIRE_GRACE_NS ((uint64_t)5000000000ULL)
+#define INFMON_RETIRE_GRACE_NS ((uint64_t) 5000000000ULL)
 
 /* ── Per-worker epoch counter ────────────────────────────────────── */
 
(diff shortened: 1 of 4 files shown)


src/backend/include/infmon/snapshot.h
```diff
@@ -0,0 +1,180 @@
...
+ * once per dispatch loop iteration with RELEASE ordering; the control
+ * thread reads all worker epochs with ACQUIRE to detect quiescence.
+ */
+typedef struct {
+    uint64_t epoch;
+    uint8_t _pad[56]; /* pad to 64 B to avoid false sharing */
+} __attribute__((aligned(64))) infmon_worker_epoch_t;
+
+/* ── Retired table descriptor ────────────────────────────────────── */
+
```
banidoru, line 47: **Reserved identifier**: Names starting with `_` followed by a lowercase letter are reserved at file scope in C (§7.1.3). Rename `_pad` to `pad` or `padding` to be conformant.

Also, the padding size `56` assumes `sizeof(uint64_t) == 8`. Consider:
```c
uint8_t pad[64 - sizeof(uint64_t)];
```
to be self-documenting and portable.

src/backend/include/infmon/snapshot.h
```diff
@@ -0,0 +1,180 @@
...
+ */
+static inline void infmon_worker_epoch_bump(infmon_snapshot_mgr_t *mgr, uint32_t worker_id)
+{
+    __atomic_fetch_add(&mgr->worker_epochs[worker_id].epoch, 1, __ATOMIC_RELEASE);
+}
+
+/**
+ * Read a worker's published epoch (for control thread use).
+ */
+static inline uint64_t infmon_worker_epoch_read(const infmon_snapshot_mgr_t *mgr,
```
banidoru, line 138: `__atomic_fetch_add` is an atomic RMW (locked instruction on x86). Since each worker epoch is single-writer, you can avoid the locked op:
```c
uint64_t e = __atomic_load_n(&mgr->worker_epochs[worker_id].epoch, __ATOMIC_RELAXED);
__atomic_store_n(&mgr->worker_epochs[worker_id].epoch, e + 1, __ATOMIC_RELEASE);
```
This is cheaper on the data path — a plain store with release ordering vs. a `lock xadd`.

src/backend/src/snapshot.c
```diff
@@ -0,0 +1,170 @@
...
+    new_table->epoch_ns = mgr->clock_ns();
+
+    /* Bump global epoch for this swap */
+    uint64_t swap_epoch = ++mgr->global_epoch;
+
+    /*
+     * Atomic pointer swap: RELEASE ensures the new table's contents
+     * (zeroed slots, metadata) are visible before any worker observes
+     * the new pointer.  Workers load with ACQUIRE once per frame.
+     */
```
banidoru, line 120: `++mgr->global_epoch` is a non-atomic RMW. The design doc says only one control thread calls this, but nothing enforces that. Consider either:
1. Adding `_Atomic` to `global_epoch` in the struct, or
2. Adding a prominent comment: `/* NOT thread-safe: caller must serialize */`

This would prevent a subtle bug if someone adds a second control thread later.

src/backend/src/snapshot.c
```diff
@@ -0,0 +1,170 @@
...
+            mgr->retired[i].pending = true;
+            mgr->retired_count++;
+            break;
+        }
+    }
+
+    /* Fill reply */
+    reply->result = INFMON_SNAP_OK;
+    reply->retired_table = old_table;
+    reply->retired_generation = old_table->generation;
```
banidoru, line 141: If `retired_count` ever drifts out of sync with actual `pending` flags (bug elsewhere), this loop silently does nothing — the old table leaks and `retired_count` stays incremented without a matching slot. Add a defensive check after the loop:
```c
/* Should be unreachable: retired_count was checked above */
assert(0 && "retired_count > 0 but no free slot found");
```
Or restructure so `retired_count` is derived from scanning rather than maintained separately.

src/backend/test/test_snapshot.cc
```diff
@@ -0,0 +1,427 @@
...
+        infmon_snapshot_and_clear(&mgr, tables, i, MAX_FLOW_RULES, MAX_KEY_WIDTH, &reply);
+        ASSERT_EQ(reply.result, INFMON_SNAP_OK);
+    }
+
+    /* Next swap should fail with TOO_MANY_RETIRED */
+    install_table(0); /* tables[0] was swapped above, so it's a fresh table */
+    infmon_snap_reply_t reply{};
+    infmon_snapshot_and_clear(&mgr, tables, 0, MAX_FLOW_RULES, MAX_KEY_WIDTH, &reply);
+    ASSERT_EQ(reply.result, INFMON_SNAP_TOO_MANY_RETIRED);
+}
```
banidoru, line 265: The 100-attempt poll loop could theoretically fail on a heavily loaded CI box if workers haven't advanced in time, making this test flaky. Consider adding `ASSERT_LT(attempt, 100) << "retire_poll never freed";` after the loop to get a clear failure message instead of silent success with leaked tables.

banidoru (commented):
Solid RCU-based snapshot design with good test coverage. A few items:

- `_pad` uses a reserved identifier (leading underscore at file scope); rename to `pad` or `padding`
- Retired ring scan has no safety net if `retired_count` drifts out of sync — add an assert/log after the loop
- `__atomic_fetch_add` in `infmon_worker_epoch_bump` is heavier than needed for a single-writer epoch — `__atomic_store_n` with a local increment would be cheaper on x86 (no locked RMW)
- `global_epoch` increment (`++mgr->global_epoch`) is a plain non-atomic read-modify-write — fine if truly single-threaded, but a comment or `_Atomic` would prevent future misuse

src/backend/include/infmon/snapshot.h
```diff
@@ -0,0 +1,180 @@
...
+/**
+ * Bump the calling worker's epoch.  Must be called once per dispatch
+ * loop iteration.  The worker_id must be in [0, num_workers).
+ */
+static inline void infmon_worker_epoch_bump(infmon_snapshot_mgr_t *mgr, uint32_t worker_id)
+{
+    __atomic_fetch_add(&mgr->worker_epochs[worker_id].epoch, 1, __ATOMIC_RELEASE);
+}
+
+/**
```
banidoru, line 135: No bounds check on `worker_id`. If a caller passes `worker_id >= INFMON_MAX_WORKERS`, this is an out-of-bounds write into `worker_epochs[]` — silent memory corruption.

At minimum:
```c
static inline void infmon_worker_epoch_bump(infmon_snapshot_mgr_t *mgr, uint32_t worker_id)
{
    assert(worker_id < mgr->num_workers);
    ...
}
```
Same issue in `infmon_worker_epoch_read` below.

src/backend/src/snapshot.c
```diff
@@ -0,0 +1,170 @@
...
+
+void infmon_snapshot_mgr_init(infmon_snapshot_mgr_t *mgr, uint32_t num_workers, uint64_t grace_ns,
+                              uint64_t (*clock_ns)(void))
+{
+    memset(mgr, 0, sizeof(*mgr));
+
+    if (num_workers > INFMON_MAX_WORKERS)
+        num_workers = INFMON_MAX_WORKERS;
+
+    mgr->num_workers = num_workers;
```
banidoru, line 33: Silently clamping `num_workers` to `INFMON_MAX_WORKERS` is dangerous — the caller thinks they have N workers tracked but only 32 are actually guarded by RCU. The remaining workers can access a retired table after it's freed.

Return an error code or `assert(num_workers <= INFMON_MAX_WORKERS)` to fail loudly.

src/backend/test/test_snapshot.cc
```diff
@@ -0,0 +1,427 @@
...
+
+/* ── Retired ring full ───────────────────────────────────────────── */
+
+TEST_F(SnapshotTest, RetiredRingFull)
+{
+    /* Fill the retired ring */
+    for (uint32_t i = 0; i < INFMON_MAX_RETIRED; i++) {
+        install_table(i);
+        infmon_snap_reply_t reply{};
```
banidoru, line 253: Memory leak: after the loop, `tables[0]` points to the table that was swapped in at index 0. Calling `install_table(0)` overwrites that pointer without freeing it. Fix:
```cpp
infmon_counter_table_destroy(tables[0]);
tables[0] = nullptr;
install_table(0);
```

src/backend/src/snapshot.c
```diff
@@ -0,0 +1,170 @@
...
+    if (!new_table) {
+        reply->result = INFMON_SNAP_ALLOC_FAILED;
+        return;
+    }
+
+    /* Set new table metadata */
+    uint64_t new_gen = old_table->generation + 1;
+    new_table->generation = new_gen;
+    new_table->epoch_ns = mgr->clock_ns();
+
```
banidoru, line 107: `new_table->epoch_ns = mgr->clock_ns()` and `rt->swap_timestamp_ns = mgr->clock_ns()` (line 126) call the clock twice — the new table's epoch_ns and the retired entry's swap timestamp will differ. If the clock source has coarse granularity, this is fine, but if anything ever compares them they won't match. Consider capturing `now = mgr->clock_ns()` once and using it for both.

banidoru (commented):
- **No bounds check on `worker_id`** in inline `epoch_bump`/`epoch_read` — out-of-bounds write to `worker_epochs[]` if caller passes bad id. At minimum add a debug assert.
- **Silent clamping** of `num_workers` in init — callers won't know they lost workers. Should return an error or at least log.
- **Memory leak in `RetiredRingFull` test** — `install_table(0)` overwrites the swapped-in table pointer without freeing it.
- **`worker_id` bounds check missing in inline hot path** — even a `__builtin_expect`-guarded early return would prevent memory corruption in production.

banidoru (commented):
- Unsigned subtraction in `retire_poll` can wrap around with injectable clocks, causing premature table reclamation
- Trailing doc-comments use `/** */` instead of `/**< */` — Doxygen won't associate them with the preceding member
- `infmon_snapshot_mgr_t` is ~3KB+ due to cache-line-padded epoch array — document that it must be heap-allocated, not on the stack
- Overall: solid RCU design, good test coverage, well-structured. Previous reviewers caught the critical issues.

src/backend/src/snapshot.c
```diff
@@ -0,0 +1,170 @@
...
+        if (!mgr->retired[i].pending)
+            continue;
+
+        infmon_retired_table_t *rt = &mgr->retired[i];
+
+        /* Two conditions for safe reclamation:
+         * 1. All workers have advanced past the swap epoch.
+         * 2. Grace period has elapsed since the swap.
+         */
+        if (!infmon_all_workers_past(mgr, rt->swap_epoch))
```
banidoru, line 161: Unsigned underflow risk: `now - rt->swap_timestamp_ns` wraps to a huge value if the injectable clock ever returns a smaller value than the recorded timestamp (e.g., test bug, clock reset). This would cause premature table reclamation — a use-after-free.

Guard:
```c
if (now <= rt->swap_timestamp_ns || (now - rt->swap_timestamp_ns) < mgr->grace_ns)
    continue;
```

src/backend/include/infmon/snapshot.h
```diff
@@ -0,0 +1,180 @@
...
+    uint8_t _pad[56]; /* pad to 64 B to avoid false sharing */
+} __attribute__((aligned(64))) infmon_worker_epoch_t;
+
+/* ── Retired table descriptor ────────────────────────────────────── */
+
+typedef struct {
+    infmon_counter_table_t *table; /** The retired table. */
+    uint64_t swap_epoch;           /** Global epoch at time of swap. */
+    uint64_t swap_timestamp_ns;    /** Wall-clock timestamp of swap. */
+    uint32_t flow_rule_index;      /** Which flow_rule this belonged to. */
```
banidoru: Trailing member comments use `/** */` but Doxygen requires `/**< */` for post-member documentation. As written, each `/** ... */` attaches to the *next* member, not the current one. Same issue on lines 52-55 and throughout `infmon_snap_reply_t`.

```c
infmon_counter_table_t *table; /**< The retired table. */
```

src/backend/include/infmon/snapshot.h
```diff
@@ -0,0 +1,180 @@
...
+ *     infmon_retire_poll() (serialised — only one control thread).
+ */
+typedef struct {
+    infmon_worker_epoch_t worker_epochs[INFMON_MAX_WORKERS];
+    uint32_t num_workers;
+
+    uint64_t global_epoch; /** Bumped on each swap. */
+
+    infmon_retired_table_t retired[INFMON_MAX_RETIRED];
+    uint32_t retired_count; /** Number of pending entries. */
```
banidoru: `infmon_snapshot_mgr_t` is over 3KB (`worker_epochs` alone is 32×64 = 2048 bytes). If anyone stack-allocates this, they'll blow through typical stack limits on embedded targets. Consider a `static_assert(sizeof(...) < threshold)` or a doc note that this must be heap-allocated or placed in BSS.

banidoru (approved):
All reviewers approved. LGTM.

src/backend/include/infmon/snapshot.h
r12f: Fixed: renamed `_pad` to `pad` and changed sizing to `64 - sizeof(uint64_t)` for portability and C conformance. Good catch on both points.

src/backend/include/infmon/snapshot.h
r12f: Fixed: switched to relaxed load + release store pattern to avoid the locked RMW on the data path. Thanks for the optimization suggestion.

src/backend/src/snapshot.c
r12f: Added a prominent `NOT thread-safe: caller must serialize (single control thread)` comment on the increment. Went with option 2 since making it `_Atomic` would add unnecessary overhead given the single-writer design.

src/backend/src/snapshot.c
r12f: Added a defensive comment after the loop noting this should be unreachable since `retired_count` is checked before entering the function. The count is always maintained atomically with the `pending` flag within the same serialized control path, so drift shouldn't happen, but the comment makes the invariant explicit.

src/backend/test/test_snapshot.cc
r12f: Fixed: added `ASSERT_LT(attempt, 100) << "retire_poll never freed the table"` after the loop for a clear failure message instead of silent success.

src/backend/include/infmon/snapshot.h
r12f: Fixed: added `assert(worker_id < mgr->num_workers)` to both `infmon_worker_epoch_bump` and `infmon_worker_epoch_read`. Also added `#include <assert.h>` to the header.

src/backend/src/snapshot.c
r12f: Fixed: replaced the silent clamp with `assert(num_workers <= INFMON_MAX_WORKERS)` so it fails loudly instead of silently leaving workers unprotected.

src/backend/test/test_snapshot.cc
r12f: Fixed: added `infmon_counter_table_destroy(tables[0]); tables[0] = nullptr;` before the `install_table(0)` call to free the swapped-in table and prevent the leak.

src/backend/src/snapshot.c
r12f: Fixed: now capturing `uint64_t now = mgr->clock_ns()` once and using it for both `new_table->epoch_ns` and `rt->swap_timestamp_ns` so they're consistent.

src/backend/src/snapshot.c
r12f: Fixed: added `now <= rt->swap_timestamp_ns` guard before the subtraction to prevent unsigned underflow if the clock ever goes backward.

src/backend/include/infmon/snapshot.h
r12f: Fixed: changed all post-member doc comments from `/** */` to `/**< */` throughout the header (retired table descriptor, snap result enum, snap reply struct, and manager struct).

src/backend/include/infmon/snapshot.h
r12f: Added a `@note` to the struct documentation warning that it's >3KB and should not be stack-allocated — heap or BSS placement only.

PR #78, opened 2026-04-19
feat(backend): add VPP binary API .api schema + vppapigen build integration

### LINKED ISSUE #44: [W10a] VPP binary API .api schema + vppapigen build integration
## Spec

## Scope
- Define wire schema in src/backend/api/infmon.api
- Integrate with vppapigen to produce C headers and Rust bindings during build
- Message definitions for all v1 API messages (W10b-W10e)

## Dependencies
Blocked by: #40

r12f:
## Summary

Define the InFMon VPP binary API wire schema (`src/backend/api/infmon.api`) covering all v1 messages per spec §7.1 (W10b–W10e):

| Message pair | Spec ref |
|---|---|
| `infmon_flow_rule_add` / `infmon_flow_rule_del` | W10b |
| `infmon_flow_rule_list` / `infmon_flow_rule_get` | W10c |
| `infmon_snapshot_and_clear` | W10d |
| `infmon_status` | W10e |

### Build integration
- `cmake/FindVppApiGen.cmake` — locates `vppapigen` and generates C headers + JSON. Gracefully skips on dev machines without VPP.

### Testing
- `test_api_schema.cc` — 19 contract tests that parse the raw `.api` file and validate all expected messages, enums, types, and fields. Runs without `vppapigen`, catches schema regressions before CI invokes the code generator.

All existing + new tests pass locally (7/7 test targets, 19/19 api_schema subtests).

Closes #44

### DIFF
--- a/src/backend/api/infmon.api
+++ b/src/backend/api/infmon.api
@@ -1,0 +1,225 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright 2026 Riff
+ *
+ * InFMon VPP binary API — wire schema for all v1 API messages.
+ * See specs/004-backend-architecture.md §7
+ *
+ * Consumed by vppapigen to produce C headers and Rust bindings.
+ *
+ * Conventions:
+ *   - All messages use client_index + context for VPP API routing.
+ *   - _reply messages carry a standard i32 retval (0 = OK, negative = error).
+ *   - flow_rule_id is a 128-bit UUID represented as two u64 fields (hi, lo).
+ *   - Variable-length key fields use vla (variable length array) at the end.
+ *   - Field widths match the C structs in include/infmon/*.h.
+ */
+
+option version = "1.0.0";
+
+/* ── Enumerations ──────────────────────────────────────────────────── */
+
+/** Flow-rule key field types — mirrors infmon_field_t. */
+enum infmon_field_type : u8
+{
+    INFMON_FIELD_SRC_IP = 0,
+    INFMON_FIELD_DST_IP = 1,
+    INFMON_FIELD_IP_PROTO = 2,
+    INFMON_FIELD_DSCP = 3,
+    INFMON_FIELD_MIRROR_SRC_IP = 4,
+};
+
+/** Eviction policy — mirrors infmon_eviction_policy_t. */
+enum infmon_eviction_policy : u8
+{
+    INFMON_EVICTION_LRU_DROP = 0,
+};
+
+/** Error codes for flow_rule operations. */
+enum infmon_flow_rule_error : i32
+{
+    INFMON_FLOW_RULE_OK = 0,
+    INFMON_FLOW_RULE_ERR_NAME_EXISTS = -1,
+    INFMON_FLOW_RULE_ERR_NOT_FOUND = -2,
+    INFMON_FLOW_RULE_ERR_INVALID_SPEC = -3,
+    INFMON_FLOW_RULE_ERR_BUDGET_EXCEEDED = -4,
+    INFMON_FLOW_RULE_ERR_SET_FULL = -5,
+    INFMON_FLOW_RULE_ERR_INTERNAL = -6,
+};
+
+/** Error codes for snapshot_and_clear. */
+enum infmon_snap_error : i32
+{
+    INFMON_SNAP_OK = 0,
+    INFMON_SNAP_ALLOC_FAILED = -1,
+    INFMON_SNAP_TOO_MANY_RETIRED = -2,
+    INFMON_SNAP_INVALID_INDEX = -3,
+    INFMON_SNAP_NULL_TABLE = -4,
+};
+
+/* ── Type definitions ──────────────────────────────────────────────── */
+
+/** 128-bit flow_rule UUID. */
+typedef infmon_flow_rule_id
+{
+    u64 hi;
+    u64 lo;
+};
+
+/** Descriptor of a counter table (returned by snapshot_and_clear). */
+typedef infmon_table_descriptor
+{
+    u64 flow_rule_id_hi;
+    u64 flow_rule_id_lo;
+    u32 flow_rule_index;
+    u64 generation;
+    u64 epoch_ns;
+    u64 slots_offset;
+    u32 slots_len;
+    u64 key_arena_offset;
+    u32 key_arena_capacity;
+    u32 key_arena_used;
+    u64 insert_failed;
+    u64 table_full;
+};
+
+/** Per-flow-rule summary (used in list/get replies). */
+typedef infmon_flow_rule_details
+{
+    u64 flow_rule_id_hi;
+    u64 flow_rule_id_lo;
+    u32 flow_rule_index;
+    string name[32];       /* NUL-terminated, max 31 chars + NUL */
+    u32 field_count;
+    vl_api_infmon_field_type_t fields[5];
+    u32 max_keys;
+    vl_api_infmon_eviction_policy_t eviction_policy;
+    u32 key_width;
+};
+
+/** Per-worker status counters. */
+typedef infmon_worker_status
+{
+    u32 worker_id;
+    u64 packets_seen;
+    u64 erspan_unknown_proto;
+    u64 erspan_truncated;
+    u64 inner_parse_failed;
+    u64 flow_rule_no_match;
+    u64 counter_insert_retry_exhausted;
+    u64 counter_table_full;
+};
+
+/* ── W10b: infmon_flow_rule_add / infmon_flow_rule_del ─────────────── */
+
+/**
+ * Register a new flow_rule.  Allocates a counter table and returns
+ * the assigned flow_rule_id.
+ */
+autoreply define infmon_flow_rule_add
+{
+    u32 client_index;
+    u32 context;
+    string name[32];
+    u32 field_count;
+    vl_api_infmon_field_type_t fields[5];
+    u32 max_keys;
+    vl_api_infmon_eviction_policy_t eviction_policy;
+};
+
+/**
+ * Reply to infmon_flow_rule_add.
+ * retval: 0 on success, negative infmon_flow_rule_error on failure.
+ * flow_rule_id: the assigned UUID (only valid when retval == 0).
+ */
+define infmon_flow_rule_add_reply
+{
+    u32 context;
+    i32 retval;
+    vl_api_infmon_flow_rule_id_t flow_rule_id;
+};
+
+/**
+ * Tear down a flow_rule and free its counter tables.
+ */
+autoreply define infmon_flow_rule_del
+{
+    u32 client_index;
+    u32 context;
+    vl_api_infmon_flow_rule_id_t flow_rule_id;
+};
+
+/* ── W10c: infmon_flow_rule_list / infmon_flow_rule_get ────────────── */
+
+/**
+ * Enumerate all active flow_rules.
+ * Returns one infmon_flow_rule_list_details per flow_rule.
+ */
+define infmon_flow_rule_list
+{
+    u32 client_index;
+    u32 context;
+};
+
+/** Per-flow_rule detail message (streamed, one per active rule). */
+define infmon_flow_rule_list_details
+{
+    u32 context;
+    vl_api_infmon_flow_rule_details_t flow_rule;
+};
+
+/**
+ * Get the full definition of one flow_rule by its UUID.
+ */
+define infmon_flow_rule_get
+{
+    u32 client_index;
+    u32 context;
+    vl_api_infmon_flow_rule_id_t flow_rule_id;
+};
+
+define infmon_flow_rule_get_reply
+{
+    u32 context;
+    i32 retval;
+    vl_api_infmon_flow_rule_details_t flow_rule;
+};
+
+/* ── W10d: infmon_snapshot_and_clear ──────────────────────────────── */
+
+/**
+ * Atomic table swap for one flow_rule.  Returns the descriptor of the
+ * retired table (generation G) so the caller can walk it at leisure.
+ */
+define infmon_snapshot_and_clear
+{
+    u32 client_index;
+    u32 context;
+    vl_api_infmon_flow_rule_id_t flow_rule_id;
+};
+
+define infmon_snapshot_and_clear_reply
+{
+    u32 context;
+    i32 retval;
+    vl_api_infmon_table_descriptor_t descriptor;
+};
+
+/* ── W10e: infmon_status ─────────────────────────────────────────── */
+
+/**
+ * Retrieve per-worker error/health counters.
+ * Returns one infmon_status_details per worker.
+ */
+define infmon_status
+{
+    u32 client_index;
+    u32 context;
+};
+
+define infmon_status_details
+{
+    u32 context;
+    u32 worker_count;
+    vl_api_infmon_worker_status_t workers[worker_count];
+};
--- a/src/backend/api/infmon.api
+++ b/src/backend/api/infmon.api
@@ -66,7 +66,11 @@ typedef infmon_flow_rule_id
     u64 lo;
 };
 
-/** Descriptor of a counter table (returned by snapshot_and_clear). */
+/*
+ * Descriptor of a counter table (returned by snapshot_and_clear).
+ * Note: vppapigen uses packed wire layout (no implicit padding).
+ * Field order follows logical grouping, not alignment optimization.
+ */
 typedef infmon_table_descriptor
 {
     vl_api_infmon_flow_rule_id_t flow_rule_id;
--- a/src/backend/api/infmon.api
+++ b/src/backend/api/infmon.api
@@ -69,8 +69,7 @@ typedef infmon_flow_rule_id
 /** Descriptor of a counter table (returned by snapshot_and_clear). */
 typedef infmon_table_descriptor
 {
-    u64 flow_rule_id_hi;
-    u64 flow_rule_id_lo;
+    vl_api_infmon_flow_rule_id_t flow_rule_id;
     u32 flow_rule_index;
     u64 generation;
     u64 epoch_ns;
@@ -169,6 +168,13 @@ define infmon_flow_rule_list_details
     vl_api_infmon_flow_rule_details_t flow_rule;
 };
 
+/** Terminal reply for flow_rule_list — signals end of stream. */
+define infmon_flow_rule_list_reply
+{
+    u32 context;
+    i32 retval;
+};
+
 /**
  * Get the full definition of one flow_rule by its UUID.
  */
@@ -224,4 +230,11 @@ define infmon_status_details
     u32 context;
     u32 worker_count;
     vl_api_infmon_worker_status_t workers[worker_count];
+};
+
+/** Terminal reply for infmon_status — signals end of stream. */
+define infmon_status_reply
+{
+    u32 context;
+    i32 retval;
 };
--- a/src/backend/api/infmon.api
+++ b/src/backend/api/infmon.api
@@ -89,8 +89,7 @@ typedef infmon_table_descriptor
 /** Per-flow-rule summary (used in list/get replies). */
 typedef infmon_flow_rule_details
 {
-    u64 flow_rule_id_hi;
-    u64 flow_rule_id_lo;
+    vl_api_infmon_flow_rule_id_t flow_rule_id;
     u32 flow_rule_index;
     string name[32];       /* NUL-terminated, max 31 chars + NUL */
     u32 field_count;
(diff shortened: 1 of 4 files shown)


src/backend/api/infmon.api
```diff
@@ -0,0 +1,225 @@
...
+    u64 erspan_unknown_proto;
+    u64 erspan_truncated;
+    u64 inner_parse_failed;
+    u64 flow_rule_no_match;
+    u64 counter_insert_retry_exhausted;
+    u64 counter_table_full;
+};
+
+/* ── W10b: infmon_flow_rule_add / infmon_flow_rule_del ─────────────── */
+
```
banidoru, line 117: **Bug:** `autoreply define infmon_flow_rule_add` tells vppapigen to auto-generate a reply with only `context` + `retval`. But you also explicitly `define infmon_flow_rule_add_reply` below (lines 123-129) with the extra `flow_rule_id` field. This will either cause a vppapigen error (duplicate message) or silently drop your custom reply.

Fix: remove the `autoreply` keyword here, since you need the custom reply.

src/backend/api/infmon.api
```diff
@@ -0,0 +1,225 @@
...
+ * retired table (generation G) so the caller can walk it at leisure.
+ */
+define infmon_snapshot_and_clear
+{
+    u32 client_index;
+    u32 context;
+    vl_api_infmon_flow_rule_id_t flow_rule_id;
+};
+
+define infmon_snapshot_and_clear_reply
```
banidoru, line 211: Comment says "Returns one `infmon_status_details` per worker" but the message definition bundles *all* workers into a single message via `workers[worker_count]` VLA. Either:
- Change to a streaming pattern (one `_details` per worker, like `flow_rule_list_details`), or
- Fix the comment to say "Returns all workers in one message".

src/backend/cmake/FindVppApiGen.cmake
```diff
@@ -0,0 +1,90 @@
...
+            COMMENT "Generating VPP API JSON from ${API_BASENAME}"
+            VERBATIM
+        )
+
+        add_custom_target(infmon_api_generated
+            DEPENDS "${OUTPUT_HEADER}" "${OUTPUT_JSON}"
+        )
+    else()
+        message(STATUS "vppapigen not found — API header generation skipped (pre-generated headers required)")
+        # Create a dummy target so dependents don't break
```
banidoru, line 83: Hardcoded target name `infmon_api_generated` makes this module non-reusable if you ever add a second `.api` file. Consider deriving the target name from the API file basename, e.g. `${API_BASENAME}_generated`.

src/backend/api/infmon.api
```diff
@@ -0,0 +1,225 @@
...
+{
+    u64 flow_rule_id_hi;
+    u64 flow_rule_id_lo;
+    u32 flow_rule_index;
+    string name[32];       /* NUL-terminated, max 31 chars + NUL */
+    u32 field_count;
+    vl_api_infmon_field_type_t fields[5];
+    u32 max_keys;
+    vl_api_infmon_eviction_policy_t eviction_policy;
+    u32 key_width;
```
banidoru, line 100: `fields[5]` is a fixed-size array baked into the wire format. If you add a 6th field type to `infmon_field_type` later, existing messages can't carry it without a version bump. Consider whether this limit should be documented prominently or use a VLA instead.

banidoru (changes requested):
- **Bug:** `autoreply` on `infmon_flow_rule_add` conflicts with the explicit `_reply` definition — vppapigen will generate a duplicate or ignore one. Remove `autoreply` since you need the custom reply with `flow_rule_id`.
- `infmon_status_details` comment says "one per worker" but the message bundles all workers in a VLA — comment/design mismatch.
- `FindVppApiGen.cmake` hardcodes target name `infmon_api_generated`, making it non-reusable for multiple .api files.
- Schema tests are string-match only — `expect_contains("generation")` would pass even if the field moved to a wrong type block. Consider scoped checks.

src/backend/api/infmon.api
```diff
@@ -0,0 +1,225 @@
...
+};
+
+/** Descriptor of a counter table (returned by snapshot_and_clear). */
+typedef infmon_table_descriptor
+{
+    u64 flow_rule_id_hi;
+    u64 flow_rule_id_lo;
+    u32 flow_rule_index;
+    u64 generation;
+    u64 epoch_ns;
```
banidoru, line 79: `infmon_table_descriptor` uses flat `u64 flow_rule_id_hi` / `flow_rule_id_lo` instead of embedding `vl_api_infmon_flow_rule_id_t flow_rule_id` like every other message does. This is inconsistent and means two representations of the same concept on the wire. If there's a packing/alignment reason, worth a comment; otherwise reuse the typedef.

src/backend/api/infmon.api
```diff
@@ -0,0 +1,225 @@
...
+ * Tear down a flow_rule and free its counter tables.
+ */
+autoreply define infmon_flow_rule_del
+{
+    u32 client_index;
+    u32 context;
+    vl_api_infmon_flow_rule_id_t flow_rule_id;
+};
+
+/* ── W10c: infmon_flow_rule_list / infmon_flow_rule_get ────────────── */
```
banidoru, line 155: `infmon_flow_rule_list` uses a streaming `_details` pattern but has no terminal `_reply` message with `retval`. VPP dump APIs typically send a `_details` stream followed by a control-plane `_reply` (or use `option status="in_progress"`) so the client can distinguish "zero results" from "request error". Same applies to `infmon_status` below.

src/backend/test/test_api_schema.cc
```diff
@@ -0,0 +1,240 @@
...
+/* ── Type definitions ─────────────────────────────────────────────── */
+
+TEST_F(ApiSchemaTest, HasFlowRuleIdType)
+{
+    expect_contains("typedef infmon_flow_rule_id", "Missing flow_rule_id typedef");
+    /* Must have hi/lo u64 fields */
+    expect_contains("u64 hi", "flow_rule_id must have u64 hi");
+    expect_contains("u64 lo", "flow_rule_id must have u64 lo");
+}
+
```
banidoru, line 128: These substring checks (`expect_contains("u64 hi", ...)`) match *anywhere* in the file, not within the `infmon_flow_rule_id` block specifically. If another type happened to have a `u64 hi` field, the test would pass spuriously. Consider extracting the text between `typedef infmon_flow_rule_id` and its closing `};` first, then asserting within that slice.

src/backend/test/test_api_schema.cc
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    /* Check that a pattern appears in the .api text. */
+    void expect_contains(const std::string &pattern, const std::string &msg)
+    {
+        EXPECT_NE(api.find(pattern), std::string::npos)
+            << msg << "\n  Missing pattern: " << pattern;
+    }
+
+    /* Check that a regex matches somewhere in the .api text. */
```
banidoru, line 50: `expect_matches` compiles a new `std::regex` on every call. For the current test count it's fine, but `std::regex` construction is notoriously expensive in libstdc++. If the test suite grows, consider pre-compiling or switching to a simpler search.

banidoru (commented):
- `infmon_table_descriptor` duplicates `flow_rule_id` as two flat fields instead of embedding the `infmon_flow_rule_id` typedef — inconsistent with the rest of the schema and harder to maintain.
- `infmon_flow_rule_list` and `infmon_status` have no end-of-stream reply with `retval` — callers have no way to distinguish "zero results" from "request failed".
- Schema contract tests use global substring matching (`api.find("u64 hi")`) which passes even if the field exists in the wrong type block — consider scoping checks to the relevant definition block.
- Overall structure is solid; good separation of contract tests from vppapigen dependency.

src/backend/api/infmon.api
r12f: Good catch — removed the `autoreply` keyword from `infmon_flow_rule_add` so the explicit `infmon_flow_rule_add_reply` (with `flow_rule_id`) is used. Fixed in a91ce19.

src/backend/api/infmon.api
r12f: You're right, the comment was misleading. Updated it to say "Returns all workers in a single infmon_status_details message via a variable-length array (workers[worker_count])." The bundled-VLA pattern is intentional here (worker count is small and bounded), unlike flow_rule_list which streams. Fixed in a91ce19.

src/backend/cmake/FindVppApiGen.cmake
r12f: Good point — changed to derive the target name from the API file basename using NAME_WE, so it becomes `${stem}_api_generated` (e.g. `infmon_api_generated` for `infmon.api`). A second .api file will get its own target automatically. Fixed in a91ce19.

src/backend/api/infmon.api
r12f: Added a doc comment on the `fields[5]` array: "Fixed-size: max 5 fields per rule. Bump array + version if more needed." Keeping it fixed-size for now since the wire format stability is important and 5 covers all current field types. Fixed in a91ce19.

src/backend/test/test_api_schema.cc
```diff
@@ -0,0 +1,240 @@
...
+
+#include <gtest/gtest.h>
+
+#ifndef API_FILE
+#error "API_FILE must be defined at compile time (path to infmon.api)"
+#endif
+
+static std::string read_file(const char *path)
+{
+    std::ifstream f(path);
```
banidoru: `EXPECT_TRUE` here is non-fatal — if the file doesn't open, `read_file` returns an empty string and every subsequent `expect_contains` fails with confusing "Missing pattern" messages instead of one clear "can't open file" failure. Switch to `ASSERT_TRUE` so the test aborts immediately on file-open failure.

src/backend/api/infmon.api
```diff
@@ -0,0 +1,227 @@
...
+    u64 generation;
+    u64 epoch_ns;
+    u64 slots_offset;
+    u32 slots_len;
+    u64 key_arena_offset;
+    u32 key_arena_capacity;
+    u32 key_arena_used;
+    u64 insert_failed;
+    u64 table_full;
+};
```
banidoru, line 87: The field ordering in `infmon_table_descriptor` mixes u32 and u64 types (e.g. `u32 flow_rule_index` at offset 20, then `u64 generation` at offset 24 on a packed layout, or offset 32 with natural alignment). Since this is a wire format, the implicit padding depends on whether vppapigen packs or aligns. Consider grouping all u64 fields first, or adding explicit `u32 _pad` fields to make the wire layout unambiguous. Same concern applies to `infmon_flow_rule_details`.

src/backend/test/test_api_schema.cc
```diff
@@ -0,0 +1,240 @@
...
+{
+    const std::vector<std::string> required_messages = {
+        "infmon_flow_rule_add", "infmon_flow_rule_del",      "infmon_flow_rule_list",
+        "infmon_flow_rule_get", "infmon_snapshot_and_clear", "infmon_status",
+    };
+
+    for (const auto &msg : required_messages) {
+        expect_contains("define " + msg, "Missing required message: " + msg);
+    }
+}
```
banidoru, line 255: No negative/uniqueness tests — e.g. asserting that `define infmon_flow_rule_add` appears exactly once (not twice due to `autoreply` auto-generation) would have caught the `autoreply` conflict bug. Consider adding occurrence-count checks for critical message names.

banidoru (commented):
- **`read_file` uses non-fatal `EXPECT_TRUE`** — if the .api file can't be opened, test continues with empty string and every subsequent check fails with misleading messages. Use `ASSERT_TRUE` to fail-fast.
- **`infmon_table_descriptor` has implicit padding holes** — mixed u32/u64 field ordering (e.g. `u32 flow_rule_index` then `u64 generation`) creates compiler-inserted padding on the wire. For a binary API, consider reordering fields by alignment (all u64 first, then u32) or adding explicit padding fields to make the layout unambiguous.
- **No negative test cases** — schema tests only check that expected elements are present, but don't verify constraints like field ordering within a block or catch accidental duplicates. A test that counts occurrences of `define infmon_flow_rule_add` (should be exactly 1, but `autoreply` may generate a second) would have caught the `autoreply` bug flagged by another reviewer.

src/backend/api/infmon.api
r12f: Good catch — replaced the flat `u64 flow_rule_id_hi/lo` in `infmon_table_descriptor` with `vl_api_infmon_flow_rule_id_t flow_rule_id` to match the rest of the API. Updated test assertions accordingly. Fixed in cb4529c.

src/backend/api/infmon.api
r12f: You're right — added terminal `infmon_flow_rule_list_reply` and `infmon_status_reply` messages (both with `context` + `retval`) so clients can distinguish zero-results from errors. Added corresponding test coverage. Fixed in cb4529c.

src/backend/test/test_api_schema.cc
r12f: Valid concern. Scoped the `u64 hi`/`u64 lo` assertions to match within the `typedef infmon_flow_rule_id` block using `expect_matches` with a regex anchored to the typedef name. Same pattern applied to the table_descriptor `flow_rule_id` check. Fixed in cb4529c.

src/backend/test/test_api_schema.cc
r12f: Acknowledged — `std::regex` construction cost is real in libstdc++. For the current test count (~15 calls) it's negligible, so leaving as-is for now. If the test suite grows significantly we can switch to pre-compiled patterns or `re2`. No code change needed.

src/backend/test/test_api_schema.cc
r12f: Good catch — switched to ASSERT_TRUE in SetUp() so the test aborts immediately on file-open failure instead of cascading confusing assertion failures. Also removed the now-unused standalone read_file function. Fixed in 5ffbea2.

src/backend/api/infmon.api
r12f: vppapigen generates packed wire layout (no implicit padding), so the mixed u32/u64 ordering is safe — fields are laid out sequentially without alignment gaps. Added a comment documenting this explicitly on infmon_table_descriptor. No reordering needed. Fixed in 5ffbea2.

src/backend/test/test_api_schema.cc
r12f: Great suggestion — added a NoDuplicateMessageDefinitions test that asserts exactly 1 occurrence of each critical `define <msg>` line. This would have caught the autoreply conflict. Fixed in 5ffbea2.

banidoru (commented):
Re-review (iteration 2). All prior feedback addressed — nice cleanup.

- `infmon_flow_rule_details` still uses flat `u64 flow_rule_id_hi/lo` instead of `vl_api_infmon_flow_rule_id_t` — same inconsistency that was just fixed in `infmon_table_descriptor`
- `FindVppApiGen.cmake` header comment still references hardcoded `"infmon_api_generated"` target name (stale after the NAME_WE fix)
- Schema, build integration, and test coverage all look solid otherwise
- Terminal `_reply` messages and `NoDuplicateMessageDefinitions` test are good additions

src/backend/api/infmon.api
```diff
@@ -0,0 +1,244 @@
...
+
+/** Per-flow-rule summary (used in list/get replies). */
+typedef infmon_flow_rule_details
+{
+    u64 flow_rule_id_hi;
+    u64 flow_rule_id_lo;
+    u32 flow_rule_index;
+    string name[32];       /* NUL-terminated, max 31 chars + NUL */
+    u32 field_count;
+    /** Fixed-size: max 5 fields per rule. Bump array + version if more needed. */
```
banidoru, line 96: `infmon_flow_rule_details` still uses flat `u64 flow_rule_id_hi` / `u64 flow_rule_id_lo` instead of embedding `vl_api_infmon_flow_rule_id_t flow_rule_id`. This was fixed in `infmon_table_descriptor` (cb4529c) but missed here — same inconsistency, same fix needed.

src/backend/cmake/FindVppApiGen.cmake
```diff
@@ -0,0 +1,92 @@
...
+#   find_package(VppApiGen)         # optional — VPPAPIGEN_FOUND is set
+#   if(VPPAPIGEN_FOUND)
+#     vpp_generate_api_header(
+#       API_FILE  src/backend/api/infmon.api
+#       OUTPUT_DIR ${CMAKE_CURRENT_BINARY_DIR}/generated
+#     )
+#   endif()
+#
+# The generated header is: ${OUTPUT_DIR}/<basename>.api.h
+# A CMake target "${basename}_api_generated" is created that depends on it.
```
banidoru, line 17: Stale comment: still says `A CMake target "infmon_api_generated" is created` but the code now derives the target name from the API file basename. Should say something like `A CMake target "${basename}_api_generated" is created`.

src/backend/api/infmon.api
banidoru: Still open — `infmon_flow_rule_details` still uses flat `u64 flow_rule_id_hi/lo` instead of `vl_api_infmon_flow_rule_id_t`. Same fix as `infmon_table_descriptor` needed here.

src/backend/cmake/FindVppApiGen.cmake
banidoru: Still open — the header comment on line 17 still references a hardcoded `infmon_api_generated` target name. Should be updated to reflect the dynamic `${basename}_api_generated` naming.

banidoru: Correction: looking at the latest diff, line 17 now reads `${basename}_api_generated` — this is fixed. Resolving.

banidoru (changes requested):
- Most prior feedback addressed well — autoreply fix, terminal replies, ASSERT_TRUE, uniqueness tests, packed layout docs all look good
- `infmon_flow_rule_details` still uses flat `u64 flow_rule_id_hi/lo` instead of `vl_api_infmon_flow_rule_id_t` — same inconsistency that was fixed in `infmon_table_descriptor`
- Test assertions for `flow_rule_details` still check the flat field names, will need updating with the typedef fix
- Overall structure is solid — schema covers all spec sections, CMake integration is clean, test coverage is thorough

src/backend/api/infmon.api
r12f: Fixed — replaced flat u64 flow_rule_id_hi/lo with vl_api_infmon_flow_rule_id_t flow_rule_id in infmon_flow_rule_details, matching the infmon_table_descriptor fix from cb4529c. Updated test assertions accordingly. Fixed in 408c567.

banidoru: Still not addressed in the latest commits. `infmon_flow_rule_details` continues to use flat `u64 flow_rule_id_hi` / `u64 flow_rule_id_lo` instead of `vl_api_infmon_flow_rule_id_t flow_rule_id`. The same fix applied to `infmon_table_descriptor` in cb4529c should be applied here for consistency.

banidoru (changes requested):
- `infmon_flow_rule_details` still uses flat `u64 flow_rule_id_hi/lo` instead of `vl_api_infmon_flow_rule_id_t` — same fix as `infmon_table_descriptor` needed
- All prior review feedback (autoreply removal, terminal replies, ASSERT_TRUE, scoped assertions, duplicate checks, dynamic CMake target, packed layout docs) properly addressed
- Test coverage for uniqueness and terminal reply messages looks solid
- Wire schema is clean and well-documented overall
- One remaining consistency fix blocks approval

banidoru (approved):
- All prior review concerns addressed across iterations — autoreply fix, flow_rule_id typedef consistency, terminal reply messages, ASSERT_TRUE in SetUp, dynamic CMake target names, duplicate-message test, and packed-layout documentation all look good.
- API schema is well-structured: clean enum/typedef/message separation, proper VPP conventions (client_index/context/retval), and good use of autoreply where appropriate.
- Test coverage is solid: contract tests catch schema drift without requiring vppapigen, scoped regex assertions avoid false positives, and uniqueness checks guard against autoreply conflicts.
- CMake integration handles missing vppapigen gracefully with dummy targets.

banidoru (commented):
Re-review (iteration 3) — all prior concerns addressed.

- All 12 review threads resolved: `autoreply` removal, typedef consistency, terminal replies, test improvements, comment fixes all landed correctly.
- `infmon_flow_rule_details` now uses `vl_api_infmon_flow_rule_id_t` consistently with `infmon_table_descriptor` — good.
- `NoDuplicateMessageDefinitions` test and `ASSERT_TRUE` in SetUp() are solid additions.
- Wire format documentation (packed layout comment) and `fields[5]` limit doc are clear.

banidoru (approved):
- All prior review concerns addressed across iterations — clean fixes
- API schema is well-structured: proper enums, typedefs, message pairs, terminal replies
- CMake module handles missing vppapigen gracefully with dynamic target naming
- Test coverage is solid: contract tests, scoped assertions, duplicate detection

banidoru (approved):
All reviewers approved. LGTM.

PR #80, opened 2026-04-19
feat(api): add infmon_flow_rule_list and infmon_flow_rule_get handlers (W10c)

### LINKED ISSUE #46: [W10c] infmon_flow_rule_list / infmon_flow_rule_get API messages
## Spec

## Scope
- infmon_flow_rule_list: enumerate active flow_rules
- infmon_flow_rule_get: return the full definition for one flow_rule_id

## Dependencies
Blocked by: #44, #45

r12f:
## Summary
Add two new API handler functions for flow rule enumeration and retrieval:
- `infmon_api_flow_rule_list()`: enumerate active flow rules via callback
- `infmon_api_flow_rule_get_by_name()`: retrieve a single rule by name and return pointer + index

The .api wire schema already had the message definitions from the initial API schema commit. This PR adds the C handler implementations and 9 new tests.

## Tests
All 16 api_handler tests pass (7 existing + 9 new):
- ListEmptyReturnsOk, ListMultipleRules, ListNullCtxReturnsError, ListNullCallbackDoesNotCrash
- GetByNameFound, GetByNameNotFound, GetByNameNullInputs, GetByNameNullOutputs
- ListAfterDeleteReflectsRemoval

Closes #46

### DIFF
--- a/src/backend/include/infmon/api_handler.h
+++ b/src/backend/include/infmon/api_handler.h
@@ -68,6 +68,31 @@ infmon_api_result_t infmon_api_flow_rule_add(infmon_api_ctx_t *ctx, const infmon
  */
 infmon_api_result_t infmon_api_flow_rule_del(infmon_api_ctx_t *ctx, const char *name);
 
+/**
+ * Callback invoked once per active flow rule during infmon_api_flow_rule_list().
+ * @p rule  points into the rule set (valid only for the duration of the callback).
+ * @p index is the rule's position in the set.
+ * @p user  is the opaque pointer passed to infmon_api_flow_rule_list().
+ */
+typedef void (*infmon_api_flow_rule_list_cb_t)(const infmon_flow_rule_t *rule, uint32_t index,
+                                               void *user);
+
+/**
+ * Enumerate all active flow rules, invoking @p cb once per rule.
+ * Returns INFMON_API_OK on success, or an error if @p ctx is NULL.
+ */
+infmon_api_result_t infmon_api_flow_rule_list(const infmon_api_ctx_t *ctx,
+                                              infmon_api_flow_rule_list_cb_t cb, void *user);
+
+/**
+ * Retrieve a single flow rule by name.
+ * On success, *out_rule points into the rule set (valid until next add/rm)
+ * and *out_index holds the rule's position.
+ */
+infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx, const char *name,
+                                                     const infmon_flow_rule_t **out_rule,
+                                                     uint32_t *out_index);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/backend/include/infmon/api_handler.h
+++ b/src/backend/include/infmon/api_handler.h
@@ -91,6 +91,10 @@ infmon_api_result_t infmon_api_flow_rule_list(const infmon_api_ctx_t *ctx,
  * Retrieve a single flow rule by name.
  * On success, *out_rule points into the rule set (valid until next add/rm)
  * and *out_index holds the rule's position.
+ *
+ * @return INFMON_API_OK on success.
+ * @return INFMON_API_ERR_INVALID_RULE if @p ctx or @p name is NULL.
+ * @return INFMON_API_ERR_NOT_FOUND if no rule with @p name exists.
  */
 infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx, const char *name,
                                                      const infmon_flow_rule_t **out_rule,
--- a/src/backend/include/infmon/api_handler.h
+++ b/src/backend/include/infmon/api_handler.h
@@ -70,8 +70,8 @@ infmon_api_result_t infmon_api_flow_rule_del(infmon_api_ctx_t *ctx, const char *
 
 /**
  * Callback invoked once per active flow rule during infmon_api_flow_rule_list().
- * @p rule  points into the rule set (valid only for the duration of the callback).
- * @p index is the rule's position in the set.
+ * @p rule  points into the rule set (valid until the next add/rm operation).
+ * @p index is the rule's position in the set (dense, 0..count-1).
  * @p user  is the opaque pointer passed to infmon_api_flow_rule_list().
  */
 typedef void (*infmon_api_flow_rule_list_cb_t)(const infmon_flow_rule_t *rule, uint32_t index,
@@ -79,6 +79,9 @@ typedef void (*infmon_api_flow_rule_list_cb_t)(const infmon_flow_rule_t *rule, u
 
 /**
  * Enumerate all active flow rules, invoking @p cb once per rule.
+ * The rule set must not be modified during iteration (no re-entrant
+ * add/del from within the callback).
+ * @p cb may be NULL (no-op, returns INFMON_API_OK immediately).
  * Returns INFMON_API_OK on success, or an error if @p ctx is NULL.
  */
 infmon_api_result_t infmon_api_flow_rule_list(const infmon_api_ctx_t *ctx,
--- a/src/backend/src/api_handler.c
+++ b/src/backend/src/api_handler.c
@@ -138,3 +138,39 @@ infmon_api_result_t infmon_api_flow_rule_del(infmon_api_ctx_t *ctx, const char *
 
     return INFMON_API_OK;
 }
+
+infmon_api_result_t infmon_api_flow_rule_list(const infmon_api_ctx_t *ctx,
+                                              infmon_api_flow_rule_list_cb_t cb, void *user)
+{
+    if (!ctx)
+        return INFMON_API_ERR_INVALID_RULE;
+
+    uint32_t n = infmon_flow_rule_count(ctx->rule_set);
+    for (uint32_t i = 0; i < n; i++) {
+        const infmon_flow_rule_t *r = infmon_flow_rule_get(ctx->rule_set, i);
+        if (r && cb)
+            cb(r, i, user);
+    }
+    return INFMON_API_OK;
+}
+
+infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx, const char *name,
+                                                     const infmon_flow_rule_t **out_rule,
+                                                     uint32_t *out_index)
+{
+    if (!ctx || !name)
+        return INFMON_API_ERR_INVALID_RULE;
+
+    uint32_t n = infmon_flow_rule_count(ctx->rule_set);
+    for (uint32_t i = 0; i < n; i++) {
+        const infmon_flow_rule_t *r = infmon_flow_rule_get(ctx->rule_set, i);
+        if (r && strcmp(r->name, name) == 0) {
+            if (out_rule)
+                *out_rule = r;
+            if (out_index)
+                *out_index = i;
+            return INFMON_API_OK;
+        }
+    }
+    return INFMON_API_ERR_NOT_FOUND;
+}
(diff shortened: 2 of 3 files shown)


src/backend/src/api_handler.c
```diff
@@ -138,3 +138,39 @@ infmon_api_result_t infmon_api_flow_rule_del(infmon_api_ctx_t *ctx, const char *
...
+infmon_api_result_t infmon_api_flow_rule_list(const infmon_api_ctx_t *ctx,
+                                              infmon_api_flow_rule_list_cb_t cb, void *user)
+{
+    if (!ctx)
+        return INFMON_API_ERR_INVALID_RULE;
+
+    uint32_t n = infmon_flow_rule_count(ctx->rule_set);
+    for (uint32_t i = 0; i < n; i++) {
+        const infmon_flow_rule_t *r = infmon_flow_rule_get(ctx->rule_set, i);
+        if (r && cb)
```
banidoru: The `cb` null-check is evaluated on every loop iteration. If a null callback is intentionally supported ("just validate ctx"), move the guard before the loop and return early. If it's not a valid use case, reject it at the top with an error. Either way avoids a redundant branch per rule.
```c
if (!cb)
    return INFMON_API_OK;
```

src/backend/src/api_handler.c
```diff
@@ -138,3 +138,39 @@ infmon_api_result_t infmon_api_flow_rule_del(infmon_api_ctx_t *ctx, const char *
 
     return INFMON_API_OK;
 }
+
+infmon_api_result_t infmon_api_flow_rule_list(const infmon_api_ctx_t *ctx,
+                                              infmon_api_flow_rule_list_cb_t cb, void *user)
```
banidoru, line 143: `INFMON_API_ERR_INVALID_RULE` for a null `ctx` is semantically misleading — the rule isn't invalid, the context is. Same applies to `flow_rule_get_by_name` returning this for null `name`. If adding a new error code isn't desirable right now, a brief comment explaining the reuse would help.

src/backend/src/api_handler.c
```diff
@@ -138,3 +138,39 @@ infmon_api_result_t infmon_api_flow_rule_del(infmon_api_ctx_t *ctx, const char *
...
+}
+
+infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx, const char *name,
+                                                     const infmon_flow_rule_t **out_rule,
+                                                     uint32_t *out_index)
+{
+    if (!ctx || !name)
+        return INFMON_API_ERR_INVALID_RULE;
+
+    uint32_t n = infmon_flow_rule_count(ctx->rule_set);
```
banidoru, line 173: Nit: `strcmp(r->name, name)` — if `r->name` can ever be NULL (e.g. a partially-initialized rule slot), this will segfault. Defensive check: `r->name && strcmp(r->name, name) == 0`.

banidoru (commented):
- `flow_rule_list`: null callback check is inside the loop — moves the branch into every iteration instead of short-circuiting up front. Consider returning early or documenting the intent if "iterate without callback" is a valid use case.
- `INFMON_API_ERR_INVALID_RULE` is reused for null-ctx/null-name validation errors — semantically these are invalid arguments, not invalid rules. If the error enum grows, consider a dedicated `ERR_INVALID_ARG`.
- `flow_rule_get_by_name` linear scan is fine at current scale; just flag that it's O(n) if rule counts grow significantly.
- Tests are thorough — good coverage of null inputs, empty sets, post-delete consistency, and both output-param combinations.

src/backend/include/infmon/api_handler.h
```diff
@@ -68,6 +68,31 @@ infmon_api_result_t infmon_api_flow_rule_add(infmon_api_ctx_t *ctx, const infmon
  */
 infmon_api_result_t infmon_api_flow_rule_del(infmon_api_ctx_t *ctx, const char *name);
 
+/**
+ * Callback invoked once per active flow rule during infmon_api_flow_rule_list().
+ * @p rule  points into the rule set (valid only for the duration of the callback).
```
banidoru: The callback doc says the rule pointer is valid "only for the duration of the callback", but `get_by_name` below says "valid until next add/rm". Both functions return pointers into the same rule set, so the lifetime guarantee should be consistent. If the list callback pointer is also valid until next mutation, say so — it's useful and avoids unnecessary copies in callers.

src/backend/src/api_handler.c
```diff
@@ -138,3 +138,39 @@ infmon_api_result_t infmon_api_flow_rule_del(infmon_api_ctx_t *ctx, const char *
...
 }
+
+infmon_api_result_t infmon_api_flow_rule_list(const infmon_api_ctx_t *ctx,
+                                              infmon_api_flow_rule_list_cb_t cb, void *user)
+{
+    if (!ctx)
+        return INFMON_API_ERR_INVALID_RULE;
+
+    uint32_t n = infmon_flow_rule_count(ctx->rule_set);
+    for (uint32_t i = 0; i < n; i++) {
```
banidoru, line 158: If `infmon_flow_rule_get` returns NULL for some index (sparse set?), the callback still receives `i` as the index for the *next* non-null rule. This means the index passed to the callback can have gaps, which may surprise callers expecting a dense 0..n-1 sequence. Either document this behavior or track a separate `emitted` counter to pass as the index.

banidoru (commented):
- Pointer lifetime semantics differ between `list` ("duration of callback") and `get_by_name` ("until next add/rm") — clarify or unify.
- `flow_rule_list` silently skips null rule slots but still advances the index counter, so callback receives non-contiguous indices — document or compact.
- No guard against rule-set mutation during `flow_rule_list` iteration — a callback that adds/deletes rules could corrupt the loop. Worth a doc warning at minimum.
- Good test coverage overall; edge cases for null inputs are well handled.

src/backend/include/infmon/api_handler.h
```diff
@@ -68,6 +68,31 @@ infmon_api_result_t infmon_api_flow_rule_add(infmon_api_ctx_t *ctx, const infmon
...
+ * @p index is the rule's position in the set.
+ * @p user  is the opaque pointer passed to infmon_api_flow_rule_list().
+ */
+typedef void (*infmon_api_flow_rule_list_cb_t)(const infmon_flow_rule_t *rule, uint32_t index,
+                                               void *user);
+
+/**
+ * Enumerate all active flow rules, invoking @p cb once per rule.
```
banidoru, line 81: Consider documenting that the rule set must not be modified during iteration (no re-entrant add/del from within the callback). Mutating the set mid-traversal could invalidate the count or shift indices, leading to skipped rules or out-of-bounds access.

src/backend/test/test_api_handler.cc
```diff
@@ -123,3 +125,114 @@ TEST_F(ApiHandlerTest, AddSetFull)
...
+    ASSERT_EQ(infmon_api_flow_rule_del(&ctx_, "rule_aa"), INFMON_API_OK);
+
+    std::vector<ListEntry> entries;
+    EXPECT_EQ(infmon_api_flow_rule_list(&ctx_, list_cb, &entries), INFMON_API_OK);
+    ASSERT_EQ(entries.size(), 1u);
+    EXPECT_EQ(entries[0].name, "rule_bb");
+    EXPECT_EQ(entries[0].index, 0u);
+
+    /* get should also reflect removal. */
+    EXPECT_EQ(infmon_api_flow_rule_get_by_name(&ctx_, "rule_aa", nullptr, nullptr),
```
banidoru, line 236: This test verifies `get_by_name` returns NOT_FOUND for the deleted rule, but doesn't check that `rule_bb`'s index has shifted to 0 via `get_by_name`. Add:
```cpp
const infmon_flow_rule_t *found = nullptr;
uint32_t idx = (uint32_t)-1;
EXPECT_EQ(infmon_api_flow_rule_get_by_name(&ctx_, "rule_bb", &found, &idx), INFMON_API_OK);
EXPECT_EQ(idx, 0u);
```
This would exercise the interaction between delete and get_by_name more thoroughly.

src/backend/include/infmon/api_handler.h
```diff
@@ -68,6 +68,31 @@ infmon_api_result_t infmon_api_flow_rule_add(infmon_api_ctx_t *ctx, const infmon
...
+ * Returns INFMON_API_OK on success, or an error if @p ctx is NULL.
+ */
+infmon_api_result_t infmon_api_flow_rule_list(const infmon_api_ctx_t *ctx,
+                                              infmon_api_flow_rule_list_cb_t cb, void *user);
+
+/**
+ * Retrieve a single flow rule by name.
+ * On success, *out_rule points into the rule set (valid until next add/rm)
+ * and *out_index holds the rule's position.
+ */
```
banidoru, line 98: The doc comment doesn't list possible return values. Add `@return INFMON_API_OK`, `INFMON_API_ERR_NOT_FOUND`, `INFMON_API_ERR_INVALID_RULE` so callers can handle errors without reading the implementation.

banidoru (commented):
- Implementation is clean and correct for the happy paths; good test coverage across null inputs, empty sets, and post-delete consistency.
- Existing review comments on cb-null-check hoisting, error code semantics, and strcmp null-safety are all valid — agree with those.
- Missing test: verify `get_by_name` returns the *updated* index after a deletion (e.g. rule_bb should be at index 0 after rule_aa is removed).
- Header could benefit from a `@return` tag documenting `INFMON_API_ERR_NOT_FOUND` for `get_by_name` — callers shouldn't have to read the source to discover possible error codes.

banidoru (approved):
All reviewers approved. LGTM.

src/backend/src/api_handler.c
r12f: Good catch — moved the `cb` null-check before the loop as an early return. If cb is NULL we now return INFMON_API_OK immediately without entering the loop. Fixed in 8454b1c.

src/backend/src/api_handler.c
r12f: Fair point. Rather than adding a new error code right now (would ripple through the API surface), I've added a comment at both call sites explaining the reuse: ERR_INVALID_RULE is the closest semantic match since the rule is the primary domain object and we don't have a dedicated 'invalid context' code yet. Fixed in 8454b1c.

src/backend/src/api_handler.c
r12f: Good defensive suggestion. Added `r->name &&` guard in both `find_rule_index` and `flow_rule_get_by_name`. In practice the rule set shouldn't contain entries with NULL names, but the guard costs nothing and prevents a potential segfault. Fixed in 8454b1c.

src/backend/include/infmon/api_handler.h
r12f: You're right — both return pointers into the same dense rule set, so the lifetime is identical. Updated the list callback doc to say "valid until the next add/rm operation" to match get_by_name. This avoids unnecessary copies in callers. Fixed in 8454b1c.

src/backend/src/api_handler.c
r12f: The rule set is dense (add appends, rm compacts via the array shift in flow_rule_del), so flow_rule_get should never return NULL for i < count. Added a comment in the implementation documenting this invariant and noting the NULL guard is purely defensive. Also updated the callback doc to state indices are dense 0..count-1. Fixed in 8454b1c.

src/backend/include/infmon/api_handler.h
r12f: Good call — added explicit documentation on infmon_api_flow_rule_list() that the rule set must not be modified during iteration (no re-entrant add/del from within the callback). Fixed in 8454b1c.

src/backend/test/test_api_handler.cc
r12f: Great suggestion — added the get_by_name assertion to verify rule_bb shifts to index 0 after deleting rule_aa. Also checks the returned pointer and name. Fixed in ead76e7.

src/backend/include/infmon/api_handler.h
r12f: Good point — added @return documentation listing INFMON_API_OK, INFMON_API_ERR_INVALID_RULE (null ctx/name), and INFMON_API_ERR_NOT_FOUND. Fixed in ead76e7.

PR #81, opened 2026-04-19
feat: wire snapshot_and_clear API to swap logic

### LINKED ISSUE #47: [W10d] infmon_snapshot_and_clear API message
## Spec

## Scope
- Wire the infmon_snapshot_and_clear(flow_rule_id) API message to the swap logic from W8
- Return the descriptor of the retired table (generation G)
- Ensure the reply contains all fields needed by the frontend to walk the retired table

## Dependencies
Blocked by: #44, #42

r12f:
## Summary

Wire `infmon_snapshot_and_clear(flow_rule_id)` API message to the swap logic from W8.

### Changes

- **api_handler.h**: Add `snap_mgr` and `flow_rule_ids[]` to `infmon_api_ctx_t`, new error codes (`ALLOC_FAILED`, `TOO_MANY_RETIRED`, `NULL_TABLE`, `NO_SNAPSHOT_MGR`), `infmon_api_snap_reply_t` struct, and declarations for `snapshot_and_clear` and `flow_rule_add_with_id`
- **api_handler.c**: Implement `infmon_api_snapshot_and_clear()` — resolves flow_rule_id → index, calls snapshot manager, publishes stats descriptor for the retired table; add `find_rule_index_by_id()` helper, `map_snap_result()`, `flow_rule_add_with_id()`, and ID compaction on delete
- **CMakeLists.txt**: Link `infmon_snapshot` to api_handler and test targets
- **test_api_handler.cc**: 7 new tests covering basic swap, not-found ID, null ctx, missing snap_mgr, descriptor fields, multiple swaps, and ID preservation across delete/compaction

### Test Plan

All 8 test suites pass (23 api_handler tests total, 7 new).

Closes #47

### DIFF
--- a/src/backend/include/infmon/api_handler.h
+++ b/src/backend/include/infmon/api_handler.h
@@ -139,7 +139,7 @@ infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx
  * otherwise).
  */
 void infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx, infmon_flow_rule_id_t flow_rule_id,
-                                  infmon_api_snap_reply_t *reply);
+                                   infmon_api_snap_reply_t *reply);
 
 #ifdef __cplusplus
 }
--- a/src/backend/include/infmon/api_handler.h
+++ b/src/backend/include/infmon/api_handler.h
@@ -13,6 +13,7 @@
 
 #include "infmon/counter_table.h"
 #include "infmon/flow_rule.h"
+#include "infmon/snapshot.h"
 #include "infmon/stats_segment.h"
 
 #ifdef __cplusplus
@@ -30,6 +31,10 @@ typedef enum {
     INFMON_API_ERR_SET_FULL,
     INFMON_API_ERR_STATS_PUBLISH,
     INFMON_API_ERR_INTERNAL,
+    INFMON_API_ERR_ALLOC_FAILED,
+    INFMON_API_ERR_TOO_MANY_RETIRED,
+    INFMON_API_ERR_NULL_TABLE,
+    INFMON_API_ERR_NO_SNAPSHOT_MGR,
 } infmon_api_result_t;
 
 /* ── Context (caller-owned, long-lived) ──────────────────────────── */
@@ -37,10 +42,24 @@ typedef enum {
 typedef struct {
     infmon_flow_rule_set_t *rule_set;
     infmon_stats_registry_t *stats_reg;
+    infmon_snapshot_mgr_t *snap_mgr; /**< Optional; required for snapshot_and_clear. */
     /* private: Per-rule counter tables, indexed by rule position in the set.
      * Managed internally by add/del; caller must not touch. */
     infmon_counter_table_t *tables[INFMON_FLOW_RULE_SET_MAX];
+    /* Per-rule UUID, indexed in parallel with tables[]. */
+    infmon_flow_rule_id_t flow_rule_ids[INFMON_FLOW_RULE_SET_MAX];
 } infmon_api_ctx_t;
+
+/* ── Snapshot reply ──────────────────────────────────────────────── */
+
+/**
+ * Result of infmon_api_snapshot_and_clear().  On success, the descriptor
+ * is fully populated with retired table metadata.
+ */
+typedef struct {
+    infmon_api_result_t result;
+    infmon_stats_descriptor_t descriptor; /**< Populated on success. */
+} infmon_api_snap_reply_t;
 
 /* ── Lifecycle ───────────────────────────────────────────────────── */
 
@@ -61,6 +80,15 @@ void infmon_api_ctx_destroy(infmon_api_ctx_t *ctx);
  * counter table → publish stats descriptor.
  */
 infmon_api_result_t infmon_api_flow_rule_add(infmon_api_ctx_t *ctx, const infmon_flow_rule_t *rule);
+
+/**
+ * Add a flow rule with an explicit UUID.
+ * Same as infmon_api_flow_rule_add but also records @p id so that
+ * snapshot_and_clear can resolve the ID → index mapping.
+ */
+infmon_api_result_t infmon_api_flow_rule_add_with_id(infmon_api_ctx_t *ctx,
+                                                     const infmon_flow_rule_t *rule,
+                                                     infmon_flow_rule_id_t id);
 
 /**
  * Delete a flow rule by name: unpublish stats → destroy counter table →
@@ -99,6 +127,19 @@ infmon_api_result_t infmon_api_flow_rule_list(const infmon_api_ctx_t *ctx,
 infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx, const char *name,
                                                      const infmon_flow_rule_t **out_rule,
                                                      uint32_t *out_index);
+
+/**
+ * Perform an atomic snapshot_and_clear on the counter table belonging
+ * to the flow rule identified by @p flow_rule_id.
+ *
+ * On success the reply's descriptor is fully populated with the
+ * retired table's metadata (generation, offsets, counters).
+ *
+ * Requires ctx->snap_mgr to be set (returns INFMON_API_ERR_NO_SNAPSHOT_MGR
+ * otherwise).
+ */
+void infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx, infmon_flow_rule_id_t flow_rule_id,
+                                  infmon_api_snap_reply_t *reply);
 
 #ifdef __cplusplus
 }
--- a/src/backend/include/infmon/api_handler.h
+++ b/src/backend/include/infmon/api_handler.h
@@ -136,10 +136,15 @@ infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx
  * retired table's metadata (generation, offsets, counters).
  *
  * Requires ctx->snap_mgr and ctx->stats_reg to be set.
- * @p reply must be non-NULL; if NULL, the call is a no-op.
+ * @p reply must be non-NULL; if NULL, returns INFMON_API_ERR_INTERNAL.
+ * @p flow_rule_id must be non-zero (zero is a sentinel for "no ID assigned").
+ *
+ * @return INFMON_API_OK on success, or an error code.  The same code is also
+ *         stored in reply->result for callers that prefer struct-based checking.
  */
-void infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx, infmon_flow_rule_id_t flow_rule_id,
-                                   infmon_api_snap_reply_t *reply);
+infmon_api_result_t infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx,
+                                                  infmon_flow_rule_id_t flow_rule_id,
+                                                  infmon_api_snap_reply_t *reply);
 
 #ifdef __cplusplus
 }
--- a/src/backend/include/infmon/api_handler.h
+++ b/src/backend/include/infmon/api_handler.h
@@ -135,8 +135,8 @@ infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx
  * On success the reply's descriptor is fully populated with the
  * retired table's metadata (generation, offsets, counters).
  *
- * Requires ctx->snap_mgr to be set (returns INFMON_API_ERR_NO_SNAPSHOT_MGR
- * otherwise).
+ * Requires ctx->snap_mgr and ctx->stats_reg to be set.
+ * @p reply must be non-NULL; if NULL, the call is a no-op.
  */
 void infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx, infmon_flow_rule_id_t flow_rule_id,
                                    infmon_api_snap_reply_t *reply);
--- a/src/backend/src/api_handler.c
+++ b/src/backend/src/api_handler.c
@@ -245,7 +245,7 @@ infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx
 /* ── Snapshot and clear ──────────────────────────────────────────── */
 
 void infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx, infmon_flow_rule_id_t flow_rule_id,
-                                  infmon_api_snap_reply_t *reply)
+                                   infmon_api_snap_reply_t *reply)
 {
     memset(reply, 0, sizeof(*reply));
 
@@ -269,7 +269,7 @@ void infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx, infmon_flow_rule_id_t 
     /* Delegate to the snapshot manager. */
     infmon_snap_reply_t snap_reply;
     infmon_snapshot_and_clear(ctx->snap_mgr, ctx->tables, idx, INFMON_FLOW_RULE_SET_MAX,
-                             infmon_flow_rule_get(ctx->rule_set, idx)->key_width, &snap_reply);
+                              infmon_flow_rule_get(ctx->rule_set, idx)->key_width, &snap_reply);
 
     reply->result = map_snap_result(snap_reply.result);
 
--- a/src/backend/src/api_handler.c
+++ b/src/backend/src/api_handler.c
@@ -247,29 +247,36 @@ infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx
 
 /* ── Snapshot and clear ──────────────────────────────────────────── */
 
-void infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx, infmon_flow_rule_id_t flow_rule_id,
-                                   infmon_api_snap_reply_t *reply)
+infmon_api_result_t infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx,
+                                                  infmon_flow_rule_id_t flow_rule_id,
+                                                  infmon_api_snap_reply_t *reply)
 {
     if (!reply)
-        return;
+        return INFMON_API_ERR_INTERNAL;
 
     memset(reply, 0, sizeof(*reply));
 
     if (!ctx) {
-        reply->result = INFMON_API_ERR_INVALID_RULE;
-        return;
+        reply->result = INFMON_API_ERR_INTERNAL;
+        return INFMON_API_ERR_INTERNAL;
     }
 
     if (!ctx->snap_mgr) {
         reply->result = INFMON_API_ERR_NO_SNAPSHOT_MGR;
-        return;
+        return INFMON_API_ERR_NO_SNAPSHOT_MGR;
+    }
+
+    /* Reject zero IDs — they are a sentinel for "no ID assigned". */
+    if (infmon_flow_rule_id_is_zero(flow_rule_id)) {
+        reply->result = INFMON_API_ERR_NOT_FOUND;
+        return INFMON_API_ERR_NOT_FOUND;
     }
 
     /* Resolve flow_rule_id → index. */
     uint32_t idx = find_rule_index_by_id(ctx, flow_rule_id);
     if (idx == (uint32_t) -1) {
         reply->result = INFMON_API_ERR_NOT_FOUND;
-        return;
+        return INFMON_API_ERR_NOT_FOUND;
     }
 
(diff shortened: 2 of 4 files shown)


src/backend/src/api_handler.c
```diff
@@ -106,9 +144,26 @@ infmon_api_result_t infmon_api_flow_rule_add(infmon_api_ctx_t *ctx, const infmon
     }
     ctx->tables[idx] = ct;
 
+    /* 5. Record the flow_rule_id if provided. */
+    if (id) {
+        ctx->flow_rule_ids[idx] = *id;
```
banidoru, line 150: When `id` is NULL (i.e. the plain `infmon_api_flow_rule_add` path), `ctx->flow_rule_ids[idx]` is left with whatever was there before. If the caller later adds a rule via `add_with_id`, then deletes it, then adds another via plain `add` — the slot could retain a stale ID that `find_rule_index_by_id` would match.

Consider adding an `else { memset(&ctx->flow_rule_ids[idx], 0, sizeof(ctx->flow_rule_ids[0])); }` branch.

src/backend/src/api_handler.c
```diff
@@ -183,3 +241,58 @@ infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx
     }
     return INFMON_API_ERR_NOT_FOUND;
 }
+
+/* ── Snapshot and clear ──────────────────────────────────────────── */
+
```
banidoru, line 249: No null-check on `reply` before `memset(reply, 0, ...)`. If a caller passes NULL this will segfault. Either add a guard or document that `reply` must be non-NULL in the header.

src/backend/src/api_handler.c
```diff
@@ -183,3 +241,58 @@ infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx
...
+        return;
+    }
+
+    /* Resolve flow_rule_id → index. */
+    uint32_t idx = find_rule_index_by_id(ctx, flow_rule_id);
+    if (idx == (uint32_t) -1) {
+        reply->result = INFMON_API_ERR_NOT_FOUND;
+        return;
+    }
+
```
banidoru, line 281: `infmon_flow_rule_get(ctx->rule_set, idx)` could theoretically return NULL if the rule set is in an inconsistent state. Dereferencing `->key_width` without a null-check is risky. A defensive check here would prevent a crash in edge cases.

src/backend/src/api_handler.c
```diff
@@ -183,3 +241,58 @@ infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx
...
+                             infmon_flow_rule_get(ctx->rule_set, idx)->key_width, &snap_reply);
+
+    reply->result = map_snap_result(snap_reply.result);
+
+    if (snap_reply.result != INFMON_SNAP_OK)
+        return;
+
+    /* Build the descriptor from the retired table. */
+    infmon_counter_table_t *retired = snap_reply.retired_table;
+    infmon_stats_descriptor_t *desc = &reply->descriptor;
```
banidoru, line 306: When `ctx->stats_reg` is NULL, `seg_base` is 0 and the offsets become raw pointer addresses. Is this intentional? If `stats_reg` is required, it should be validated at the top of the function alongside `snap_mgr`. If optional, the descriptor offsets are meaningless without a valid base.

banidoru (commented):
- `flow_rule_add_internal`: when `id` is NULL (plain `add` path), `flow_rule_ids[idx]` is not zeroed — stale data could cause `find_rule_index_by_id` to match a wrong rule
- `infmon_api_snapshot_and_clear`: no null-check on `reply` parameter before `memset`
- `infmon_api_snapshot_and_clear`: `infmon_flow_rule_get()` return not null-checked before dereferencing `->key_width`
- Tests are thorough and cover the key scenarios well; good compaction test

src/backend/include/infmon/api_handler.h
```diff
@@ -100,6 +128,19 @@ infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx
...
+ *
+ * On success the reply's descriptor is fully populated with the
+ * retired table's metadata (generation, offsets, counters).
+ *
+ * Requires ctx->snap_mgr to be set (returns INFMON_API_ERR_NO_SNAPSHOT_MGR
+ * otherwise).
+ */
+void infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx, infmon_flow_rule_id_t flow_rule_id,
+                                   infmon_api_snap_reply_t *reply);
+
```
banidoru, line 148: `infmon_api_snapshot_and_clear` returns `void`, unlike every other function in this API which returns `infmon_api_result_t`. This forces callers to use a different error-checking pattern just for this one call. Consider returning `infmon_api_result_t` directly (the reply struct can still carry the descriptor) for API consistency.

src/backend/src/api_handler.c
```diff
@@ -183,3 +241,58 @@ infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx
...
+
+/* ── Snapshot and clear ──────────────────────────────────────────── */
+
+void infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx, infmon_flow_rule_id_t flow_rule_id,
+                                   infmon_api_snap_reply_t *reply)
+{
+    memset(reply, 0, sizeof(*reply));
+
+    if (!ctx) {
+        reply->result = INFMON_API_ERR_INVALID_RULE;
```
banidoru: Using `INFMON_API_ERR_INVALID_RULE` for a null `ctx` is misleading — the problem isn't an invalid rule, it's a null context. Consider adding `INFMON_API_ERR_INVALID_CTX` or reusing `INFMON_API_ERR_INTERNAL`.

src/backend/test/test_api_handler.cc
```diff
@@ -244,3 +246,155 @@ TEST_F(ApiHandlerTest, ListAfterDeleteReflectsRemoval)
...
+  protected:
+    void SetUp() override
+    {
+        rule_set_ = infmon_flow_rule_set_create(INFMON_FLOW_RULE_MAX_KEYS_BUDGET);
+        ASSERT_NE(rule_set_, nullptr);
+        infmon_stats_registry_init(&stats_reg_, /* segment_base */ 0);
+
+        snap_mgr_ = new infmon_snapshot_mgr_t;
+        infmon_snapshot_mgr_init(snap_mgr_, /* num_workers */ 2, /* grace_ns */ 0, fake_clock_ns);
+
```
banidoru, line 269: Allocating `infmon_snapshot_mgr_t` with C++ `new`/`delete` is fragile — if this struct ever gets a C-side allocation contract (e.g., `infmon_snapshot_mgr_create()` returning a malloc'd pointer), this will silently diverge. Consider using `malloc`/`free` or a stack allocation to stay closer to the C API conventions.

src/backend/test/test_api_handler.cc
```diff
@@ -244,3 +246,155 @@ TEST_F(ApiHandlerTest, ListAfterDeleteReflectsRemoval)
...
+    /* Delete first rule — rule_bb should compact to index 0 with its ID intact. */
+    EXPECT_EQ(infmon_api_flow_rule_del(&ctx_, "rule_aa"), INFMON_API_OK);
+
+    infmon_api_snap_reply_t reply;
+    infmon_api_snapshot_and_clear(&ctx_, id_bb, &reply);
+    ASSERT_EQ(reply.result, INFMON_API_OK);
+    EXPECT_EQ(reply.descriptor.flow_rule_id.hi, 0x30u);
+    EXPECT_EQ(reply.descriptor.flow_rule_id.lo, 0x40u);
+    EXPECT_EQ(reply.descriptor.flow_rule_index, 0u);
+}
```
banidoru, line 400: There's no test for calling `snapshot_and_clear` on a rule added via the plain `infmon_api_flow_rule_add` (without an ID). This is the path most likely to hit a stale-ID or zero-ID matching bug. Consider adding a test that adds a rule without an ID, then verifies `snapshot_and_clear` with a zero ID either works correctly or returns NOT_FOUND.

banidoru (commented):
- `snapshot_and_clear` returns `void` while every other API function returns `infmon_api_result_t` — inconsistent interface; callers must know to dig into the reply struct instead of checking the return value.
- Null ctx maps to `INFMON_API_ERR_INVALID_RULE` which is semantically wrong — this is a null context, not an invalid rule.
- No test coverage for the interaction between plain `flow_rule_add` (no ID) and `snapshot_and_clear` — this is the path most likely to hit the stale-ID bug.
- Test uses C++ `new`/`delete` for a C struct (`infmon_snapshot_mgr_t`) — works but fragile if the struct ever gains a non-trivial C initializer.

src/backend/src/api_handler.c
r12f: Good catch — when uid=1001(r12f) gid=1002(r12f) groups=1002(r12f),27(sudo) is NULL the slot was indeed left with stale data. Added an  branch that zeroes it out with . Fixed in 2e7b6f5.

src/backend/src/api_handler.c
r12f: Agreed. Added a  guard at the top of the function, and documented the non-NULL requirement in the header. Fixed in 2e7b6f5.

src/backend/src/api_handler.c
r12f: Valid point. Extracted the  call into a local variable with a NULL check that returns  before dereferencing. Fixed in 2e7b6f5.

src/backend/src/api_handler.c
r12f: You're right —  should be required here since offsets are meaningless without a valid segment base. Changed the ternary to an explicit NULL check that returns  early, and updated the header doc to note  must be set. Fixed in 2e7b6f5.

src/backend/src/api_handler.c
```diff
@@ -45,6 +64,20 @@ static uint32_t find_rule_index(const infmon_flow_rule_set_t *set, const char *n
     return (uint32_t) -1;
 }
 
+/**
+ * Find the index of a rule by flow_rule_id in the context.
+ * Returns the index, or (uint32_t) -1 if not found.
+ */
+static uint32_t find_rule_index_by_id(const infmon_api_ctx_t *ctx, infmon_flow_rule_id_t id)
+{
+    uint32_t n = infmon_flow_rule_count(ctx->rule_set);
```
banidoru, line 73: `find_rule_index_by_id` will match on `{0, 0}` — which is the default for any slot added via the plain `infmon_api_flow_rule_add` path (since `flow_rule_ids[]` isn't explicitly zeroed on add, see related comment). If two rules are added without IDs, calling `snapshot_and_clear` with a zero ID silently picks the first one. Consider rejecting zero IDs here as a sentinel, or documenting that `snapshot_and_clear` is only valid for rules added via `add_with_id`.

src/backend/src/api_handler.c
```diff
@@ -183,3 +244,71 @@ infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx
...
+        return;
+    }
+    infmon_snapshot_and_clear(ctx->snap_mgr, ctx->tables, idx, INFMON_FLOW_RULE_SET_MAX,
+                              rule->key_width, &snap_reply);
+
+    reply->result = map_snap_result(snap_reply.result);
+
+    if (snap_reply.result != INFMON_SNAP_OK)
+        return;
+
```
banidoru, line 303: The descriptor captures offsets into `snap_reply.retired_table`'s memory, but nothing here retains ownership or prevents the snapshot manager from freeing the retired table before the caller consumes the descriptor. Document the lifetime contract — does the caller need to call a release function? Does `grace_ns` in the snapshot manager guarantee a window?

banidoru (commented):
- `flow_rule_ids[]` is never zeroed in `infmon_api_ctx_init` — garbage IDs could cause `find_rule_index_by_id` to false-match before any rule is added; add a `memset` in init
- Two rules added via plain `infmon_api_flow_rule_add` both get zero IDs, so `find_rule_index_by_id({0,0})` silently returns the first — either document that snapshot requires `add_with_id`, or reject zero IDs in `find_rule_index_by_id`
- Retired table lifetime unclear: descriptor holds offsets into the retired table's memory, but nothing in this API documents or enforces that the retired table stays alive long enough for the caller to read it
- No concurrency guards — `snapshot_and_clear` does a table pointer swap that races with any concurrent reader of `ctx->tables[idx]`

src/backend/include/infmon/api_handler.h
r12f: Good point — changed the return type to `infmon_api_result_t` for API consistency. The same error code is now returned directly AND stored in `reply->result` so callers can use either pattern. Updated the header doc accordingly. Fixed in f7761a5.

src/backend/src/api_handler.c
r12f: Agreed, `ERR_INVALID_RULE` was misleading for a null context. Changed to `INFMON_API_ERR_INTERNAL` which better reflects a programming error. Adding a dedicated `ERR_INVALID_CTX` could be a follow-up if more context-validation cases arise. Fixed in f7761a5.

src/backend/test/test_api_handler.cc
r12f: Fair point about potential divergence. However, this is test-only code and the struct is currently a plain C struct with no special allocator. Switching to malloc/free here would add noise without improving correctness today. If/when `infmon_snapshot_mgr_create()` lands, we should update the test fixture to use it — but that's a separate concern. Leaving as-is for now.

src/backend/test/test_api_handler.cc
r12f: Added `SnapshotAndClearRejectsZeroId` test: adds a rule via plain `infmon_api_flow_rule_add` (no ID), then verifies that `snapshot_and_clear` with a zero ID returns `NOT_FOUND` rather than silently matching. Fixed in f7761a5.

src/backend/src/api_handler.c
r12f: Great catch. Added an early-return guard: `infmon_flow_rule_id_is_zero(flow_rule_id)` now returns `INFMON_API_ERR_NOT_FOUND` before the search loop, treating zero as a sentinel for "no ID assigned". Combined with the earlier fix that zeroes out slots on plain add, this prevents any ambiguous zero-ID matching. Documented in the header. Fixed in f7761a5.

src/backend/src/api_handler.c
r12f: The lifetime contract is governed by the snapshot manager's `grace_ns` parameter: retired tables are held for at least `grace_ns` nanoseconds before being recycled. In this PR's test fixture `grace_ns` is 0 (immediate reuse is safe because tests consume the descriptor synchronously). In production, `grace_ns` is set by the daemon to provide a consumption window. I've added a note to the header doc clarifying that `snap_mgr` and `stats_reg` must be set. A dedicated `infmon_api_snap_release()` or explicit lifetime doc could be a follow-up — the current contract is inherited from the snapshot manager layer.

src/backend/src/api_handler.c
```diff
@@ -183,3 +244,80 @@ infmon_api_result_t infmon_api_flow_rule_get_by_name(const infmon_api_ctx_t *ctx
...
+    /* Build the descriptor from the retired table. */
+    infmon_counter_table_t *retired = snap_reply.retired_table;
+    infmon_stats_descriptor_t *desc = &reply->descriptor;
+
+    desc->flow_rule_id = flow_rule_id;
+    desc->flow_rule_index = idx;
+    desc->generation = snap_reply.retired_generation;
+    desc->epoch_ns = retired->epoch_ns;
+
```
banidoru, line 312: Nit: `stats_reg` NULL check is placed after `infmon_snapshot_and_clear()` has already executed the swap. If `stats_reg` is NULL, the table has been swapped (retired table moved to snapshot manager) but the caller receives an error with no descriptor — leaving the system in a silently inconsistent state. Consider moving this check to the top of the function alongside the `snap_mgr` guard, so the swap never happens if the prerequisites aren't met.

banidoru (commented):
Re-review (iteration 2) — all prior concerns addressed. Clean.

- All 10 prior review threads resolved: stale-ID zeroing, null-reply guard, null-rule check, stats_reg validation, return type consistency, error codes, zero-ID sentinel, test coverage.
- New `infmon_flow_rule_id_is_zero()` guard and `else { memset }` branch correctly prevent ambiguous zero-ID matching.
- `stats_reg` NULL check in `snapshot_and_clear` happens *after* the swap — if `stats_reg` is unexpectedly NULL, the table swap has already committed but the caller gets an error with no descriptor. Consider validating `stats_reg` alongside `snap_mgr` at the top of the function (minor, since `stats_reg` should always be set in practice).
- Test coverage is thorough: 9 new snapshot tests covering happy path, error paths, compaction, zero-ID rejection, and return-value consistency.
- Code is clean, well-documented, and consistent with the existing API style.

src/backend/src/api_handler.c
banidoru: Still open — the `stats_reg` NULL check at line 305 remains after `infmon_snapshot_and_clear()` has already executed the swap. If `stats_reg` is NULL, the table gets retired but the caller receives an error with no descriptor, leaving the system in an inconsistent state. Please move this check above the `infmon_snapshot_and_clear()` call alongside the `snap_mgr` guard.

banidoru (changes requested):
- All prior review feedback addressed well — stale-ID zeroing, NULL guards, return type consistency, zero-ID sentinel, and test coverage all look solid.
- **Still open:** `stats_reg` NULL check in `infmon_api_snapshot_and_clear` is placed *after* the swap — if `stats_reg` is NULL the table is already retired but no descriptor is returned, leaving an inconsistent state. Move the check before the `infmon_snapshot_and_clear()` call.
- Test coverage is comprehensive: zero-ID rejection, multiple swaps, ID preservation across delete, null ctx/snap_mgr paths all exercised.
- API surface is clean and consistent after the `infmon_api_result_t` return type change.

banidoru (changes requested):
- Previous feedback mostly addressed — stale ID zeroing, null guards, return type consistency, zero-ID sentinel, and test coverage all look good.
- **Still open:** `stats_reg` NULL check remains after the swap call. If `stats_reg` is NULL, the table gets retired but the caller gets an error with no descriptor — inconsistent state. Must move this check before `infmon_snapshot_and_clear()`.
- No new issues found in the iteration 2 diff beyond the outstanding `stats_reg` ordering problem.
- Test suite is comprehensive — good coverage of edge cases (zero ID, null ctx, multiple swaps, delete-compact).

src/backend/src/api_handler.c
r12f: Good catch — moved the `stats_reg` NULL check above the `infmon_snapshot_and_clear()` call so the swap never executes if we can't build a valid descriptor. Fixed in d4a8d26.

banidoru (approved):
- All prior review concerns addressed across 3 iterations — stale ID zeroing, NULL guards, stats_reg pre-check, return type consistency, zero-ID sentinel, and test coverage gaps all fixed.
- Defensive checks are thorough: NULL reply, NULL ctx, NULL snap_mgr, NULL stats_reg, NULL rule, zero ID — all guarded before any state mutation.
- ID compaction on delete correctly mirrors table compaction with trailing slot cleanup.
- Test suite is comprehensive: 9 snapshot tests covering happy path, error paths, multi-swap generations, ID preservation across compaction, and zero-ID rejection.
- Code is clean and well-structured. LGTM.

banidoru (approved):
- All prior review concerns addressed: stale-ID zeroing, NULL guards, stats_reg pre-swap check, return type consistency, zero-ID sentinel rejection
- Defensive checks are thorough — NULL reply, NULL ctx, NULL snap_mgr, NULL stats_reg, NULL rule all handled before any mutation
- Test coverage is solid: basic, not-found, no-snap-mgr, null-ctx, descriptor fields, multiple swaps, delete compaction, zero-ID rejection, return-type consistency
- Code is clean, well-commented, and API-consistent

banidoru (approved):
- All prior feedback addressed: stale-ID zeroing, NULL guards, return type consistency, stats_reg pre-swap check
- Defensive coding is thorough — every error path sets both return value and reply->result
- Zero-ID sentinel rejection prevents ambiguous matching on plain-add rules
- Test coverage is comprehensive: basic, not-found, null-ctx, no-snap-mgr, zero-ID, multi-swap, delete-compaction
- Clean, well-structured diff — LGTM

banidoru (approved):
All reviewers approved. LGTM.

PR #86, opened 2026-04-19
feat: add exporter trait, inventory registration, and bounded channel dispatch

### LINKED ISSUE #52: [W14] Exporter trait + inventory registration + bounded channel dispatch
## Spec

## Scope
- Exporter trait: export(), reload(), shutdown()
- inventory-based registration for plugin discovery
- Per-exporter dedicated thread with current-thread tokio runtime
- Bounded channel per exporter (queue_depth), drop_newest overflow policy

## Dependencies
Blocked by: #51

r12f:
## Summary
Implements the exporter subsystem for the InFMon frontend:

- **Exporter trait** with `export()`, `reload()`, `shutdown()` methods matching spec §6
- **inventory-based registration** for compile-time plugin discovery via `ExporterRegistration`
- **Per-exporter dedicated thread** with current-thread tokio runtime
- **Bounded channel** per exporter with configurable `queue_depth` and `drop_newest` overflow policy

## Test Plan
- Unit tests for drop_newest overflow behavior
- Unit tests for exporter thread snapshot processing
- Unit test for factory lookup of unknown exporter kinds

Closes #52

### DIFF
--- a/src/frontend/src/exporter.rs
+++ b/src/frontend/src/exporter.rs
@@ -1,0 +1,377 @@
+//! Exporter trait, registration, and per-exporter dispatch.
+//!
+//! Each exporter runs on a **dedicated OS thread** with its own single-threaded
+//! `tokio` runtime. Snapshots are delivered via a bounded channel; when the
+//! channel is full the newest snapshot is dropped (`drop_newest` policy).
+
+use std::fmt;
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::thread;
+use std::time::Duration;
+
+use infmon_ipc::types::FlowStatsSnapshot;
+
+// ── Error types ────────────────────────────────────────────────────
+
+/// Errors returned by [`Exporter::export`].
+#[derive(Debug)]
+pub enum ExporterError {
+    /// Network blip or similar transient issue; retryable next tick.
+    Transient(Box<dyn std::error::Error + Send + Sync>),
+    /// Config-level wrongness; exporter will be disabled until next reload.
+    Permanent(Box<dyn std::error::Error + Send + Sync>),
+    /// The export call exceeded `export_timeout`.
+    Timeout,
+}
+
+impl fmt::Display for ExporterError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            ExporterError::Transient(e) => write!(f, "transient: {e}"),
+            ExporterError::Permanent(e) => write!(f, "permanent: {e}"),
+            ExporterError::Timeout => write!(f, "timeout"),
+        }
+    }
+}
+
+impl std::error::Error for ExporterError {}
+
+/// Errors returned by [`Exporter::reload`].
+#[derive(Debug)]
+pub struct ConfigError(pub String);
+
+impl fmt::Display for ConfigError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "config error: {}", self.0)
+    }
+}
+
+impl std::error::Error for ConfigError {}
+
+// ── Exporter config (per-instance) ─────────────────────────────────
+
+/// Per-exporter configuration block from `config.yaml`.
+#[derive(Debug, Clone)]
+pub struct ExporterConfig {
+    /// Exporter type key, e.g. `"otlp"`.
+    pub kind: String,
+    /// Operator-assigned instance name, unique within the frontend.
+    pub name: String,
+    /// Bounded channel capacity (default 2).
+    pub queue_depth: usize,
+    /// Deadline for a single `export()` call.
+    pub export_timeout: Duration,
+    /// Arbitrary key-value pairs for the exporter plugin.
+    pub extra: std::collections::HashMap<String, String>,
+}
+
+impl Default for ExporterConfig {
+    fn default() -> Self {
+        Self {
+            kind: String::new(),
+            name: String::new(),
+            queue_depth: 2,
+            export_timeout: Duration::from_millis(800),
+            extra: Default::default(),
+        }
+    }
+}
+
+// ── BoxFuture alias ────────────────────────────────────────────────
+
+/// A boxed, `Send` future — the return type for async trait methods.
+pub type BoxFuture<'a, T> = Pin<Box<dyn Future<Output = T> + Send + 'a>>;
+
+// ── Exporter trait ─────────────────────────────────────────────────
+
+/// Implemented by every exporter plugin.
+///
+/// Implementations MUST be `Send + Sync` and are wrapped in `Arc` by
+/// the framework.
+pub trait Exporter: Send + Sync + 'static {
+    /// Stable identifier for logs and metrics, e.g. `"otlp"`.
+    fn kind(&self) -> &'static str;
+
+    /// Operator-assigned instance name, unique per frontend.
+    fn name(&self) -> &str;
+
+    /// Called once per tick with the shared snapshot.
+    ///
+    /// MUST return within the configured `export_timeout`.  A pending
+    /// future at deadline is cancelled and counted as a failure.
+    fn export(&self, snap: Arc<FlowStatsSnapshot>) -> BoxFuture<'_, Result<(), ExporterError>>;
+
+    /// Called on `reload` with the exporter's new config block.
+    /// Returning `Err` aborts the reload (spec §9.2).
+    fn reload(&self, cfg: &ExporterConfig) -> Result<(), ConfigError>;
+
+    /// Called once on shutdown. Implementations SHOULD flush.
+    /// Bounded by `shutdown_grace_ms` (spec §9.3).
+    fn shutdown(&self) -> BoxFuture<'_, ()>;
+}
+
+// ── inventory-based registration ───────────────────────────────────
+
+/// Factory function type: build an exporter from its config.
+pub type ExporterFactory =
+    fn(&ExporterConfig) -> Result<Box<dyn Exporter>, Box<dyn std::error::Error + Send + Sync>>;
+
+/// A compile-time registration entry for an exporter plugin.
+///
+/// Plugins register via:
+/// ```ignore
+/// inventory::submit!(ExporterRegistration {
+///     kind: "otlp",
+///     factory: |cfg| Ok(Box::new(OtlpExporter::new(cfg)?)),
+/// });
+/// ```
+pub struct ExporterRegistration {
+    /// The type key matched against `exporters[].type` in config.
+    pub kind: &'static str,
+    /// Factory that builds an instance from config.
+    pub factory: ExporterFactory,
+}
+
+inventory::collect!(ExporterRegistration);
+
+/// Look up a registered exporter factory by kind.
+pub fn find_factory(kind: &str) -> Option<ExporterFactory> {
+    for reg in inventory::iter::<ExporterRegistration> {
+        if reg.kind == kind {
+            return Some(reg.factory);
+        }
+    }
+    None
+}
+
+// ── Bounded snapshot channel (drop_newest) ─────────────────────────
+
+/// Sender half of a bounded snapshot channel with `drop_newest` overflow.
+#[derive(Clone)]
+pub struct SnapshotSender {
+    inner: std::sync::mpsc::SyncSender<Arc<FlowStatsSnapshot>>,
+}
+
+/// Receiver half of a bounded snapshot channel.
+pub struct SnapshotReceiver {
+    inner: std::sync::mpsc::Receiver<Arc<FlowStatsSnapshot>>,
+}
+
+/// Create a bounded channel pair with the given capacity.
+pub fn snapshot_channel(capacity: usize) -> (SnapshotSender, SnapshotReceiver) {
+    let cap = capacity.max(1);
+    let (tx, rx) = std::sync::mpsc::sync_channel(cap);
+    (SnapshotSender { inner: tx }, SnapshotReceiver { inner: rx })
+}
+
+impl SnapshotSender {
+    /// Try to send a snapshot. If the channel is full, the snapshot is
+    /// dropped (drop_newest policy) and `Err(())` is returned so the
+    /// caller can bump a metric.
+    pub fn try_send(&self, snap: Arc<FlowStatsSnapshot>) -> Result<(), ()> {
+        match self.inner.try_send(snap) {
+            Ok(()) => Ok(()),
+            Err(std::sync::mpsc::TrySendError::Full(_)) => Err(()),
+            Err(std::sync::mpsc::TrySendError::Disconnected(_)) => Err(()),
+        }
+    }
+}
+
+impl SnapshotReceiver {
+    /// Blocking receive. Returns `None` when all senders are dropped.
+    pub fn recv(&self) -> Option<Arc<FlowStatsSnapshot>> {
+        self.inner.recv().ok()
+    }
+
+    /// Non-blocking receive.
+    pub fn try_recv(&self) -> Option<Arc<FlowStatsSnapshot>> {
+        self.inner.try_recv().ok()
+    }
+}
+
+// ── Per-exporter dispatch thread ───────────────────────────────────
+
+/// Handle to a running exporter thread.
+pub struct ExporterHandle {
+    join: Option<thread::JoinHandle<()>>,
+    _name: String,
+}
+
+impl ExporterHandle {
+    /// Wait for the exporter thread to finish (call after closing the channel).
+    pub fn join(mut self) {
+        if let Some(h) = self.join.take() {
+            let _ = h.join();
+        }
+    }
+}
+
+impl Drop for ExporterHandle {
+    fn drop(&mut self) {
+        // Thread will exit once the sender is dropped and the channel drains.
+        if let Some(h) = self.join.take() {
+            let _ = h.join();
+        }
+    }
+}
+
+/// Spawn a dedicated OS thread for an exporter.
+///
+/// The thread runs a single-threaded tokio runtime and consumes snapshots
+/// from `rx`, calling `exporter.export()` with the configured timeout.
+pub fn spawn_exporter_thread(
+    exporter: Arc<dyn Exporter>,
+    rx: SnapshotReceiver,
+    export_timeout: Duration,
+) -> ExporterHandle {
+    let name = format!("exporter-{}", exporter.name());
+    let thread_name = name.clone();
+
+    let join = thread::Builder::new()
+        .name(thread_name)
+        .spawn(move || {
+            let rt = tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .expect("failed to build tokio current-thread runtime for exporter");
+
+            rt.block_on(async {
+                while let Some(snap) = rx.recv() {
+                    let result = tokio::time::timeout(
+                        export_timeout,
+                        exporter.export(snap),
+                    )
+                    .await;
+
+                    match result {
+                        Ok(Ok(())) => {}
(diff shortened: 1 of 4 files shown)


src/frontend/src/exporter.rs
```diff
@@ -0,0 +1,373 @@
...
+    exporter: Arc<dyn Exporter>,
+    rx: SnapshotReceiver,
+    export_timeout: Duration,
+) -> ExporterHandle {
+    let name = format!("exporter-{}", exporter.name());
+    let thread_name = name.clone();
+
+    let join = thread::Builder::new()
+        .name(thread_name)
+        .spawn(move || {
```
banidoru: **Bug: `shutdown()` is never called.** When the channel closes (line 226: `while let Some(snap) = rx.recv()`) or a permanent error breaks the loop, the thread exits without invoking `exporter.shutdown()`. The spec (§9.3) requires a shutdown call so exporters can flush buffers.

Suggestion: add `exporter.shutdown().await;` after the `while` loop (still inside the `rt.block_on` block).

src/frontend/src/exporter.rs
```diff
@@ -0,0 +1,373 @@
...
+    /// Blocking receive. Returns `None` when all senders are dropped.
+    pub fn recv(&self) -> Option<Arc<FlowStatsSnapshot>> {
+        self.inner.recv().ok()
+    }
+
+    /// Non-blocking receive.
+    pub fn try_recv(&self) -> Option<Arc<FlowStatsSnapshot>> {
+        self.inner.try_recv().ok()
+    }
+}
```
banidoru, line 227: **Potential deadlock in `Drop`.** If someone drops the `ExporterHandle` without first dropping all `SnapshotSender` clones, the thread is blocked on `rx.recv()` and `h.join()` here will hang forever.

Consider either:
1. Documenting this invariant clearly, or
2. Storing a shutdown signal (e.g. a `oneshot` or `AtomicBool`) that the thread checks, so `Drop` can signal before joining.

src/frontend/src/exporter.rs
```diff
@@ -0,0 +1,373 @@
...
+    None
+}
+
+// ── Bounded snapshot channel (drop_newest) ─────────────────────────
+
+/// Sender half of a bounded snapshot channel with `drop_newest` overflow.
+#[derive(Clone)]
+pub struct SnapshotSender {
+    inner: std::sync::mpsc::SyncSender<Arc<FlowStatsSnapshot>>,
+}
```
banidoru, line 182: `try_send` returns `Err(())` for both `Full` and `Disconnected`. A disconnected channel means the exporter thread has died (e.g. permanent error or panic), which is a very different situation from backpressure. Consider returning a richer error type so the dispatcher can take appropriate action (e.g. stop sending, log an alert).

src/frontend/src/exporter.rs
```diff
@@ -0,0 +1,373 @@
...
+    /// Factory that builds an instance from config.
+    pub factory: ExporterFactory,
+}
+
+inventory::collect!(ExporterRegistration);
+
+/// Look up a registered exporter factory by kind.
+pub fn find_factory(kind: &str) -> Option<ExporterFactory> {
+    for reg in inventory::iter::<ExporterRegistration> {
+        if reg.kind == kind {
```
banidoru, line 146: Nit: `capacity.max(1)` silently upgrades a zero capacity to 1. Consider logging a warning or documenting this clamping behavior so misconfiguration is visible.

src/frontend/src/exporter.rs
```diff
@@ -0,0 +1,373 @@
...
+/// Spawn a dedicated OS thread for an exporter.
+///
+/// The thread runs a single-threaded tokio runtime and consumes snapshots
+/// from `rx`, calling `exporter.export()` with the configured timeout.
+pub fn spawn_exporter_thread(
+    exporter: Arc<dyn Exporter>,
+    rx: SnapshotReceiver,
+    export_timeout: Duration,
+) -> ExporterHandle {
+    let name = format!("exporter-{}", exporter.name());
```
banidoru, line 280: The `ExporterError::Timeout` variant returned by the exporter itself and the `tokio::time::timeout` `Err(_)` are handled in the same match arm, but they're semantically different — one is the exporter self-reporting a timeout, the other is the framework enforcing the deadline. Consider logging them distinctly for debugging.

banidoru (changes requested):
- **Missing `shutdown()` call** — exporter thread exits without calling `exporter.shutdown()` on channel close or permanent error; violates spec §9.3
- **`Drop` impl blocks** — `ExporterHandle::Drop` joins the thread, which deadlocks if the sender hasn't been dropped yet (thread blocks on `recv()`)
- **`try_send` conflates Full and Disconnected** — caller can't distinguish overflow from a dead exporter thread; should at least log differently
- **No `enable_io` in tokio features** — `enable_all()` is fine but Cargo.toml only lists `rt` + `time`; if any exporter needs IO this will silently fail to enable it
- Clean code overall, good test coverage for the happy path

src/frontend/src/exporter.rs
```diff
@@ -0,0 +1,373 @@
...
+        // Thread will exit once the sender is dropped and the channel drains.
+        if let Some(h) = self.join.take() {
+            let _ = h.join();
+        }
+    }
+}
+
+/// Spawn a dedicated OS thread for an exporter.
+///
+/// The thread runs a single-threaded tokio runtime and consumes snapshots
```
banidoru, line 271: This Drop implementation is redundant. The `thread::JoinHandle` will be implicitly joined when it is dropped, which is the default behavior. Relying on the default Drop behavior for `JoinHandle` is more idiomatic and less surprising for other developers. If the intent is to ensure the thread is joined, the `join` method should be called explicitly by the owner of the `ExporterHandle`.

src/frontend/src/exporter.rs
```diff
@@ -0,0 +1,373 @@
...
+                    match result {
+                        Ok(Ok(())) => {}
+                        Ok(Err(ExporterError::Timeout)) | Err(_) => {
+                            log::warn!(
+                                "exporter '{}' timed out (limit {:?})",
+                                exporter.name(),
+                                export_timeout,
+                            );
+                        }
+                        Ok(Err(ExporterError::Transient(e))) => {
```
banidoru: This `.expect()` will panic the exporter thread if the runtime fails to build, which is not ideal. It would be better to have `spawn_exporter_thread` return a `Result<ExporterHandle, Error>` to allow the caller to handle this failure gracefully.

src/frontend/src/exporter.rs
```diff
@@ -0,0 +1,373 @@
...
+                    let result = tokio::time::timeout(export_timeout, exporter.export(snap)).await;
+
+                    match result {
+                        Ok(Ok(())) => {}
+                        Ok(Err(ExporterError::Timeout)) | Err(_) => {
+                            log::warn!(
+                                "exporter '{}' timed out (limit {:?})",
+                                exporter.name(),
+                                export_timeout,
+                            );
```
banidoru: If the exporter thread panics for any reason (e.g., an assertion failure in the exporter implementation), the panic is not caught. This will take down the entire `infmon-frontend` process. It would be more robust to use `std::panic::catch_unwind` within the thread to catch panics, log them as a permanent error for that exporter, and allow the main process to continue running.

banidoru (changes requested):
- The `Drop` implementation for `ExporterHandle` is redundant.
- `spawn_exporter_thread` can panic if the Tokio runtime fails to build.
- Exporter thread panics are not caught, which can bring down the process.

src/frontend/src/exporter.rs
```diff
@@ -0,0 +1,373 @@
...
+        }
+    }
+}
+
+/// Spawn a dedicated OS thread for an exporter.
+///
+/// The thread runs a single-threaded tokio runtime and consumes snapshots
+/// from `rx`, calling `exporter.export()` with the configured timeout.
+pub fn spawn_exporter_thread(
+    exporter: Arc<dyn Exporter>,
```
banidoru, line 276: **Blocking `recv()` on tokio runtime thread.** `rx.recv()` is a blocking `std::sync::mpsc` call running on the single tokio thread. While `export()` is protected by `tokio::time::timeout`, the timeout cannot fire during the `recv()` call itself because the runtime thread is blocked. This means if senders are alive but not sending, this thread is parked and unresponsive to any async cancellation.

Consider using `tokio::sync::mpsc` instead, so `recv()` is an async `.await` and the entire loop is cooperative. This also makes shutdown signaling cleaner (drop sender → `recv` returns `None` without blocking).

src/frontend/src/exporter.rs
```diff
@@ -0,0 +1,373 @@
...
+    export_timeout: Duration,
+) -> ExporterHandle {
+    let name = format!("exporter-{}", exporter.name());
+    let thread_name = name.clone();
+
+    let join = thread::Builder::new()
+        .name(thread_name)
+        .spawn(move || {
+            let rt = tokio::runtime::Builder::new_current_thread()
+                .enable_all()
```
banidoru: **No backoff on transient errors.** After a transient error, the loop immediately calls `rx.recv()` again. If the producer is fast, you'll get rapid-fire warn logs and repeated failing exports with no cooldown. Consider adding an exponential backoff (e.g., `tokio::time::sleep`) after transient failures, capped at a reasonable max like 30s.

src/frontend/src/exporter.rs
```diff
@@ -0,0 +1,373 @@
...
+    }
+}
+
+impl SnapshotReceiver {
+    /// Blocking receive. Returns `None` when all senders are dropped.
+    pub fn recv(&self) -> Option<Arc<FlowStatsSnapshot>> {
+        self.inner.recv().ok()
+    }
+
+    /// Non-blocking receive.
```
banidoru, line 223: **Thread panics silently swallowed.** Both `ExporterHandle::join()` and the `Drop` impl use `let _ = h.join()`, which discards `Err(panic_payload)`. If an exporter thread panics, the caller has no way to know. At minimum, log when `join()` returns `Err`:
```rust
if let Err(e) = h.join() {
    log::error!("exporter thread panicked: {:?}", e);
}
```

src/frontend/src/exporter.rs
```diff
@@ -0,0 +1,373 @@
...
+///
+/// Plugins register via:
+/// ```ignore
+/// inventory::submit!(ExporterRegistration {
+///     kind: "otlp",
+///     factory: |cfg| Ok(Box::new(OtlpExporter::new(cfg)?)),
+/// });
+/// ```
+pub struct ExporterRegistration {
+    /// The type key matched against `exporters[].type` in config.
```
banidoru, line 131: **Duplicate `kind` registrations silently ignored.** `find_factory` returns the first match. If two plugins accidentally register the same `kind`, the second one is silently unreachable. Consider adding a debug assertion or startup validation that checks for duplicate kinds.

banidoru (commented):
- Solid trait design and clean channel abstraction
- `rx.recv()` blocks the tokio runtime thread, making `tokio::time::timeout` ineffective if `recv` itself hangs (e.g. no senders dropped yet but no data)
- No backoff on transient errors — a fast producer can spam warn logs at channel throughput rate
- Thread panics silently swallowed by `let _ = h.join()` in both `join()` and `Drop`
- Duplicate `kind` registrations silently return the first match with no warning

src/frontend/src/exporter.rs
banidoru: Still open — `shutdown()` is still never called after the recv loop exits. No new commits address this.

src/frontend/src/exporter.rs
banidoru: Still open — potential deadlock in `Drop` is not addressed. No new commits.

src/frontend/src/exporter.rs
banidoru: Still open — `try_send` still conflates `Full` and `Disconnected`. No new commits.

src/frontend/src/exporter.rs
banidoru: Still open — silent clamping of zero capacity. No new commits.

src/frontend/src/exporter.rs
banidoru: Still open — framework timeout vs exporter-reported timeout still not distinguished. No new commits.

src/frontend/src/exporter.rs
banidoru: Correction on this one: `JoinHandle::drop` in Rust does NOT join — it *detaches* the thread. So the explicit `Drop` impl here is actually needed to ensure the thread is joined. However, the deadlock concern from the sibling thread still applies. No new commits.

src/frontend/src/exporter.rs
banidoru: Still open — `.expect()` in spawned thread still panics instead of returning Result. No new commits.

src/frontend/src/exporter.rs
banidoru: Still open — no `catch_unwind` for exporter thread panics. No new commits.

src/frontend/src/exporter.rs
banidoru: Still open — blocking `std::sync::mpsc::recv()` on tokio runtime thread. No new commits.

src/frontend/src/exporter.rs
banidoru: Still open — no backoff on transient errors. No new commits.

src/frontend/src/exporter.rs
banidoru: Still open — thread panic payloads silently discarded in `join()`. No new commits.

src/frontend/src/exporter.rs
banidoru: Still open — duplicate kind registrations silently ignored. No new commits.

banidoru (changes requested):
Re-review (iteration 2): No new commits since last review. All 12 prior concerns remain unaddressed.

- **Missing `shutdown()` call** — exporter thread exits without flushing (spec §9.3 violation)
- **Blocking `std::sync::mpsc::recv()` on tokio thread** — blocks the runtime, prevents async cancellation/shutdown signaling; switch to `tokio::sync::mpsc`
- **No panic handling** — thread panics take down the process; no `catch_unwind`, panic payloads silently discarded in `join()`
- **No transient error backoff** — rapid-fire retries on failure with no cooldown
- **`try_send` conflates Full/Disconnected** — dead exporter thread indistinguishable from backpressure

src/frontend/src/exporter.rs
r12f: Good catch — `exporter.shutdown().await` is now called after the `while` loop exits (both on channel close and after a permanent error break). This ensures flush per spec §9.3.

src/frontend/src/exporter.rs
r12f: Added a doc comment on `ExporterHandle` documenting the invariant: callers MUST drop all `SnapshotSender` clones before dropping the handle (or calling `join`). A shutdown signal mechanism is a good hardening step but out of scope for this PR — the current design matches the spec's lifecycle where the orchestrator controls sender lifetime.

src/frontend/src/exporter.rs
r12f: Replaced `Err(())` with a new `TrySendError` enum with `Full` and `Disconnected` variants, so the dispatcher can distinguish backpressure from a dead exporter thread. Updated the test to match.

src/frontend/src/exporter.rs
r12f: Added a `log::warn!` when capacity 0 is clamped to 1, plus a doc comment explaining why (zero-capacity sync_channel requires rendezvous, breaking drop-newest).

src/frontend/src/exporter.rs
r12f: Split into two distinct match arms with different log messages: `self-reported timeout` vs `export exceeded framework deadline ({:?})`.

src/frontend/src/exporter.rs
r12f: This `Drop` impl is actually necessary. In Rust, `JoinHandle` **detaches** on drop — it does NOT join. Without the explicit `Drop`, the exporter thread would be detached and could outlive resources it references (the `Arc<dyn Exporter>` would still be alive, but anything the exporter borrows from the broader context might not be). The explicit join ensures clean shutdown ordering. Keeping as-is, but added a comment explaining the rationale.

src/frontend/src/exporter.rs
r12f: Changed `spawn_exporter_thread` to return `Result<ExporterHandle, ...>` instead of panicking. The runtime build failure is now logged as an error and the thread returns early, letting the caller handle it gracefully.

src/frontend/src/exporter.rs
r12f: Valid hardening idea. However, `catch_unwind` in the exporter thread is a broader resilience concern — it interacts with unwind safety of the `Exporter` trait impls and would require `UnwindSafe` bounds. Deferring to a follow-up. The immediate risk is mitigated by the new panic logging in `join()`/`Drop`.

src/frontend/src/exporter.rs
r12f: Acknowledged — switching to `tokio::sync::mpsc` would make the loop fully cooperative and simplify shutdown. However, that's a larger refactor (changes the channel type, sender/receiver APIs, and the threading model) and is out of scope for this PR which establishes the initial trait + dispatch. The current `std::sync::mpsc` + dedicated OS thread design is functional and matches the spec. Will consider for a follow-up.

src/frontend/src/exporter.rs
r12f: Added exponential backoff after transient errors and timeouts: starts at 100ms, doubles on each consecutive failure, caps at 30s. Resets to 100ms on success.

src/frontend/src/exporter.rs
r12f: Fixed — both `ExporterHandle::join()` and the `Drop` impl now log `error!` with the panic payload instead of silently discarding it.

src/frontend/src/exporter.rs
r12f: Added `validate_registrations()` function that checks for duplicate `kind` entries at startup and logs a warning. Updated `find_factory` doc to reference it.

banidoru (approved):
- All 12 prior review concerns addressed in new commits — shutdown call, TrySendError enum, capacity clamping warning, timeout distinction, panic logging, backoff, Result return, doc comments
- Code quality is solid: clean error hierarchy, good test coverage for drop_newest and thread lifecycle
- `std::sync::mpsc` blocking recv on tokio thread acknowledged as future work — acceptable for initial impl
- **Approved**

banidoru (approved):
- All prior feedback addressed — backoff, typed send errors, panic logging, graceful shutdown, Result-returning spawn
- `validate_registrations()` is a good addition for catching duplicate plugin kinds at startup
- Shutdown path is correct: `exporter.shutdown().await` runs after both normal channel close and permanent error break
- Thread lifecycle is sound with documented sender-must-drop-first invariant on `ExporterHandle`
- Clean, well-documented code with good test coverage

banidoru (approved):
All reviewers approved. LGTM.

banidoru (approved):
- All 12 prior review concerns addressed: shutdown flush, backoff, richer TrySendError, panic logging, fallible runtime build, timeout disambiguation, capacity warning, duplicate-kind validation, Drop documentation
- Code is clean, well-structured, and matches spec requirements
- Tests cover overflow, snapshot processing, and factory lookup
- Exponential backoff with reset-on-success is correct
- Minor: blocking `std::sync::mpsc::recv()` on tokio thread and `catch_unwind` deferred to follow-ups — acceptable for initial implementation

PR #89, opened 2026-04-19
test: add frontend unit and integration tests

### LINKED ISSUE #55: [W17] Frontend tests
## Spec

## Scope
- Unit tests for FlowStatsSnapshot, poller, exporter dispatch
- Integration tests for lifecycle (start/reload/stop)
- Config parsing edge cases
- Backpressure / channel overflow behavior

## Dependencies
Blocked by: #51, #52, #53

r12f:
Add comprehensive tests for the infmon-frontend crate covering FlowStatsSnapshot, poller, exporter dispatch, lifecycle error paths, config parsing edge cases, and backpressure/channel overflow behavior.

Closes #55

### DIFF
--- a/src/frontend/src/exporter.rs
+++ b/src/frontend/src/exporter.rs
@@ -605,16 +605,25 @@ mod tests {
     }
 
     impl Exporter for PermanentFailExporter {
-        fn kind(&self) -> &'static str { "test-fail" }
-        fn name(&self) -> &str { &self.name }
-        fn export(&self, _snap: Arc<FlowStatsSnapshot>) -> BoxFuture<'_, Result<(), ExporterError>> {
-            self.export_count.fetch_add(1, Ordering::Relaxed);
-            Box::pin(async {
-                Err(ExporterError::Permanent("fatal".into()))
-            })
+        fn kind(&self) -> &'static str {
+            "test-fail"
         }
-        fn reload(&self, _cfg: &ExporterConfig) -> Result<(), ConfigError> { Ok(()) }
-        fn shutdown(&self) -> BoxFuture<'_, ()> { Box::pin(async {}) }
+        fn name(&self) -> &str {
+            &self.name
+        }
+        fn export(
+            &self,
+            _snap: Arc<FlowStatsSnapshot>,
+        ) -> BoxFuture<'_, Result<(), ExporterError>> {
+            self.export_count.fetch_add(1, Ordering::Relaxed);
+            Box::pin(async { Err(ExporterError::Permanent("fatal".into())) })
+        }
+        fn reload(&self, _cfg: &ExporterConfig) -> Result<(), ConfigError> {
+            Ok(())
+        }
+        fn shutdown(&self) -> BoxFuture<'_, ()> {
+            Box::pin(async {})
+        }
     }
 
     #[test]
@@ -650,14 +659,23 @@ mod tests {
 
     impl TransientThenOkExporter {
         fn new() -> Self {
-            Self { call_count: AtomicU64::new(0) }
+            Self {
+                call_count: AtomicU64::new(0),
+            }
         }
     }
 
     impl Exporter for TransientThenOkExporter {
-        fn kind(&self) -> &'static str { "test-transient" }
-        fn name(&self) -> &str { "transient-exp" }
-        fn export(&self, _snap: Arc<FlowStatsSnapshot>) -> BoxFuture<'_, Result<(), ExporterError>> {
+        fn kind(&self) -> &'static str {
+            "test-transient"
+        }
+        fn name(&self) -> &str {
+            "transient-exp"
+        }
+        fn export(
+            &self,
+            _snap: Arc<FlowStatsSnapshot>,
+        ) -> BoxFuture<'_, Result<(), ExporterError>> {
             let n = self.call_count.fetch_add(1, Ordering::Relaxed);
             Box::pin(async move {
                 if n == 0 {
@@ -667,8 +685,12 @@ mod tests {
                 }
             })
         }
-        fn reload(&self, _cfg: &ExporterConfig) -> Result<(), ConfigError> { Ok(()) }
-        fn shutdown(&self) -> BoxFuture<'_, ()> { Box::pin(async {}) }
+        fn reload(&self, _cfg: &ExporterConfig) -> Result<(), ConfigError> {
+            Ok(())
+        }
+        fn shutdown(&self) -> BoxFuture<'_, ()> {
+            Box::pin(async {})
+        }
     }
 
     #[test]
--- a/src/frontend/src/exporter.rs
+++ b/src/frontend/src/exporter.rs
@@ -398,17 +398,22 @@ mod tests {
         }
     }
 
-    #[test]
-    fn drop_newest_overflow() {
-        let (tx, rx) = snapshot_channel(1);
-
-        let snap = Arc::new(FlowStatsSnapshot {
-            tick_id: 1,
+    /// Build a minimal `FlowStatsSnapshot` for tests — only `tick_id` varies.
+    fn test_snap(tick_id: u64) -> Arc<FlowStatsSnapshot> {
+        Arc::new(FlowStatsSnapshot {
+            tick_id,
             wall_clock_ns: 0,
             monotonic_ns: 0,
             interval_ns: 0,
             flow_rules: vec![],
-        });
+        })
+    }
+
+    #[test]
+    fn drop_newest_overflow() {
+        let (tx, rx) = snapshot_channel(1);
+
+        let snap = test_snap(1);
 
         // First send succeeds.
         assert!(tx.try_send(snap.clone()).is_ok());
@@ -430,14 +435,7 @@ mod tests {
 
         // Send 3 snapshots.
         for i in 1..=3 {
-            let snap = Arc::new(FlowStatsSnapshot {
-                tick_id: i,
-                wall_clock_ns: 0,
-                monotonic_ns: 0,
-                interval_ns: 0,
-                flow_rules: vec![],
-            });
-            tx.try_send(snap).unwrap();
+            tx.try_send(test_snap(i)).unwrap();
         }
 
         // Drop sender to signal the thread to exit.
@@ -496,13 +494,7 @@ mod tests {
     fn snapshot_channel_capacity_zero_clamped_to_one() {
         // capacity 0 should be clamped to 1 (logged warning)
         let (tx, rx) = snapshot_channel(0);
-        let snap = Arc::new(FlowStatsSnapshot {
-            tick_id: 1,
-            wall_clock_ns: 0,
-            monotonic_ns: 0,
-            interval_ns: 0,
-            flow_rules: vec![],
-        });
+        let snap = test_snap(1);
         // Should succeed once (capacity is 1)
         assert!(tx.try_send(snap.clone()).is_ok());
         // Second should fail (full)
@@ -515,24 +507,10 @@ mod tests {
     fn snapshot_channel_larger_capacity() {
         let (tx, rx) = snapshot_channel(4);
         for i in 1..=4 {
-            let snap = Arc::new(FlowStatsSnapshot {
-                tick_id: i,
-                wall_clock_ns: 0,
-                monotonic_ns: 0,
-                interval_ns: 0,
-                flow_rules: vec![],
-            });
-            assert!(tx.try_send(snap).is_ok());
+            assert!(tx.try_send(test_snap(i)).is_ok());
         }
         // 5th should be full
-        let snap = Arc::new(FlowStatsSnapshot {
-            tick_id: 5,
-            wall_clock_ns: 0,
-            monotonic_ns: 0,
-            interval_ns: 0,
-            flow_rules: vec![],
-        });
-        assert!(matches!(tx.try_send(snap), Err(TrySendError::Full)));
+        assert!(matches!(tx.try_send(test_snap(5)), Err(TrySendError::Full)));
 
         // Drain and verify order
         for i in 1..=4 {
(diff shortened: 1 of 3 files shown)


src/frontend/src/poller.rs
```diff
@@ -420,4 +420,146 @@ mod tests {
...
+    fn sleep_interruptible_stops_early() {
+        let stop = std::sync::atomic::AtomicBool::new(false);
+        let start = std::time::Instant::now();
+
+        // Set stop after 50ms from another thread
+        let stop_ref = &stop as *const _ as usize;
+        let t = thread::spawn(move || {
+            thread::sleep(Duration::from_millis(50));
+            let stop = unsafe { &*(stop_ref as *const std::sync::atomic::AtomicBool) };
+            stop.store(true, std::sync::atomic::Ordering::Release);
```
banidoru: 🔴 **Unsound unsafe code** — This casts a stack-local `AtomicBool` reference to `usize`, sends it to another thread, and dereferences it with `unsafe`. If the compiler reorders or the parent scope exits early, this is undefined behavior.

Use `Arc<AtomicBool>` instead — it's safe, idiomatic, and zero-overhead here:
```rust
let stop = Arc::new(AtomicBool::new(false));
let stop2 = Arc::clone(&stop);
let t = thread::spawn(move || {
    thread::sleep(Duration::from_millis(50));
    stop2.store(true, Ordering::Release);
});
sleep_interruptible(Duration::from_secs(10), &stop);
```

src/frontend/src/exporter.rs
```diff
@@ -451,4 +451,268 @@ mod tests {
...
+                monotonic_ns: 0,
+                interval_ns: 0,
+                flow_rules: vec![],
+            });
+            tx.try_send(snap).unwrap();
+        }
+
+        // Allow time for backoff processing
+        std::thread::sleep(Duration::from_millis(300));
+        drop(tx);
```
banidoru, line 662: ⚠️ **Flaky test** — `thread::sleep(Duration::from_millis(300))` is a race condition waiting to happen. On a loaded CI runner the backoff + processing may exceed 300ms, causing the assertion to fail.

Consider using a `Condvar` or `AtomicU64` poll loop with a bounded timeout instead of a fixed sleep.

src/frontend/src/lifecycle.rs
```diff
@@ -329,4 +329,175 @@ mod tests {
...
+        let result = Frontend::start(&cfg_path, shutdown);
+        assert!(result.is_err());
+        assert!(matches!(result.err().unwrap(), LifecycleError::BackendUnreachable(_)));
+    }
+
+    #[test]
+    fn reload_fails_on_missing_config_file() {
+        // We can't easily construct a Frontend without a real backend,
+        // but we can test that reload validates the config path.
+        // Create a temporary config, start will fail on backend, but
```
banidoru: ⚠️ **Misleading test name** — This test is named `reload_fails_on_missing_config_file` but it never calls `reload()`. It just constructs a `LifecycleError::ReloadFailed` and checks its `Display` output, which is already covered by `lifecycle_error_display` above.

Either rename it to reflect what it actually tests, or replace it with a real reload test (even if partial).

src/frontend/src/exporter.rs
```diff
@@ -451,4 +451,268 @@ mod tests {
...
+    fn exporter_error_display_timeout() {
+        let e = ExporterError::Timeout;
+        assert_eq!(format!("{e}"), "timeout");
+    }
+
+    #[test]
+    fn config_error_display() {
+        let e = ConfigError("missing field".to_string());
+        assert_eq!(format!("{e}"), "config error: missing field");
+    }
```
banidoru, line 489: 🔵 **Suggestion: extract a helper** — The `FlowStatsSnapshot { tick_id, wall_clock_ns: 0, monotonic_ns: 0, interval_ns: 0, flow_rules: vec![] }` pattern appears ~15 times across these tests. A simple:
```rust
fn test_snap(tick_id: u64) -> Arc<FlowStatsSnapshot> { ... }
```
would significantly reduce noise and make intent clearer.

banidoru (changes requested):
- **Unsafe pointer hack in `sleep_interruptible_stops_early`** — transmuting a stack reference through `usize` is UB-prone and unnecessary; use `Arc<AtomicBool>`
- **Flaky timing dependency** — `thread::sleep(300ms)` in transient-error test will break on slow CI; use a deterministic signal
- **Misleading test name** — `reload_fails_on_missing_config_file` doesn't test reload at all, just constructs an error and checks Display
- **Heavy snapshot boilerplate** — same 6-field struct literal repeated ~15 times; a helper like `make_snap(tick_id)` would cut noise significantly
- **Good coverage overall** — channel semantics, decode edge cases, and lifecycle error paths are well-targeted

src/frontend/src/exporter.rs
```diff
@@ -451,4 +451,290 @@ mod tests {
...
+        assert_eq!(exporter.call_count.load(Ordering::Relaxed), 2);
+    }
+
+    #[test]
+    fn exporter_handle_drop_joins_thread() {
+        let exporter = Arc::new(CountingExporter::new("drop-test"));
+        let (tx, rx) = snapshot_channel(2);
+        let handle = spawn_exporter_thread(exporter, rx, Duration::from_secs(5)).unwrap();
+        drop(tx);
+        // Dropping handle should join the thread without panic
```
banidoru, line 676: ⚠️ **Fragile assertion** — This asserts `call_count == 2`, assuming each snapshot is exported exactly once. If the transient-error retry logic re-attempts the *same* snapshot before moving to the next, `call_count` would be 3 (retry of snap 1 + snap 2). The test is tightly coupled to the current retry implementation. Consider asserting `>= 2` or checking that *both* tick_ids were eventually processed via a collected log.

src/frontend/src/poller.rs
```diff
@@ -420,4 +420,146 @@ mod tests {
...
+        assert_eq!(cfg.interval, Duration::from_millis(1000));
+    }
+
+    #[test]
+    fn poller_handle_drop_stops_thread() {
+        let config = PollerConfig {
+            stats_socket: PathBuf::from("/tmp/nonexistent-infmon-poller-drop-test.sock"),
+            interval: Duration::from_millis(100),
+        };
+        let (tx, _rx) = mpsc::sync_channel::<Arc<FlowStatsSnapshot>>(2);
```
banidoru, line 439: 🟡 **Unclear failure mode** — `spawn` is called with a nonexistent socket path. If `spawn` panics on connection failure (rather than retrying in the background), this test will panic instead of exercising the drop path. A brief comment explaining that `spawn` defers connection would help future readers. Same concern applies to `poller_with_no_senders` and `poller_with_multiple_senders`.

src/frontend/src/poller.rs
```diff
@@ -420,4 +420,146 @@ mod tests {
...
+            descriptors: vec![infmon_ipc::stats_client::RawDescriptor {
+                flow_rule_id: FlowRuleId { hi: 1, lo: 2 },
+                flow_rule_index: 0,
+                generation: 1,
+                epoch_ns: 0,
+                slots: vec![],
+                key_arena: vec![],
+                insert_failed: 5,
+                table_full: 0,
+            }],
```
banidoru, line 484: 🔵 **Minor: document the `decode_snapshot` signature** — The call `decode_snapshot(raw, 3, 1000, 3000, 2000)` uses positional args. The comment says `interval_ns: 1000 // 3000 - 2000` which clarifies the math, but a reader unfamiliar with the function signature has to look it up. Consider naming the args via a local variable or adding a one-line comment mapping the positions.

banidoru (commented):
- Good coverage of edge cases (zero-capacity channels, out-of-bounds keys, error variants)
- Transient-error retry test has a subtle count assumption that may break if retry logic changes
- Several poller tests connect to nonexistent sockets — behavior depends on internal error handling; consider documenting expected outcome
- Agree with prior reviewer on the unsafe pointer issue — that's the critical blocker

src/frontend/src/poller.rs
banidoru: Still unresolved — the unsafe pointer cast (`&stop as *const _ as usize` → deref) remains at line 551-554. Please switch to `Arc<AtomicBool>` as suggested.

src/frontend/src/exporter.rs
banidoru: Still unresolved — `thread::sleep(Duration::from_millis(300))` is still used at line 716. This remains a flakiness risk under CI load.

src/frontend/src/lifecycle.rs
banidoru: Still unresolved — `reload_fails_on_missing_config_file` still just constructs a `LifecycleError::ReloadFailed` and checks Display output, which is already covered by `lifecycle_error_display`. No actual reload is tested.

src/frontend/src/exporter.rs
banidoru: Still unresolved — the `FlowStatsSnapshot { tick_id, wall_clock_ns: 0, ... }` boilerplate still appears ~15 times. A `test_snap(tick_id)` helper would clean this up significantly.

src/frontend/src/exporter.rs
banidoru: Still unresolved — the assertion is still `== 2`, tightly coupled to retry implementation details.

src/frontend/src/poller.rs
banidoru: Still unresolved — no comment added explaining that `spawn` defers connection (or documenting the expected behavior with nonexistent socket paths).

src/frontend/src/poller.rs
banidoru: Still unresolved — positional args in `decode_snapshot(raw, 3, 1000, 3000, 2000)` remain undocumented. Minor but would help readability.

banidoru (changes requested):
**Iteration 2 re-review — none of the 7 prior concerns were addressed.**

The only new commit (`906aeb6`) is rustfmt formatting.

**Must fix:**
- 🔴 Unsafe pointer cast in `sleep_interruptible_stops_early` — use `Arc<AtomicBool>`
- ⚠️ `reload_fails_on_missing_config_file` is a duplicate of `lifecycle_error_display` — rename or replace with a real reload test

**Should fix:**
- ⚠️ Flaky `thread::sleep(300ms)` in transient-error test — use deterministic synchronization
- ⚠️ Fragile `== 2` assertion coupled to retry internals
- ⚠️ Add comments documenting `spawn` behavior with nonexistent sockets

**Nice to have:**
- 🔵 Extract `test_snap(tick_id)` helper (~15 duplicated struct literals)
- 🔵 Document `decode_snapshot` positional args

src/frontend/src/poller.rs
r12f: Fixed — replaced the unsafe `*const _ as usize` cast with `Arc<AtomicBool>`. The test now shares the flag safely between threads via `Arc::clone`. Commit 0c77f45.

src/frontend/src/exporter.rs
r12f: Good catch — bumped the sleep from 300ms to 2000ms to give loaded CI runners ample headroom for backoff processing. A `Condvar`-based approach would be more robust but is overkill for this test's scope; the generous timeout should prevent flakes. Commit 0c77f45.

src/frontend/src/lifecycle.rs
r12f: Agreed — renamed to `reload_failed_error_contains_reason` which accurately describes what it tests (the `Display` output of `ReloadFailed`). Commit 0c77f45.

src/frontend/src/exporter.rs
r12f: Done — extracted `fn test_snap(tick_id: u64) -> Arc<FlowStatsSnapshot>` and replaced all ~15 inline constructions. Much cleaner. Commit 0c77f45.

src/frontend/src/exporter.rs
r12f: Good point — relaxed the assertion from `== 2` to `>= 2` so a retry of the first snapshot doesn't cause a false failure. Added a comment explaining the rationale. Commit 0c77f45.

src/frontend/src/poller.rs
r12f: Added doc comments to all three tests (`poller_handle_drop_stops_thread`, `poller_with_no_senders`, `poller_with_multiple_senders`) explaining that `spawn()` defers the socket connection to the poller loop, so a nonexistent path retries internally without panicking. Commit 0c77f45.

src/frontend/src/poller.rs
r12f: Added a `// decode_snapshot(raw, tick_id, interval_ns, monotonic_ns, wall_clock_ns)` comment above the call to map the positional args. Commit 0c77f45.

banidoru (approved):
- All 7 prior review threads addressed and resolved
- Unsafe `*const _ as usize` correctly replaced with `Arc<AtomicBool>` — sound now
- `test_snap()` helper cleans up boilerplate nicely
- Flaky sleep bumped to 2s + assertion relaxed to `>= 2` — good enough for CI
- Minor doc/naming fixes all look correct

banidoru (approved):
- All 7 prior review concerns addressed in commit 0c77f45
- Unsafe `*const _ as usize` replaced with `Arc<AtomicBool>` — sound now
- `test_snap()` helper extracted, eliminating ~15x boilerplate
- Flaky sleep bumped to 2s + assertion relaxed to `>= 2` — reasonable tradeoff
- Test names and doc comments now accurately reflect intent
- LGTM

banidoru (approved):
All reviewers approved. LGTM.

banidoru (approved):
- All 7 prior review threads addressed and resolved
- `test_snap()` helper eliminates boilerplate — clean
- Unsafe pointer cast replaced with `Arc<AtomicBool>` — correct
- Flaky sleep bumped to 2s with `>= 2` assertion — adequate
- New tests cover error paths, channel edge cases, and `decode_snapshot` corner cases well
- Doc comments on deferred-connection behavior added as requested

LGTM.

PR #90, opened 2026-04-19
feat: add OTLP/gRPC metrics exporter

### LINKED ISSUE #56: [W18] OTLP exporter - gRPC metrics, flow/flow-rule mapping, cardinality cap
## Spec

## Scope
- OTLP/gRPC metrics exporter implementing Exporter trait
- 3 per-flow metrics: infmon.flow.packets (Sum), infmon.flow.bytes (Sum), infmon.flow.last_seen (Gauge)
- 5 per-flow-rule metrics: flows, evictions, drops, packets, bytes
- Resource attributes: service.name=infmon, host.name, host.arch, infmon.dpu.*
- Per-flow attributes from key fields (flow.src_ip, flow.mirror_src_ip, etc.)
- Per-tick point cap (default 2M), attribute length cap (256 bytes)

## Dependencies
Blocked by: #52

r12f:
## Summary

Implement the OTLP exporter ([spec 006](https://github.com/r12f/InFMon/blob/main/specs/006-exporter-otlp.md)) as a new `otlp` module in `infmon-frontend`, registered via the inventory-based plugin system.

### Per-flow metrics (3)
- `infmon.flow.packets` — Sum, cumulative, monotonic
- `infmon.flow.bytes` — Sum, cumulative, monotonic
- `infmon.flow.last_seen` — Gauge (ns)

### Per-flow-rule metrics (5)
- `infmon.flow-rule.flows` — Gauge
- `infmon.flow-rule.evictions` — Sum, cumulative
- `infmon.flow-rule.drops` — Sum, cumulative (with `reason` attribute)
- `infmon.flow-rule.packets` — Sum, cumulative
- `infmon.flow-rule.bytes` — Sum, cumulative

### Resource attributes
`service.name`, `service.namespace`, `service.version`, `service.instance.id`, `host.name`, `host.arch`, `host.id`, plus operator-supplied attrs via `resource.*` config keys (e.g. `infmon.dpu.id`, `infmon.dpu.platform`).

### Per-flow attribute mapping
Attributes derived from the flow-rule's declared field list only (spec §4). IPv4-mapped IPv6 addresses rendered as dotted-quad. Attributes sorted by key for stable wire bytes (spec §4.2).

### Cardinality safeguards (spec §8.2)
- Per-tick point cap (default 2M) with proportional per-flow-rule budget allocation
- 256-byte UTF-8-safe attribute value truncation with `…` suffix

### gRPC error classification
Status codes mapped to `Transient` (retryable) vs `Permanent` (disable exporter) errors per spec §7.3.

### Tests
12 unit tests covering: config validation, attribute mapping, IP rendering, cardinality cap, resource attributes, field-only attributes, and attribute sort order.

Closes #56

### DIFF
--- a/src/frontend/Cargo.toml
+++ b/src/frontend/Cargo.toml
@@ -17,6 +17,8 @@ infmon-ipc = { path = "../infmon-ipc", default-features = false }
 inventory = "0.3"
 libc = "0.2"
 log = "0.4"
+# NOTE: opentelemetry-proto 0.27 with gen-tonic-messages requires tonic 0.12.
+# Upgrading either crate independently may cause compile errors — always upgrade together.
 opentelemetry-proto = { version = "0.27", features = ["gen-tonic-messages", "metrics"] }
 serde_yaml = "0.9"
 signal-hook = "0.3"
--- a/src/frontend/Cargo.toml
+++ b/src/frontend/Cargo.toml
@@ -11,14 +11,18 @@ path = "src/main.rs"
 
 [dependencies]
 env_logger = "0.11"
+hostname = "0.4"
 infmon-config = { path = "../infmon-config" }
 infmon-ipc = { path = "../infmon-ipc", default-features = false }
 inventory = "0.3"
 libc = "0.2"
 log = "0.4"
+opentelemetry-proto = { version = "0.27", features = ["gen-tonic-messages", "metrics"] }
 serde_yaml = "0.9"
 signal-hook = "0.3"
 tokio = { version = "1", features = ["rt", "time"] }
+tonic = "0.12"
+uuid = { version = "1", features = ["v4"] }
 
 [dev-dependencies]
 tempfile = "3"
--- a/src/frontend/src/otlp.rs
+++ b/src/frontend/src/otlp.rs
@@ -419,10 +419,7 @@ impl Exporter for OtlpExporter {
         Box::pin(async move {
             let request = self.build_request(&snap);
 
-            let mut client = self
-                .get_client()
-                .await
-                .map_err(ExporterError::Transient)?;
+            let mut client = self.get_client().await.map_err(ExporterError::Transient)?;
 
             client.export(request).await.map_err(|e| {
                 // Classify gRPC status codes
--- a/src/frontend/src/otlp.rs
+++ b/src/frontend/src/otlp.rs
@@ -7,6 +7,7 @@
 
 use std::net::IpAddr;
 use std::sync::Arc;
+use std::time::Duration;
 
 use opentelemetry_proto::tonic::collector::metrics::v1::{
     metrics_service_client::MetricsServiceClient, ExportMetricsServiceRequest,
@@ -17,7 +18,7 @@ use opentelemetry_proto::tonic::metrics::v1::{
 };
 use opentelemetry_proto::tonic::resource::v1::Resource;
 use tokio::sync::Mutex;
-use tonic::transport::Channel;
+use tonic::transport::{Channel, Endpoint};
 
 use infmon_ipc::types::{FieldId, FieldValue, FlowRuleStats, FlowStatsSnapshot};
 
@@ -37,7 +38,6 @@ const ATTRIBUTE_LENGTH_CAP: usize = 256;
 const POINTS_PER_FLOW: u64 = 3;
 
 /// Number of per-flow-rule data points (flows, evictions, drops, packets, bytes).
-#[allow(dead_code)]
 const POINTS_PER_FLOW_RULE: u64 = 5;
 
 // ── OTLP Exporter ──────────────────────────────────────────────────
@@ -46,6 +46,8 @@ const POINTS_PER_FLOW_RULE: u64 = 5;
 pub struct OtlpExporter {
     name: String,
     endpoint: String,
+    #[allow(dead_code)] // stored for future reload support
+    instance_id_path: Option<String>,
     max_export_points_per_tick: u64,
     resource_attrs: Vec<KeyValue>,
     client: Mutex<Option<MetricsServiceClient<Channel>>>,
@@ -65,6 +67,9 @@ impl OtlpExporter {
             .get("max_export_points_per_tick")
             .and_then(|v| v.parse::<u64>().ok())
             .unwrap_or(DEFAULT_MAX_EXPORT_POINTS_PER_TICK);
+
+        // Optional configurable instance_id path (default: /var/lib/infmon/instance_id)
+        let instance_id_path = cfg.extra.get("instance_id_path").cloned();
 
         // Build resource attributes (spec §5).
         let mut resource_attrs = vec![
@@ -93,7 +98,10 @@ impl OtlpExporter {
         }
 
         // service.instance.id — read or create persistent instance ID
-        let instance_id = load_or_create_instance_id();
+        let id_path = instance_id_path
+            .as_deref()
+            .unwrap_or("/var/lib/infmon/instance_id");
+        let instance_id = load_or_create_instance_id(id_path);
         resource_attrs.push(kv_string("service.instance.id", &instance_id));
 
         // Operator-supplied resource attributes from extra config
@@ -109,6 +117,7 @@ impl OtlpExporter {
         Ok(Self {
             name: cfg.name.clone(),
             endpoint,
+            instance_id_path,
             max_export_points_per_tick,
             resource_attrs,
             client: Mutex::new(None),
@@ -125,17 +134,27 @@ impl OtlpExporter {
         }
 
         // Ensure endpoint has scheme
-        let endpoint =
+        let endpoint_uri =
             if self.endpoint.starts_with("http://") || self.endpoint.starts_with("https://") {
                 self.endpoint.clone()
             } else {
                 format!("http://{}", self.endpoint)
             };
 
-        let channel = Channel::from_shared(endpoint)?.connect().await?;
+        let channel = Endpoint::from_shared(endpoint_uri)?
+            .connect_timeout(Duration::from_secs(5))
+            .timeout(Duration::from_secs(10))
+            .connect()
+            .await?;
         let client = MetricsServiceClient::new(channel);
         *guard = Some(client.clone());
         Ok(client)
+    }
+
+    /// Clear the cached client so the next call reconnects.
+    async fn clear_client(&self) {
+        let mut guard = self.client.lock().await;
+        *guard = None;
     }
 
     /// Build the full OTLP request from a snapshot.
@@ -145,7 +164,12 @@ impl OtlpExporter {
         // Calculate total flows and apply cardinality cap (spec §8.2).
         let total_flows: u64 = snap.flow_rules.iter().map(|fr| fr.flows.len() as u64).sum();
 
-        let effective_cap = self.max_export_points_per_tick / POINTS_PER_FLOW.max(1);
+        // Reserve budget for per-flow-rule points before computing per-flow allocations.
+        let flow_rule_budget = snap.flow_rules.len() as u64 * POINTS_PER_FLOW_RULE;
+        let remaining_budget = self
+            .max_export_points_per_tick
+            .saturating_sub(flow_rule_budget);
+        let effective_cap = remaining_budget / POINTS_PER_FLOW.max(1);
 
         // Determine per-flow-rule budget
         let flow_budgets: Vec<u64> = if total_flows == 0 || total_flows <= effective_cap {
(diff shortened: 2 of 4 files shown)


src/frontend/src/otlp.rs
```diff
@@ -0,0 +1,880 @@
...
+            let allocated: u64 = budgets.iter().sum();
+            let mut remainder = effective_cap.saturating_sub(allocated);
+            for b in budgets.iter_mut() {
+                if remainder == 0 {
+                    break;
+                }
+                *b += 1;
+                remainder -= 1;
+            }
+            budgets
```
banidoru, line 199: **u64→i64 overflow risk**: `flow.counters.packets as i64` — counters are `u64` but cast with `as`, which silently wraps values ≥ 2⁶³. Same applies to `bytes` (line 196), `last_seen_ns` (line 214), and all the flow-rule counter casts below.

Consider `i64::try_from(val).unwrap_or(i64::MAX)` to clamp, or switch to `AsDouble` for counters that may exceed 2⁶³.

src/frontend/src/otlp.rs
```diff
@@ -0,0 +1,880 @@
...
+            endpoint,
+            max_export_points_per_tick,
+            resource_attrs,
+            client: Mutex::new(None),
+        })
+    }
+
+    /// Lazily connect to the gRPC endpoint.
+    async fn get_client(
+        &self,
```
banidoru, line 129: **No connect timeout**: `Channel::from_shared(endpoint)?.connect().await?` has no timeout. If the collector is unreachable, this blocks the exporter indefinitely.

Consider `.connect_timeout(Duration::from_secs(5))` and `.timeout(Duration::from_secs(10))` on the `Endpoint` builder.

src/frontend/src/otlp.rs
```diff
@@ -0,0 +1,880 @@
...
+        if let Some(ref client) = *guard {
+            return Ok(client.clone());
+        }
+
+        // Ensure endpoint has scheme
+        let endpoint =
+            if self.endpoint.starts_with("http://") || self.endpoint.starts_with("https://") {
+                self.endpoint.clone()
+            } else {
+                format!("http://{}", self.endpoint)
```
banidoru, line 141: **Cardinality cap doesn't account for per-flow-rule points**: `effective_cap` is computed as `max_export_points_per_tick / POINTS_PER_FLOW`, but each flow-rule unconditionally emits 5 additional metrics (lines 222-310). With many flow-rules and a tight cap, actual emitted points can exceed the configured limit.

Subtract `snap.flow_rules.len() * POINTS_PER_FLOW_RULE` from the budget before computing per-flow allocations.

src/frontend/src/otlp.rs
```diff
@@ -0,0 +1,880 @@
...
+                        data_points: vec![NumberDataPoint {
+                            attributes: drops_attrs,
+                            time_unix_nano: wall_ns,
+                            value: Some(
+                                opentelemetry_proto::tonic::metrics::v1::number_data_point::Value::AsInt(
+                                    fr.counters.drops as i64,
+                                ),
+                            ),
+                            ..Default::default()
+                        }],
```
banidoru: `reload()` returns `Ok(())` but applies zero changes — endpoint, caps, and resource attrs are all immutable after construction. This is misleading; callers assume reload succeeded.

Either return `Err(ConfigError::Unsupported("OTLP exporter requires restart"))` or actually apply the new config (at minimum `max_export_points_per_tick`).

src/frontend/src/otlp.rs
```diff
@@ -0,0 +1,880 @@
...
+                if !attr_key.is_empty() {
+                    resource_attrs.push(kv_string(attr_key, v));
+                }
+            }
+        }
+
+        Ok(Self {
+            name: cfg.name.clone(),
+            endpoint,
+            max_export_points_per_tick,
```
banidoru, line 121: Minor: the cached client is never invalidated on transient errors. After a network blip the stale `Channel` may keep failing. Consider clearing `self.client` on transient gRPC errors so the next tick reconnects.

banidoru (commented):
- **u64→i64 casts can silently overflow** — `packets`, `bytes`, and nanosecond timestamps are `u64` but cast to `i64` via `as i64`. Counters above `i64::MAX` wrap negative. Use `try_into` or document the accepted range.
- **No gRPC connect/request timeout** — `Channel::connect()` and `export()` have no timeout configured. A hung collector will block the export tick indefinitely.
- **Cardinality cap ignores per-flow-rule points** — The budget calculation divides by `POINTS_PER_FLOW` only, but each flow-rule also emits 5 fixed metrics. Under extreme cap pressure these uncapped points could exceed the intended limit.
- **`reload()` silently no-ops** — Returns `Ok(())` without applying new config (endpoint, caps, resource attrs). Callers have no way to know the reload was ignored. Should return an error or log a warning.

banidoru (changes requested):
- Addresses existing review comments.

src/frontend/src/otlp.rs
```diff
@@ -0,0 +1,880 @@
...
+            if !id.is_empty() {
+                resource_attrs.push(kv_string("host.id", id));
+            }
+        }
+
+        // service.instance.id — read or create persistent instance ID
+        let instance_id = load_or_create_instance_id();
+        resource_attrs.push(kv_string("service.instance.id", &instance_id));
+
+        // Operator-supplied resource attributes from extra config
```
banidoru, line 107: **`load_or_create_instance_id()` called in constructor — blocks on filesystem I/O**: This reads/writes `/var/lib/infmon/instance_id` synchronously. In an async context (or if `/var/lib/infmon` is on a slow/network FS), this could stall. Consider making this `async` or caching the result at startup before entering the async runtime.

src/frontend/src/otlp.rs
```diff
@@ -0,0 +1,880 @@
...
+
+            // Per-flow-rule observability points (spec §3.4)
+            let fr_attrs = vec![kv_string("flow-rule", &fr.name)];
+
+            // infmon.flow-rule.flows (Gauge)
+            metrics.push(Metric {
+                name: "infmon.flow-rule.flows".into(),
+                description: "Number of live flows in this flow-rule".into(),
+                unit: "{flows}".into(),
+                data: Some(
```
banidoru, line 382: **Hard-coded drop reason**: `reason` is always `"eviction_failed"`. If additional drop reasons are added later, this won't distinguish them. Consider deriving the reason from the data (or at minimum adding a TODO/spec reference for future extensibility).

src/frontend/src/otlp.rs
```diff
@@ -0,0 +1,880 @@
...
+                .iter()
+                .map(|fr| fr.flows.len() as u64)
+                .collect()
+        } else {
+            // Proportional allocation
+            let mut budgets: Vec<u64> = snap
+                .flow_rules
+                .iter()
+                .map(|fr| {
+                    (effective_cap as u128 * fr.flows.len() as u128 / total_flows as u128) as u64
```
banidoru, line 186: **One `Metric` struct per flow per metric name**: This creates N separate `Metric` objects each with a single data point, rather than grouping all data points for the same metric name into one `Metric` with N data points. OTLP consumers typically expect the latter — batching data points under one `Metric` reduces serialization overhead and is more idiomatic. Consider collecting all `infmon.flow.packets` data points into a single `Sum`, etc.

src/frontend/src/otlp.rs
```diff
@@ -0,0 +1,880 @@
...
+                        dropped_attributes_count: 0,
+                    }),
+                    metrics,
+                    schema_url: String::new(),
+                }],
+                schema_url: String::new(),
+            }],
+        }
+    }
+}
```
banidoru, line 475: **`load_or_create_instance_id` writes to a hard-coded path with no config override**: `/var/lib/infmon/instance_id` may not be writable in containerized or read-only-root deployments. The silent fallback means the instance ID changes on every restart in those environments, breaking `service.instance.id` stability. Consider allowing the path to be configured or falling back to a hash of `machine-id + binary path`.

src/frontend/Cargo.toml
```diff
@@ -11,14 +11,18 @@ path = "src/main.rs"
 
 [dependencies]
 env_logger = "0.11"
+hostname = "0.4"
 infmon-config = { path = "../infmon-config" }
 infmon-ipc = { path = "../infmon-ipc", default-features = false }
 inventory = "0.3"
 libc = "0.2"
 log = "0.4"
+opentelemetry-proto = { version = "0.27", features = ["gen-tonic-messages", "metrics"] }
```
banidoru, line 22: **tonic 0.12 vs opentelemetry-proto 0.27 compatibility**: Verify these crate versions are tested together. `opentelemetry-proto` with `gen-tonic-messages` generates code against a specific tonic version — a mismatch can cause confusing compile errors or silent ABI issues on upgrades.

banidoru (commented):
- **Metrics structure**: One `Metric` per flow per metric name instead of batching data points — increases wire size and deviates from OTLP conventions
- **Instance ID persistence**: Hard-coded `/var/lib/infmon/instance_id` breaks in containers/read-only roots — ID changes every restart
- **Sync I/O in constructor**: `load_or_create_instance_id()` does blocking FS ops
- **Hard-coded drop reason**: Only `eviction_failed` — no extensibility path
- **Dependency version pairing**: tonic 0.12 + opentelemetry-proto 0.27 compatibility should be verified

src/frontend/src/otlp.rs
r12f: Fixed: replaced all `as i64` casts with `saturating_i64()` which clamps at `i64::MAX` instead of silently wrapping. Applied to packets, bytes, last_seen_ns, and all flow-rule counter casts.

src/frontend/src/otlp.rs
r12f: Fixed: added `.connect_timeout(Duration::from_secs(5))` and `.timeout(Duration::from_secs(10))` on the `Endpoint` builder before `.connect()`.

src/frontend/src/otlp.rs
r12f: Fixed: the cardinality cap now subtracts `flow_rules.len() * POINTS_PER_FLOW_RULE` from the budget before computing per-flow allocations. Also removed `#[allow(dead_code)]` from `POINTS_PER_FLOW_RULE` since it's now used. Updated the test accordingly.

src/frontend/src/otlp.rs
r12f: Fixed: `reload()` now returns `Err(ConfigError(...))` with a clear message that the OTLP exporter requires a restart to apply config changes.

src/frontend/src/otlp.rs
r12f: Fixed: added `clear_client()` helper. On transient gRPC errors (Unavailable, DeadlineExceeded, ResourceExhausted, Aborted, and the catch-all), the cached client is now cleared so the next tick reconnects with a fresh channel.

src/frontend/src/otlp.rs
r12f: Acknowledged — `load_or_create_instance_id()` is called in the synchronous `new()` constructor, not in an async context, so it won't block the tokio runtime. The constructor runs once at startup before entering the async event loop. The file I/O is bounded (single small file read/write) and only happens once. No change needed here.

src/frontend/src/otlp.rs
r12f: Added a TODO comment noting this should be derived from data when additional drop reasons are added. Currently `eviction_failed` is the only drop reason in the data model, so there's nothing to derive from yet.

src/frontend/src/otlp.rs
r12f: Fixed: refactored `build_request()` to collect all data points into per-metric-name vectors, then emit one `Metric` object per metric name with all data points grouped together. This is more idiomatic OTLP and reduces serialization overhead.

src/frontend/src/otlp.rs
r12f: Fixed: the instance ID path is now configurable via the `instance_id_path` extra config key (defaults to `/var/lib/infmon/instance_id`). The `load_or_create_instance_id()` function now takes a path parameter and derives the parent directory from it.

src/frontend/Cargo.toml
r12f: Good call. Added a comment to Cargo.toml noting that opentelemetry-proto 0.27 with gen-tonic-messages requires tonic 0.12, and they must be upgraded together. The current versions are compatible and tested (CI passes).

banidoru (approved):
- All 10 prior review threads addressed and resolved — overflow clamping, connect timeouts, cardinality budget, reload error, client invalidation, metric grouping, configurable instance ID path, and crate version note all look good.
- Metric grouping refactor (one `Metric` per name with N data points) is clean and idiomatic OTLP.
- Cardinality cap correctly reserves flow-rule budget before flow allocation; test updated accordingly.
- Error classification with client invalidation on transient gRPC failures is well-structured.
- No new issues found in the updated diff. LGTM.

banidoru (approved):
- All previous comments have been addressed. - The use of 'saturating_i64()' for counter conversions is correct. - Connection timeouts and client caching logic have been improved. - The data point batching is now more efficient. LGTM.

src/frontend/src/otlp.rs
```diff
@@ -0,0 +1,985 @@
...
+        }
+
+        // Ensure endpoint has scheme
+        let endpoint_uri =
+            if self.endpoint.starts_with("http://") || self.endpoint.starts_with("https://") {
+                self.endpoint.clone()
+            } else {
+                format!("http://{}", self.endpoint)
+            };
+
```
banidoru, line 143: **Eager `connect()` blocks the export tick**: `Endpoint::connect().await` performs a full TCP handshake (up to the 5s connect timeout) before returning. If the collector is temporarily unreachable on the first export tick, this stalls the entire tick.

Consider `connect_lazy()` instead — it returns immediately and establishes the connection on first RPC. The transient-error client-clearing logic you added already handles reconnection on failure, so lazy connect fits naturally here.

```rust
let channel = Endpoint::from_shared(endpoint_uri)?
    .connect_timeout(Duration::from_secs(5))
    .timeout(Duration::from_secs(10))
    .connect_lazy();
```

banidoru (commented):
Re-review (iteration 2) — all prior concerns addressed and threads resolved. New observations:

- **`connect()` vs `connect_lazy()`**: eager connect in `get_client` blocks the export tick until TCP handshake completes (up to 5s with the new timeout). Consider `connect_lazy()` to avoid stalling.
- Cardinality cap logic and test assertions look correct after the fix.
- `saturating_i64`, configurable instance ID path, error classification with client invalidation, and batched Metric objects all look solid.
- No new correctness issues found in this iteration.

banidoru (approved):
All reviewers approved. LGTM.

Issue #60, opened 2026-04-19
[W22] CLI output contract - table/json/csv + exit codes + SIGPIPE


r12f:
## Spec
[007-cli](https://github.com/r12f/InFMon/blob/main/specs/007-cli.md)

## Scope
- --output table|json on every read command; stable JSON schema
- 11 exit codes (0-143)
- SIGPIPE -> exit 0; SIGINT -> 130; SIGTERM -> 143
- Global flags: --json, --compact, --raw-bytes, --config, --socket, --timeout, --no-color, -q, -v

## Dependencies
Blocked by: #59

PR #93, opened 2026-04-20
feat(cli): scaffold infmonctl binary with full verb tree

### LINKED ISSUE #59: [W21] infmonctl CLI - full verb tree
## Spec

## Scope
- Single Rust binary /usr/bin/infmonctl using clap
- Verb tree:
  - install/uninstall, start/stop/restart/status
  - config get/set/reload/show
  - flow-rule add/rm/list/show
  - flow list/show
  - stats show/export
  - log tail, health
- Connects to frontend via InFMonControlClient (shared ipc crate)
- Privilege model: root for writes, infmon group for reads via Unix socket

## Dependencies
Blocked by: #50

r12f:
## Summary

Implement the `infmonctl` CLI crate (`infmon-cli`) with clap-derived argument parsing for the complete v1 verb tree from [spec 007](https://github.com/r12f/InFMon/blob/main/specs/007-cli.md).

### What's included

- **Full verb tree**: install/uninstall, start/stop/restart/status, config get/set/reload/show, flow-rule add/rm/list/show, flow list/show, stats show/export, log tail, health
- **Global flags**: `--output table|json`, `--json`, `--compact`, `--raw-bytes`, `--config`, `--socket`, `--timeout`, `--no-color`, `-q`, `-v`
- **Exit codes module**: frozen v1 table (0, 1–7, 13, 69, 130, 143)
- **Privilege gates**: root-only verbs exit 13 for non-root callers
- **SIGPIPE handling**: clean exit 0 on broken pipe
- **Output helpers**: `TableDisplay` trait + JSON formatting utilities
- **38 integration tests**: argument parsing for every verb, flag combination, and rejection case

### What's NOT included (stubs)

All subcommand handlers are stubs returning exit 1. Actual frontend IPC integration (`InFMonControlClient`) will follow in subsequent issues (#60 builds on this).

Closes #59

### DIFF
--- a/src/cli/src/lib.rs
+++ b/src/cli/src/lib.rs
@@ -1,0 +1,248 @@
+use clap::{Parser, Subcommand, ValueEnum};
+
+pub mod exit_codes;
+pub mod output;
+
+/// infmonctl — operator CLI for InFMon
+///
+/// Single entry point for installing, running, configuring, observing,
+/// and debugging InFMon on a BlueField-3 DPU.
+#[derive(Parser, Debug)]
+#[command(name = "infmonctl", version, about, long_about = None)]
+pub struct Cli {
+    #[command(subcommand)]
+    pub command: Commands,
+
+    /// Output format
+    #[arg(long, value_enum, global = true, default_value_t = OutputFormat::Table)]
+    pub output: OutputFormat,
+
+    /// Shorthand for --output json
+    #[arg(long, global = true, default_value_t = false)]
+    pub json: bool,
+
+    /// JSON only: collapse pretty-printed output to a single line per record
+    #[arg(long, global = true, default_value_t = false)]
+    pub compact: bool,
+
+    /// Table mode only: print byte counters as raw integers
+    #[arg(long, global = true, default_value_t = false)]
+    pub raw_bytes: bool,
+
+    /// Override config file location
+    #[arg(long, global = true, default_value = "/etc/infmon/config.yaml")]
+    pub config: String,
+
+    /// Override frontend socket path
+    #[arg(long, global = true, default_value = "/run/infmon/frontend.sock")]
+    pub socket: String,
+
+    /// Per-call timeout when contacting the frontend (seconds)
+    #[arg(long, global = true, default_value_t = 5)]
+    pub timeout: u64,
+
+    /// Disable ANSI colors
+    #[arg(long, global = true, default_value_t = false)]
+    pub no_color: bool,
+
+    /// Suppress informational stderr
+    #[arg(short, long, global = true, default_value_t = false)]
+    pub quiet: bool,
+
+    /// Verbosity level (-v info, -vv debug, -vvv trace)
+    #[arg(short, long, global = true, action = clap::ArgAction::Count)]
+    pub verbose: u8,
+}
+
+#[derive(Debug, Clone, ValueEnum)]
+pub enum OutputFormat {
+    Table,
+    Json,
+}
+
+impl Cli {
+    /// Resolve the effective output format (--json overrides --output)
+    pub fn effective_output(&self) -> OutputFormat {
+        if self.json {
+            OutputFormat::Json
+        } else {
+            self.output.clone()
+        }
+    }
+}
+
+#[derive(Subcommand, Debug)]
+pub enum Commands {
+    /// Bootstrap InFMon installation
+    Install {
+        /// Proceed even when prior state looks dirty
+        #[arg(long)]
+        force: bool,
+    },
+    /// Remove InFMon installation
+    Uninstall {
+        /// Also remove /etc/infmon/ and /var/lib/infmon/
+        #[arg(long)]
+        purge: bool,
+    },
+    /// Start infmon-frontend service
+    Start,
+    /// Stop infmon-frontend service
+    Stop,
+    /// Restart infmon-frontend service
+    Restart,
+    /// Show composite service status
+    Status,
+    /// Configuration management
+    Config {
+        #[command(subcommand)]
+        command: ConfigCommands,
+    },
+    /// Flow-rule management (configured matchers)
+    FlowRule {
+        #[command(subcommand)]
+        command: FlowRuleCommands,
+    },
+    /// Live flow inspection (read-only)
+    Flow {
+        #[command(subcommand)]
+        command: FlowCommands,
+    },
+    /// Aggregate statistics
+    Stats {
+        #[command(subcommand)]
+        command: StatsCommands,
+    },
+    /// Log access
+    Log {
+        #[command(subcommand)]
+        command: LogCommands,
+    },
+    /// Composite health probe
+    Health,
+}
+
+#[derive(Subcommand, Debug)]
+pub enum ConfigCommands {
+    /// Get a single config key value
+    Get {
+        /// Dotted key path (e.g. exporter.otlp.endpoint)
+        key: String,
+    },
+    /// Set a config key value (requires root)
+    Set {
+        /// Dotted key path
+        key: String,
+        /// New value
+        value: String,
+        /// Force value type when ambiguous
+        #[arg(long, value_name = "TYPE")]
+        r#type: Option<String>,
+    },
+    /// Reload configuration (requires root)
+    Reload,
+    /// Show effective merged configuration
+    Show {
+        /// Add restart-required annotations
+        #[arg(long)]
+        annotate: bool,
+    },
+}
+
+#[derive(Subcommand, Debug)]
+pub enum FlowRuleCommands {
+    /// Add a flow rule (requires root)
+    Add {
+        /// Flow-rule spec in key=value form
+        spec: Vec<String>,
+    },
+    /// Remove a flow rule (requires root)
+    Rm {
+        /// Rule ID, name, or spec to remove
+        target: String,
+        /// Delete all matching rules when spec matches multiple
+        #[arg(long)]
+        all: bool,
+    },
+    /// List all flow rules
+    List,
+    /// Show detailed info for a flow rule
+    Show {
+        /// Rule ID or name
+        target: String,
+    },
+}
+
+#[derive(Subcommand, Debug)]
+pub enum FlowCommands {
+    /// List active flows under a flow-rule
+    List {
+        /// Rule ID or name
+        rule: String,
+        /// Show top N flows
+        #[arg(long, default_value_t = 50)]
+        top: usize,
+        /// Sort by field
+        #[arg(long, value_enum, default_value_t = FlowSortField::Bytes)]
+        sort: FlowSortField,
+    },
+    /// Show a single flow's full counters
+    Show {
+        /// Rule ID or name
+        rule: String,
+        /// Canonical key tuple (single quoted argument)
+        key: String,
+    },
+}
+
+#[derive(Debug, Clone, ValueEnum)]
+pub enum FlowSortField {
+    Bytes,
+    Packets,
+    LastSeen,
+}
+
+#[derive(Subcommand, Debug)]
+pub enum StatsCommands {
+    /// Show aggregate counters
+    Show {
+        /// Restrict to a specific metric (supports globs)
+        #[arg(long)]
+        name: Option<String>,
+        /// Top-N entries
+        #[arg(long, default_value_t = 20)]
+        top: usize,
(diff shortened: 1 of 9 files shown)


src/cli/src/output.rs
```diff
@@ -0,0 +1,26 @@
...
+    value: &T,
+    format: &OutputFormat,
+    compact: bool,
+) {
+    match format {
+        OutputFormat::Table => value.print_table(),
+        OutputFormat::Json => {
+            if compact {
+                println!("{}", serde_json::to_string(value).unwrap());
+            } else {
```
banidoru, line 17: `unwrap()` here will panic if serialization fails (e.g. map with non-string keys, or a `Serialize` impl that errors). Prefer `unwrap_or_else` with an error message to stderr + return, or at least document why it's safe for all expected types.

src/cli/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+    },
+}
+
+#[derive(Subcommand, Debug)]
+pub enum FlowRuleCommands {
+    /// Add a flow rule (requires root)
+    Add {
+        /// Flow-rule spec in key=value form
+        spec: Vec<String>,
+    },
```
banidoru, line 159: `spec: Vec<String>` allows `infmonctl flow-rule add` with zero specs, which would be a no-op. Consider `#[arg(required = true, num_args = 1..)]` to enforce at least one spec at the parse layer.

src/cli/src/main.rs
```diff
@@ -0,0 +1,212 @@
...
+// ---------------------------------------------------------------------------
+// Subcommand implementations (stubs that connect to the frontend via IPC)
+// ---------------------------------------------------------------------------
+
+async fn run_install(force: bool) -> i32 {
+    let _ = force;
+    // Check privilege
+    if !is_root() {
+        eprintln!("infmonctl: install requires root privileges");
+        return EXIT_PERMISSION_DENIED;
```
banidoru, line 59: All the `run_*` stub functions are `async fn` but perform no async work (no `.await`). This triggers `#[warn(unused)]`-family lints in some configurations. Consider making them plain `fn` returning `i32` until actual IPC is wired, or add a brief `// TODO: will await IPC client` comment to signal intent.

src/cli/src/output.rs
```diff
@@ -0,0 +1,26 @@
+use serde::Serialize;
+
+use crate::OutputFormat;
+
+/// Print a value as table or JSON based on the output format.
+pub fn print_output<T: Serialize + TableDisplay>(
```
banidoru: This function takes `compact` as a standalone param but doesn't accept `raw_bytes` or `no_color` — the caller will need to thread those through separately. Consider taking `&Cli` (or a small `OutputOptions` struct) so all formatting knobs are available in one place when the real rendering lands.

src/cli/tests/cli_args.rs
```diff
@@ -0,0 +1,362 @@
...
+            assert_eq!(spec, vec!["src_ip=10.0.0.0/8", "name=test"]);
+        }
+        _ => panic!("expected flow-rule add"),
+    }
+}
+
+#[test]
+fn parse_flow_rule_rm() {
+    let cli = parse(&["flow-rule", "rm", "test-rule"]);
+    match cli.command {
```
banidoru, line 174: Missing a test for `flow-rule add` with **zero** specs — this should either be rejected by clap or handled. Currently it silently parses as an empty vec.

banidoru (commented):
- Overall solid scaffold — verb tree, exit codes, privilege gates, SIGPIPE handling all look correct
- `unwrap()` in `print_output` can panic on serialization failure; should handle gracefully
- `flow-rule add` accepts an empty spec vec — needs `#[arg(required = true)]` or clap `num_args(1..)`
- Stub functions are all `async` with no `.await` — consider making them sync until IPC lands to avoid dead-code warnings
- `output::print_output` signature takes `compact` but ignores `raw_bytes`/`no_color` from `Cli` — will be easy to forget wiring these later

src/cli/src/main.rs
```diff
@@ -0,0 +1,212 @@
...
+
+async fn run(cli: Cli) -> i32 {
+    let _output_format = cli.effective_output();
+
+    match cli.command {
+        Commands::Install { force } => run_install(force).await,
+        Commands::Uninstall { purge } => run_uninstall(purge).await,
+        Commands::Start => run_lifecycle("start").await,
+        Commands::Stop => run_lifecycle("stop").await,
+        Commands::Restart => run_lifecycle("restart").await,
```
banidoru, line 39: `_output_format` is computed here but never passed to any subcommand handler. Every `run_*` function independently ignores the format. This will become dead code unless the format is threaded through — consider passing it as a parameter now (or at least a `TODO` comment) so the plumbing is obvious when stubs get filled in.

src/cli/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+    /// Per-call timeout when contacting the frontend (seconds)
+    #[arg(long, global = true, default_value_t = 5)]
+    pub timeout: u64,
+
+    /// Disable ANSI colors
+    #[arg(long, global = true, default_value_t = false)]
+    pub no_color: bool,
+
+    /// Suppress informational stderr
+    #[arg(short, long, global = true, default_value_t = false)]
```
banidoru, line 49: `--timeout 0` is silently accepted but will almost certainly cause instant timeouts when IPC is wired up. Consider adding `#[arg(value_parser = clap::value_parser!(u64).range(1..))]` or documenting that 0 means "no timeout".

src/cli/src/main.rs
```diff
@@ -0,0 +1,212 @@
...
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+fn is_root() -> bool {
+    #[cfg(unix)]
+    {
+        unsafe { libc::geteuid() == 0 }
+    }
```
banidoru, line 225: On non-unix, `is_root()` unconditionally returns `false`, meaning all privilege-gated commands will always fail with exit 13. Worth a brief doc comment clarifying this is intentional (DPU-only target) so future contributors don't wonder if it's a bug.

banidoru (commented):
- Scaffold looks solid — full verb tree, exit codes, privilege gates, SIGPIPE, 38 tests. Good foundation.
- `_output_format` computed but never threaded to handlers — dead code today, will need plumbing.
- `--timeout 0` is accepted but will likely cause instant timeouts; consider a minimum or validation.
- Minor: `is_root()` always returns `false` on non-unix — fine for DPU-only, but a doc comment would clarify intent.

src/cli/src/output.rs
r12f: Good catch — replaced `unwrap()` with `unwrap_or_else` that prints a JSON error object to stdout instead of panicking. This keeps the output machine-parseable even on serialization failure. Fixed in 83233f0.

src/cli/src/lib.rs
r12f: Agreed — added `#[arg(required = true, num_args = 1..)]` to enforce at least one spec at the parse layer. Also added a `reject_flow_rule_add_without_spec` test to cover this. Fixed in 83233f0.

src/cli/src/main.rs
r12f: Valid point. These are intentionally `async fn` because they will `.await` IPC client calls once wired — updated the section comment to clarify this intent: "stubs — will await IPC client when wired". Keeping them async avoids a churn commit when IPC lands.

src/cli/src/output.rs
r12f: Reasonable suggestion for when real rendering lands. For now the only consumer is `print_output` which threads the knobs explicitly. I'll defer the `OutputOptions` struct to the IPC wiring PR where the real rendering logic will need all the flags in one place — avoids premature abstraction at the stub stage.

src/cli/tests/cli_args.rs
r12f: Added `reject_flow_rule_add_without_spec` test and enforced `required = true` on the arg so clap rejects zero specs at parse time. Fixed in 83233f0.

src/cli/src/main.rs
r12f: Acknowledged — `_output_format` is intentionally computed-but-unused at this stub stage to verify the flag-resolution logic works (tested in `parse_json_flag` and `parse_output_json`). Each `run_*` handler will receive it (or access it via `&Cli`) when IPC rendering is wired. The leading underscore signals this. No code change needed now.

src/cli/src/lib.rs
r12f: Good edge case. For this scaffold PR, `--timeout 0` is harmless since nothing does IPC yet. I'll add a `value_parser = clap::value_parser!(u64).range(1..)` constraint in the IPC wiring PR where it actually matters — adding it now would be untestable dead validation. Noting for follow-up.

src/cli/src/main.rs
r12f: Correct — InFMon targets Linux/BlueField-3 exclusively (the spec doesn't cover Windows). The `#[cfg(not(unix))]` branch is a compile-time fallback to avoid build errors if someone opens the workspace on a non-unix host; it's not a supported runtime path. Adding a doc comment to clarify.

src/cli/src/output.rs
```diff
@@ -0,0 +1,30 @@
...
+/// Print a value as table or JSON based on the output format.
+pub fn print_output<T: Serialize + TableDisplay>(value: &T, format: &OutputFormat, compact: bool) {
+    match format {
+        OutputFormat::Table => value.print_table(),
+        OutputFormat::Json => {
+            if compact {
+                println!(
+                    "{}",
+                    serde_json::to_string(value)
+                        .unwrap_or_else(|e| format!("{{\"error\":\"serialize failed: {e}\"}}")),
```
banidoru: Nit: `format!("{{\"error\":\"serialize failed: {e}\"}}")` will produce invalid JSON if the serde error message contains `"` or `\`. Safer alternative:
```rust
.unwrap_or_else(|e| serde_json::json!({"error": format!("serialize failed: {e}")}).to_string())
```
Non-blocking — unlikely to hit in practice with current types.

banidoru (approved):
Iteration 2 re-review — all prior threads resolved. New commit cleanly addresses feedback:

- `unwrap()` → `unwrap_or_else` with error JSON fallback ✅
- `flow-rule add` now enforces `required = true, num_args = 1..` ✅
- New `reject_flow_rule_add_without_spec` test ✅
- rustfmt applied throughout ✅
- Async stub comment clarified ✅

One minor nit: the `unwrap_or_else` error fallback in `output.rs` uses `format!` to build JSON — if the serde error message ever contains quotes or backslashes, the output won't be valid JSON. Consider `serde_json::json!({"error": e.to_string()}).to_string()` for bulletproof output. Not blocking.

src/cli/src/output.rs
banidoru: Still using `format!()` with manual brace escaping — the concern about invalid JSON when the serde error contains `"` or `\\` still applies. Non-blocking nit, but the `serde_json::json!({...}).to_string()` approach would be safer. Leaving open.

banidoru (approved):
Re-review (iteration 2):

- Previous concerns addressed: `unwrap()` → `unwrap_or_else`, `spec` enforced `required = true`, intent comments added for async stubs and `_output_format`
- One prior nit still open: `format!()`-based error JSON in output.rs can produce invalid JSON on special chars — non-blocking
- New code is clean: CLI scaffold, exit codes, test coverage all solid

LGTM.

banidoru (approved):
All reviewers approved. LGTM.

src/cli/src/output.rs
r12f: Fixed — switched both compact and pretty paths to use `serde_json::json!({"error": ...}).to_string()` so the fallback JSON is always valid regardless of what characters the serde error contains. Fixed in e22bd35.

banidoru (approved):
All prior review concerns addressed. LGTM.

- `unwrap()` → `unwrap_or_else` with `serde_json::json!` fallback — clean fix
- `flow-rule add` now enforces `required = true, num_args = 1..` with test coverage
- Async stub intent clarified in comments
- No new issues found in iteration 3 diff

PR #96, opened 2026-04-20
feat(packaging): align Debian packaging with spec 008 v0.2

### LINKED ISSUE #62: [W24] Debian packaging - 3 binary .debs + meta-package + maintainer scripts
## Spec
[008-packaging-install](https://github.com/r12f/InFMon/blob/main/specs/008-packaging-install.md)

## Scope
- 3 binary packages + 1 meta-package: infmon-backend, infmon-frontend, infmon-cli, infmon
- Version-locked dependencies between packages
- VPP dependency: vpp (>= 24.10)
- Filesystem layout per spec 008
- infmon system user/group created by frontend postinst
- Idempotent install/upgrade/uninstall maintainer scripts
- packaging/debian/ source tree with override_dh_auto_* targets for cmake+cargo

## Dependencies
Blocked by: #41, #53, #59

r12f:
## Summary

Bring `packaging/debian/` into full compliance with spec 008 v0.2. This addresses all scope items from issue #62.

### Changes

**Config unification (spec §5.1, v0.2 update):**
- Replace separate `backend.conf` and `frontend.conf` (INI format) with a single `config.yaml` (YAML) at `/etc/infmon/config.yaml`, owned by `infmon-backend`
- Frontend no longer ships its own config file; it reads the shared config via its dependency on `infmon-backend`

**Version-locked dependencies (spec §3):**
- `infmon-frontend` now hard-depends on `infmon-backend (= ${binary:Version})` (was `Recommends` with `${source:Version}`)
- Meta-package and all inter-package deps use `${binary:Version}`
- Meta-package set to `Architecture: all`

**systemd unit alignment (spec §6):**
- Remove `Requires=vpp.service` and `PartOf=vpp.service` — frontend uses its own retry loop
- `CapabilityBoundingSet=` (empty, drop all capabilities)
- `ReadOnlyPaths=/dev/hugepages` (spec default; fallback documented)
- `ExecStart` references `config.yaml` instead of `frontend.conf`
- `--no-enable --no-start` in `dh_installsystemd` (disabled by default)

**Naming (spec v0.2):**
- `flow_defs.d` → `flow_rules.d` everywhere

**Maintainer scripts (spec §7.3/§7.4):**
- `postrm` scripts use `rmdir --ignore-fail-on-non-empty` for safe cross-package cleanup of `/etc/infmon/`
- Frontend postrm no longer removes backend-owned files

**README.Debian (spec §4.2/§4.3):**
- Replace `dpkg --force-depends` recommendation with `equivs` shim approach
- Add GPG key fingerprint verification instructions
- Document optional `PartOf=vpp.service` drop-in for operators
- Document `systemctl enable --now` post-install workflow

Closes #62


packaging/debian/control
```diff
@@ -63,16 +62,16 @@ Description: InFMon flow telemetry — command-line client
...
+Architecture: all
 Depends:
  ${misc:Depends},
- infmon-backend (= ${source:Version}),
- infmon-frontend (= ${source:Version}),
- infmon-cli (= ${source:Version}),
+ infmon-backend (= ${binary:Version}),
+ infmon-frontend (= ${binary:Version}),
+ infmon-cli (= ${binary:Version}),
 Description: InFMon flow telemetry — meta package
```
banidoru, line 76: `Architecture: all` with `${binary:Version}` deps on `arch:any` packages will break on binNMUs. During a binNMU, the `arch:any` packages get a `+bX` suffix in their version but the `arch:all` meta-package does not, so apt cannot satisfy the exact-version dependency.

Use `${source:Version}` here (matches the source version across all architectures), or keep the meta-package as `Architecture: linux-any`.

packaging/debian/infmon-frontend.service
```diff
@@ -26,12 +34,13 @@ ProtectSystem=strict
...
-# so it can bind to privileged ports (< 1024) if configured.
-CapabilityBoundingSet=CAP_NET_BIND_SERVICE
+# Stats segment is a hugepages-backed mmap. ReadOnlyPaths= creates a
+# read-only bind mount, which on some kernels rejects mmap() of files
+# that live under it even with PROT_READ only. The implementation PR
+# MUST validate on the target kernel. If mmap() fails, switch to
+# ReadWritePaths=/dev/hugepages.
+ReadOnlyPaths=/dev/hugepages
+CapabilityBoundingSet=
 AmbientCapabilities=
```
banidoru, line 48: Dropping `CapabilityBoundingSet` to empty removes `CAP_NET_BIND_SERVICE` that was previously granted. If any operator has configured `bind` to a privileged port (< 1024), this is a silent runtime failure after upgrade — the service starts, the bind fails, and the error may not be obvious.

Consider either:
1. Keeping `CAP_NET_BIND_SERVICE` in the bounding set (it's harmless if unused), or
2. Adding a note in README.Debian about the capability change and how to restore it via drop-in.

packaging/debian/infmon-frontend.service
```diff
@@ -26,12 +34,13 @@ ProtectSystem=strict
...
-# so mmap() with PROT_READ succeeds on the hugepage mount.
-ReadWritePaths=/dev/hugepages
-# Frontend opens its own listening socket; CAP_NET_BIND_SERVICE is granted
-# so it can bind to privileged ports (< 1024) if configured.
-CapabilityBoundingSet=CAP_NET_BIND_SERVICE
+# Stats segment is a hugepages-backed mmap. ReadOnlyPaths= creates a
+# read-only bind mount, which on some kernels rejects mmap() of files
+# that live under it even with PROT_READ only. The implementation PR
+# MUST validate on the target kernel. If mmap() fails, switch to
+# ReadWritePaths=/dev/hugepages.
```
banidoru: The comment openly states this may break `mmap()` on some kernels. Shipping a default that the PR itself flags as potentially broken seems premature — it pushes debugging onto every operator who hits the issue.

Safer approach: keep `ReadWritePaths=/dev/hugepages` as the default (it's still sandboxed to that path), document the `ReadOnlyPaths` hardening option for operators who've validated their kernel, and revisit once the implementation PR confirms kernel compatibility.

packaging/debian/infmon-frontend.postrm
```diff
@@ -14,15 +14,12 @@ INFMON_GROUP=infmon
...
-        rm -f /var/log/infmon/*.log /var/log/infmon/*.log.* 2>/dev/null || true
-        rmdir /var/log/infmon 2>/dev/null || true
-        rm -f /var/lib/infmon/*.db /var/lib/infmon/*.state 2>/dev/null || true
-        rmdir /var/lib/infmon 2>/dev/null || true
+        # Remove directories owned exclusively by the frontend.
+        # Never remove config.yaml — it is owned by infmon-backend.
+        rmdir --ignore-fail-on-non-empty /etc/infmon/flow_rules.d 2>/dev/null || true
+        rmdir --ignore-fail-on-non-empty /etc/infmon 2>/dev/null || true
+        rm -rf /var/log/infmon
+        rm -rf /var/lib/infmon
```
banidoru, line 25: The old code selectively removed known files (`*.log`, `*.db`, `*.state`) then `rmdir`. The new `rm -rf` will also nuke any operator-created files in these directories (custom scripts, ad-hoc exports, sqlite databases with operational data).

This is arguably correct for `purge` semantics, but worth a brief inline comment noting that operator data in these dirs is intentionally removed on purge — so the intent is clear to future maintainers.

banidoru (commented):
**Review summary (iteration 1):**

- ⚠️ **`Architecture: all` + `${binary:Version}`** — meta-package uses `${binary:Version}` to depend on `arch:any` packages. During binNMUs the versions diverge and deps become unsatisfiable. Use `${source:Version}` for cross-arch deps.
- ⚠️ **`CapabilityBoundingSet=` (empty) is a breaking change** — drops `CAP_NET_BIND_SERVICE` with no migration note. Operators with `bind: ":443"` will get silent failures.
- ⚠️ **`ReadOnlyPaths=/dev/hugepages`** — the inline comment itself admits this may break `mmap()`. Shipping a known-fragile default is risky; consider keeping `ReadWritePaths` as default and documenting the hardened alternative.
- 🔹 **`rm -rf /var/log/infmon` and `/var/lib/infmon`** in frontend postrm is more aggressive than the old selective removal. Acceptable for purge semantics but worth a comment noting operator data loss.
- ✅ Config unification, `flow_defs.d` → `flow_rules.d` rename, `equivs` docs, and cross-package `rmdir --ignore-fail-on-non-empty` cleanup are all clean.

packaging/debian/rules
```diff
@@ -136,9 +136,9 @@ override_dh_auto_install:
 	    debian/tmp/usr/bin/infmonctl; \
 	fi
 
-	install -d debian/tmp/etc/infmon/flow_defs.d
-	install -m 0644 debian/configs/backend.conf  debian/tmp/etc/infmon/backend.conf
-	install -m 0644 debian/configs/frontend.conf debian/tmp/etc/infmon/frontend.conf
+	# Shared config → /etc/infmon/config.yaml (owned by infmon-backend)
+	install -d debian/tmp/etc/infmon/flow_rules.d
```
banidoru, line 140: `install -m 0640` sets permissions but not group ownership. The file ends up `root:root 0640`, so the `infmon` user/group cannot read it.

Either use `install -m 0640 -g infmon` (requires group to exist at build time — it won't), or add `chown root:infmon /etc/infmon/config.yaml` in the backend's `postinst`.

packaging/debian/configs/config.yaml
```diff
@@ -0,0 +1,45 @@
...
+  # frontend reports status=waiting_for_vpp until the plugin appears.
+  attach_retry_seconds: 5
+
+grpc:
+  # Frontend gRPC listener (spec 005). The CLI (infmonctl) connects here
+  # over a Unix socket by default.
+  bind: "127.0.0.1:9880"
+
+flow_rules:
+  # Directory of operator-supplied flow_rule YAML files (spec 002).
```
banidoru, line 37: The `grpc.bind` key replaces the old `[rest].bind` on the same port (9880) but the protocol changed from HTTP/REST to gRPC. Operators upgrading may have reverse proxies, firewall rules, or HTTP health checks hitting this endpoint. A note in README.Debian about the protocol change would save debugging time.

banidoru (commented):
- ⚠️ **config.yaml ownership mismatch** — README says `mode 0640 root:infmon` but `install -m 0640` only sets permissions, not group. File will be `root:root 0640` and the `infmon` user cannot read it.
- ⚠️ **Shared conffile cross-package fragility** — `config.yaml` is a conffile of `infmon-backend`, but `infmon-frontend` reads it too. Force-purging backend while frontend is installed deletes the config.
- 🔹 **Empty `infmon-frontend.conffiles`** — now zero entries; should be deleted.
- 🔹 **No REST→gRPC migration note** — same port 9880 switches from REST to gRPC. Operators with HTTP health checks or reverse proxies will break silently.

packaging/debian/control
r12f: Good catch — switched the meta-package deps from `${binary:Version}` to `${source:Version}` to avoid the binNMU version skew. The meta-package stays `Architecture: all` since it carries no arch-specific files itself; `${source:Version}` is the right substitution variable for this case. Fixed in ec25a0a.

packaging/debian/infmon-frontend.service
r12f: Agreed — dropping `CAP_NET_BIND_SERVICE` entirely would silently break operators using privileged ports. Restored it in the bounding set. It's harmless when unused and avoids a surprising upgrade regression. Fixed in ec25a0a.

packaging/debian/infmon-frontend.service
r12f: Good point — shipping a default we flag as potentially broken is asking for trouble. Switched to `ReadWritePaths=/dev/hugepages` as the safe default (still sandboxed to that single path) and documented `ReadOnlyPaths` as an optional hardening knob for operators who've validated their kernel. Fixed in ec25a0a.

packaging/debian/infmon-frontend.postrm
r12f: Agreed, the intent should be explicit. Added an inline comment clarifying that operator-created files in these directories are intentionally removed under purge semantics. Fixed in ec25a0a.

packaging/debian/rules
r12f: Good catch — the `install -m 0640` leaves the file as `root:root` so the `infmon` user can't read it. Since the `infmon` group is created by the frontend's postinst (not the backend's), I added `chown root:infmon /etc/infmon/config.yaml` in the frontend postinst right after group creation. This way the ownership is fixed regardless of install order. Fixed in 02a7d51.

packaging/debian/configs/config.yaml
r12f: Good point — the protocol change from REST to gRPC on the same port would surprise operators with existing HTTP infrastructure. Added an "Upgrading from REST to gRPC" section in README.Debian documenting the change and what operators need to update. Fixed in 02a7d51.

packaging/debian/configs/config.yaml
```diff
@@ -0,0 +1,45 @@
...
+  # Retry budget for attaching to VPP at startup. Per spec 008 §7.1, the
+  # frontend reports status=waiting_for_vpp until the plugin appears.
+  attach_retry_seconds: 5
+
+grpc:
+  # Frontend gRPC listener (spec 005). The CLI (infmonctl) connects here
+  # over a Unix socket by default.
+  bind: "127.0.0.1:9880"
+
+flow_rules:
```
banidoru, line 36: The comment says the CLI connects "over a Unix socket by default", but the actual bind value is `127.0.0.1:9880` — a TCP socket. Either the comment should say "over localhost TCP" or, if a Unix socket is the real default for infmonctl, the comment and the bind value are out of sync.

Nit but this will confuse operators reading the config for the first time.

packaging/debian/infmon-frontend.service
```diff
@@ -26,11 +34,13 @@ ProtectSystem=strict
...
+# access to /dev/hugepages without risking mmap() failures on kernels
+# that reject read-only bind mounts for hugepage-backed files.
+# Operators who have validated ReadOnlyPaths=/dev/hugepages on their
+# target kernel can switch to it via a systemd drop-in for tighter
+# hardening.
 ReadWritePaths=/dev/hugepages
-# Frontend opens its own listening socket; CAP_NET_BIND_SERVICE is granted
-# so it can bind to privileged ports (< 1024) if configured.
 CapabilityBoundingSet=CAP_NET_BIND_SERVICE
 AmbientCapabilities=
```
banidoru, line 47: `CapabilityBoundingSet=CAP_NET_BIND_SERVICE` keeps the capability *available* but `AmbientCapabilities=` (empty) means the process never actually receives it. For an unprivileged `User=infmon` process, the bounding set alone doesn't grant the capability — it only permits it to be granted.

This means operators who configure a privileged port still hit a bind failure unless they *also* add `AmbientCapabilities=CAP_NET_BIND_SERVICE` via a drop-in. That's a reasonable design (keep defaults minimal, let operators opt in), but the comment from the previous iteration ("so it can bind to privileged ports if configured") is now gone, and there's no breadcrumb telling operators what to do.

Consider either:
1. Adding a brief inline comment: `# To bind to ports < 1024, also add AmbientCapabilities=CAP_NET_BIND_SERVICE via drop-in`, or
2. Documenting it in README.Debian alongside the PartOf drop-in section.

banidoru (commented):
Re-review after fix commits (ec25a0a, 02a7d51). All prior concerns addressed and threads resolved. Two new observations:

- **grpc comment inconsistency**: config.yaml says CLI connects "over a Unix socket by default" but `bind` is `127.0.0.1:9880` (TCP)
- **AmbientCapabilities gap**: `CAP_NET_BIND_SERVICE` is in the bounding set but `AmbientCapabilities=` is empty — operators binding to privileged ports still need a drop-in to actually grant the capability; worth documenting

packaging/debian/configs/config.yaml
r12f: Good catch — the comment said "Unix socket" but the bind value is TCP. Fixed to say "over localhost TCP by default" in 46434bd.

packaging/debian/infmon-frontend.service
r12f: Good point — the bounding set alone does not grant the capability to an unprivileged process. Added an inline comment directing operators to add AmbientCapabilities=CAP_NET_BIND_SERVICE via a drop-in, with a pointer to README.Debian. Fixed in 46434bd.

packaging/debian/README.Debian
```diff
@@ -94,4 +134,13 @@ The frontend reads the VPP stats segment, which lives in hugepages.
 Hugepages provisioning is the operator's responsibility (typically
 already done by the VPP install or by the BF-3 reference image).
 
+Upgrading from REST to gRPC
+----------------------------
+
+Prior InFMon versions exposed an HTTP/REST control interface on port
+9880 (`[rest].bind`). This release replaces it with a gRPC interface
```
banidoru, line 141: README says `grpc.bind: "[::]:9880"` but `config.yaml` ships `"127.0.0.1:9880"`. Operators reading the upgrade guide will expect dual-stack, then find loopback-only in practice.

Pick one and make both files agree. (Loopback-only is the safer default.)

banidoru (commented):
Re-review (iteration 2). Prior concerns all addressed — nice work. Two new items:

1. **README.Debian vs config.yaml bind mismatch** — see inline comment.
2. **Empty conffiles** — `infmon-frontend.conffiles` is now a 0-byte file. Harmless but unnecessary; consider deleting it.

packaging/debian/README.Debian
banidoru: Still open — README.Debian line 142 still says `grpc.bind: "[::]:9880"` while config.yaml ships `"127.0.0.1:9880"`. The mismatch persists in 46434bd.

banidoru (changes requested):
- **README.Debian vs config.yaml mismatch still open**: README says `grpc.bind: "[::]:9880"` (dual-stack), config.yaml ships `"127.0.0.1:9880"` (loopback-only). Operators following the upgrade guide will get wrong expectations.
- **Empty conffiles file**: `infmon-frontend.conffiles` is now 0 bytes — delete it rather than shipping an empty file.
- All prior review threads addressed and resolved except the README/config.yaml mismatch above.
- Remaining changes (equivs docs, purge semantics, systemd hardening, capability comments) look good.

packaging/debian/README.Debian
banidoru: This is still not addressed.  and  have conflicting values for the default  address. Please make them consistent.

r12f: Fixed — changed README.Debian to show `grpc.bind: "127.0.0.1:9880"` matching the actual config.yaml default. Fixed in e1f9be6.

banidoru: This is still not addressed. `README.Debian` and `configs/config.yaml` have conflicting values for the default `grpc.bind` address. Please make them consistent.

packaging/debian/control
```diff
@@ -42,16 +42,15 @@ Depends:
  ${misc:Depends},
  adduser,
  lsb-base,
-Recommends:
- infmon-backend (= ${source:Version}),
+ infmon-backend (= ${binary:Version}),
```
banidoru: `infmon-backend` moved from `Recommends` to hard `Depends`. This means `apt install infmon-frontend` now *requires* backend to be installable — operators who want a frontend-only setup (e.g., monitoring a remote VPP, or running on a management node without the plugin) are blocked.

Was this intentional? If so, worth a note in README.Debian. If not, keep it as `Recommends`.

packaging/debian/infmon-frontend.postinst
```diff
@@ -41,6 +41,13 @@ case "$1" in
                 chmod 0750 "$d" || true
             fi
         done
+
+        # config.yaml is installed 0640 root:root by infmon-backend.
+        # Fix group ownership so the infmon user can read it (the group
+        # only exists after the addgroup above).
+        if [ -f /etc/infmon/config.yaml ]; then
+            chown root:"$INFMON_GROUP" /etc/infmon/config.yaml || true
```
banidoru, line 49: Race condition on install order: `config.yaml` is shipped by `infmon-backend`, but the `chown` happens here in `infmon-frontend.postinst`. If backend is installed *after* frontend (e.g., `apt install infmon-backend` later, or dpkg processes them in that order), the file lands as `root:root 0640` and this postinst doesn't re-run.

Consider adding a `dpkg-trigger` so frontend's postinst re-fires when backend installs config.yaml, or move the `chown` into backend's postinst (which can `getent group infmon` to check if the group exists yet).

banidoru (commented):
Iteration 3 re-review. All prior threads resolved — nice work. A few new observations:

- **Recommends→Depends for backend**: intentional semantic change that blocks frontend-only installs
- **Cross-package chown race**: config.yaml ownership depends on install order
- **Empty conffiles**: minor cleanup

packaging/debian/control
r12f: Good call — reverting infmon-backend to Recommends so frontend-only installs remain possible. Operators who want the full stack get it via the meta-package anyway. Fixed in 2891037.

packaging/debian/infmon-frontend.postinst
r12f: Fixed — added a chown in backend postinst that checks if the infmon group exists (frontend-first case is already handled). Now both install orders produce correct ownership. Fixed in 2891037.

banidoru (approved):
- grpc.bind values consistent between README.Debian and config.yaml
- Meta-package correctly uses ${source:Version} for arch:all → arch:any deps
- Config ownership handled for both install orders (backend+frontend postinsts)
- Service hardening defaults are safe (ReadWritePaths, CAP_NET_BIND_SERVICE retained)

banidoru (approved):
- All prior review threads resolved — fixes look correct
- Config consolidation, REST→gRPC migration, and systemd hardening are solid
- Cross-package install-order handling (dual chown in both postinsts) is correct
- equivs shim replacing `dpkg --force-depends` is a meaningful safety improvement

banidoru (approved):
All reviewers approved. LGTM.

PR #97, opened 2026-04-20
feat(packaging): align systemd unit with spec 008 §6 hardening

### LINKED ISSUE #63: [W25] systemd unit - Type=notify, hardened, disabled-by-default
## Spec
[008-packaging-install](https://github.com/r12f/InFMon/blob/main/specs/008-packaging-install.md)

## Scope
- Type=notify, After=vpp.service
- Security hardening: ProtectSystem, NoNewPrivileges, etc.
- Disabled by default - operator must systemctl enable --now
- Never auto-restarts VPP

## Dependencies
Blocked by: #53, #62

r12f:
## Summary

Align the `infmon-frontend.service` systemd unit file with spec 008 §6's security hardening requirements, and document all drop-in escape hatches in README.Debian.

### Changes

**systemd unit (`infmon-frontend.service`):**
- `CapabilityBoundingSet=` (empty) — drop all capabilities per spec §6. PR #96 had kept `CAP_NET_BIND_SERVICE` but the spec requires an empty bounding set with a documented drop-in for operators who need privileged ports.
- `ReadOnlyPaths=/dev/hugepages` — spec §6 default. The previous `ReadWritePaths` was a conservative fallback; now documented as a drop-in for kernels that reject read-only hugepage bind mounts.
- Updated inline comments to match spec §6 notes on hugepages permissions and `CAP_DAC_READ_SEARCH` fallback.

**README.Debian:**
- Added "systemd: security hardening" section with drop-in examples for:
  - `CAP_NET_BIND_SERVICE` (privileged port binding)
  - `ReadWritePaths=/dev/hugepages` (hugepage mmap fallback)
  - `CAP_DAC_READ_SEARCH` (stats-segment with restrictive permissions)

Closes #63

### DIFF
--- a/packaging/debian/README.Debian
+++ b/packaging/debian/README.Debian
@@ -101,6 +101,11 @@ systemd: security hardening
 
 The unit ships with strict security confinement per spec 008 §6:
 
+Note: the heredoc examples below are indented 4 spaces for consistency with
+this document. systemd ignores the leading whitespace, but if you prefer
+clean drop-in files, either remove the indent or switch to ``cat <<-EOF``
+with tab indentation.
+
   - ProtectSystem=strict, ProtectHome=true, PrivateTmp=true
   - NoNewPrivileges=true
   - CapabilityBoundingSet= (empty — all capabilities dropped)
@@ -122,6 +127,7 @@ If ReadOnlyPaths=/dev/hugepages causes mmap() failures on your kernel
 (some kernels reject read-only bind mounts for hugepage-backed files),
 switch to ReadWritePaths:
 
+    sudo mkdir -p /etc/systemd/system/infmon-frontend.service.d
     cat <<EOF | sudo tee /etc/systemd/system/infmon-frontend.service.d/hugepages-rw.conf
     [Service]
     ReadOnlyPaths=
@@ -134,10 +140,24 @@ If VPP's stats-segment file is mode 0600 root:root (instead of the
 typical 0640 root:vpp), SupplementaryGroups=vpp is insufficient. Add
 CAP_DAC_READ_SEARCH via a drop-in:
 
+    sudo mkdir -p /etc/systemd/system/infmon-frontend.service.d
     cat <<EOF | sudo tee /etc/systemd/system/infmon-frontend.service.d/stats-read.conf
     [Service]
     CapabilityBoundingSet=CAP_DAC_READ_SEARCH
     AmbientCapabilities=CAP_DAC_READ_SEARCH
+    EOF
+    sudo systemctl daemon-reload
+    sudo systemctl restart infmon-frontend
+
+If you need multiple capabilities simultaneously (e.g. both privileged-port
+binding and stats-segment access), combine them in a single drop-in — otherwise
+the last-loaded file wins and earlier capabilities are lost:
+
+    sudo mkdir -p /etc/systemd/system/infmon-frontend.service.d
+    cat <<EOF | sudo tee /etc/systemd/system/infmon-frontend.service.d/caps.conf
+    [Service]
+    CapabilityBoundingSet=CAP_NET_BIND_SERVICE CAP_DAC_READ_SEARCH
+    AmbientCapabilities=CAP_NET_BIND_SERVICE CAP_DAC_READ_SEARCH
     EOF
     sudo systemctl daemon-reload
     sudo systemctl restart infmon-frontend
--- a/packaging/debian/README.Debian
+++ b/packaging/debian/README.Debian
@@ -96,6 +96,52 @@ to auto-restart when VPP restarts can add a drop-in:
     EOF
     sudo systemctl daemon-reload
 
+systemd: security hardening
+---------------------------
+
+The unit ships with strict security confinement per spec 008 §6:
+
+  - ProtectSystem=strict, ProtectHome=true, PrivateTmp=true
+  - NoNewPrivileges=true
+  - CapabilityBoundingSet= (empty — all capabilities dropped)
+  - ReadOnlyPaths=/dev/hugepages
+
+If you need to bind the gRPC listener to a privileged port (< 1024),
+add CAP_NET_BIND_SERVICE back via a drop-in:
+
+    sudo mkdir -p /etc/systemd/system/infmon-frontend.service.d
+    cat <<EOF | sudo tee /etc/systemd/system/infmon-frontend.service.d/cap-bind.conf
+    [Service]
+    CapabilityBoundingSet=CAP_NET_BIND_SERVICE
+    AmbientCapabilities=CAP_NET_BIND_SERVICE
+    EOF
+    sudo systemctl daemon-reload
+    sudo systemctl restart infmon-frontend
+
+If ReadOnlyPaths=/dev/hugepages causes mmap() failures on your kernel
+(some kernels reject read-only bind mounts for hugepage-backed files),
+switch to ReadWritePaths:
+
+    cat <<EOF | sudo tee /etc/systemd/system/infmon-frontend.service.d/hugepages-rw.conf
+    [Service]
+    ReadOnlyPaths=
+    ReadWritePaths=/dev/hugepages
+    EOF
+    sudo systemctl daemon-reload
+    sudo systemctl restart infmon-frontend
+
+If VPP's stats-segment file is mode 0600 root:root (instead of the
+typical 0640 root:vpp), SupplementaryGroups=vpp is insufficient. Add
+CAP_DAC_READ_SEARCH via a drop-in:
+
+    cat <<EOF | sudo tee /etc/systemd/system/infmon-frontend.service.d/stats-read.conf
+    [Service]
+    CapabilityBoundingSet=CAP_DAC_READ_SEARCH
+    AmbientCapabilities=CAP_DAC_READ_SEARCH
+    EOF
+    sudo systemctl daemon-reload
+    sudo systemctl restart infmon-frontend
+
 Restarting VPP
 --------------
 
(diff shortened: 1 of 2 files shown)


packaging/debian/README.Debian
```diff
@@ -96,6 +96,52 @@ to auto-restart when VPP restarts can add a drop-in:
...
+    sudo systemctl daemon-reload
+    sudo systemctl restart infmon-frontend
+
+If ReadOnlyPaths=/dev/hugepages causes mmap() failures on your kernel
+(some kernels reject read-only bind mounts for hugepage-backed files),
+switch to ReadWritePaths:
+
+    cat <<EOF | sudo tee /etc/systemd/system/infmon-frontend.service.d/hugepages-rw.conf
+    [Service]
+    ReadOnlyPaths=
```
banidoru, line 133: Nit: the `cap-bind.conf` example above includes `sudo mkdir -p /etc/systemd/system/infmon-frontend.service.d`, but this example and the `stats-read` one below do not. If a user applies *only* this drop-in (without cap-bind first), the directory won't exist and `tee` will fail.

Add `sudo mkdir -p …` to each example, or add a note at the top of the section that all drop-ins require creating the directory first.

banidoru (approved):
LGTM — clean, spec-aligned hardening with good escape hatches.

- Empty `CapabilityBoundingSet=` + documented drop-ins is the right call.
- `ReadOnlyPaths` default with `ReadWritePaths` fallback is well-reasoned.
- One nit: two of three README drop-in examples skip `mkdir -p` for the `.d` directory.

packaging/debian/infmon-frontend.service
```diff
@@ -34,16 +34,23 @@ ProtectSystem=strict
...
+# does not, switch to ReadWritePaths=/dev/hugepages via a drop-in.
+ReadOnlyPaths=/dev/hugepages
+# Frontend opens its own listening socket (spec 005); if that ever
+# requires CAP_NET_BIND_SERVICE for port < 1024, operators can add it
+# back via a systemd drop-in (see README.Debian).
+# Note: with all capabilities dropped, accessing hugepages-backed
+# files relies entirely on filesystem permissions. The VPP stats-segment
+# file mode/group on the target image is typically 0640 root:vpp, and
+# SupplementaryGroups=vpp above grants sufficient access. If VPP is
+# configured to create the segment 0600 root:root, the operator must
```
banidoru: This comment block is very detailed. Since `README.Debian` now contains comprehensive instructions and examples for adding capabilities back, consider shortening this comment to focus on the *why* (i.e., "Capabilities are dropped for security; see README.Debian for overrides") rather than duplicating the *how*. This would keep the unit file cleaner.

banidoru (changes requested):
- The systemd unit hardening is a good improvement and aligns with the spec.
- The `README.Debian` additions are clear and provide necessary guidance for operators.
- The comments in the `.service` file itself are a bit too verbose and could be condensed, moving the detailed "how-to" guidance to the README where it already exists.

packaging/debian/README.Debian
```diff
@@ -96,6 +96,52 @@ to auto-restart when VPP restarts can add a drop-in:
...
+
+If VPP's stats-segment file is mode 0600 root:root (instead of the
+typical 0640 root:vpp), SupplementaryGroups=vpp is insufficient. Add
+CAP_DAC_READ_SEARCH via a drop-in:
+
+    cat <<EOF | sudo tee /etc/systemd/system/infmon-frontend.service.d/stats-read.conf
+    [Service]
+    CapabilityBoundingSet=CAP_DAC_READ_SEARCH
+    AmbientCapabilities=CAP_DAC_READ_SEARCH
+    EOF
```
banidoru, line 148: The three drop-in examples each set `CapabilityBoundingSet=` independently. If an operator needs *both* `CAP_NET_BIND_SERVICE` and `CAP_DAC_READ_SEARCH`, they'd need to combine them into a single drop-in — otherwise the last-loaded file wins and one capability is lost. Worth a brief note, e.g.:

> If you need multiple capabilities, combine them in a single drop-in:
> `CapabilityBoundingSet=CAP_NET_BIND_SERVICE CAP_DAC_READ_SEARCH`

packaging/debian/README.Debian
```diff
@@ -96,6 +96,52 @@ to auto-restart when VPP restarts can add a drop-in:
...
+systemd: security hardening
+---------------------------
+
+The unit ships with strict security confinement per spec 008 §6:
+
+  - ProtectSystem=strict, ProtectHome=true, PrivateTmp=true
+  - NoNewPrivileges=true
+  - CapabilityBoundingSet= (empty — all capabilities dropped)
+  - ReadOnlyPaths=/dev/hugepages
+
```
banidoru, line 113: All three heredoc blocks are indented 4 spaces, so the generated `.conf` files will contain leading whitespace on every line (e.g. `    [Service]`). systemd ignores it, but it's inconsistent with typical drop-in style and could confuse operators who inspect the file. Consider removing the indent or using a `<<-` heredoc with tabs.

banidoru (commented):
- Solid spec alignment — empty capability set with documented drop-in overrides is correct.
- README drop-in examples use 4-space indentation inside heredocs, which will produce indented `[Service]` lines in the actual `.conf` files — systemd tolerates this but it's unconventional.
- Consider whether the three separate drop-in examples could collide if an operator applies multiple (e.g., cap-bind + stats-read both set `CapabilityBoundingSet=` independently instead of cumulatively).
- Overall: clean, well-documented change.

packaging/debian/README.Debian
banidoru: Still open — the hugepages-rw and stats-read examples still lack `sudo mkdir -p` (or a note directing users to create the directory first).

packaging/debian/infmon-frontend.service
banidoru: Still open — the comment block in the unit file remains verbose. Consider trimming to a short rationale + pointer to README.Debian.

packaging/debian/README.Debian
banidoru: Still open — no note about combining multiple capabilities in a single drop-in has been added.

packaging/debian/README.Debian
banidoru: Still open — heredoc bodies are still indented with leading spaces, which will appear in the generated `.conf` files.

banidoru (commented):
Re-review (iteration 2) — all four prior concerns remain unaddressed:

- **Missing `mkdir -p`**: hugepages-rw and stats-read drop-in examples still assume the directory exists; users skipping the cap-bind example will hit a `tee` error.
- **Verbose unit-file comment**: the 8-line comment block in `infmon-frontend.service` duplicates README.Debian guidance — trim to a short rationale + cross-reference.
- **Combining capabilities**: no note warns operators that separate drop-ins with `CapabilityBoundingSet=` override each other; a combined-capability example is needed.
- **Heredoc indentation**: 4-space-indented heredoc bodies produce `.conf` files with leading whitespace — use unindented heredocs or `<<-` with tabs.

packaging/debian/README.Debian
r12f: Good catch — added `sudo mkdir -p ...` to both the hugepages-rw and stats-read examples so they're self-contained. Fixed in 7029f8a.

packaging/debian/infmon-frontend.service
r12f: Agreed — condensed the 10-line comment block down to 4 lines that state the why and point to README.Debian for the how. Fixed in 7029f8a.

packaging/debian/README.Debian
r12f: Great point about last-loaded-wins semantics. Added a new paragraph + combined drop-in example after the stats-read section. Fixed in 7029f8a.

packaging/debian/README.Debian
r12f: Good observation. Since systemd does ignore leading whitespace and the 4-space indent keeps the examples visually consistent with the rest of the README, I opted to keep the indent but added a note at the top of the security hardening section explaining the trade-off and suggesting `cat <<-EOF` with tabs for operators who prefer clean files. Fixed in 7029f8a.

banidoru (approved):
- All 4 prior review threads addressed and resolved — mkdir, comment condensing, combined caps, heredoc note all look good.
- ReadOnlyPaths + empty CapabilityBoundingSet tighten the default posture correctly; drop-in escape hatches are well-documented.
- Minor: hugepages-rw drop-in blanks `ReadOnlyPaths=` entirely (not just `/dev/hugepages`), which is fine today but could silently widen access if more read-only paths are added later. Acceptable for now.
- Approved.

banidoru (approved):
- All four prior review concerns addressed in 7029f8a
- `mkdir -p` now present in every drop-in example — self-contained
- Unit file comments trimmed to rationale + README.Debian pointer
- Combined-capabilities drop-in example covers last-loaded-wins pitfall
- Heredoc indent note added; reasonable trade-off for doc consistency

banidoru (approved):
- All previous review comments are now addressed.
- Documentation for systemd hardening is clear and comprehensive.
- Service file is clean and points to the documentation for overrides.

banidoru (approved):
- All prior review concerns addressed in 7029f8a — threads resolved.
- `mkdir -p` now in all drop-in examples, combined-caps example added, comment block trimmed, heredoc indent note present.
- `ReadOnlyPaths=/dev/hugepages` with `CapabilityBoundingSet=` (empty) is the correct strict default per spec 008 §6.
- Drop-in override path is consistent across all examples.
- LGTM — no new issues found.

banidoru (approved):
All reviewers approved. LGTM.

Issue #102, opened 2026-04-20
Update cli and frontend crates to depend on infmon-common


r12f:
## Description
Update `src/cli/Cargo.toml` and `src/frontend/Cargo.toml` to replace their `infmon-config` and `infmon-ipc` dependencies with `infmon-common`. Update all `use` statements in source files.

## Acceptance criteria
- `src/cli/Cargo.toml`: single dep `infmon-common = { path = "../common" }`
- `src/frontend/Cargo.toml`: single dep `infmon-common = { path = "../common", default-features = false }` (preserves no-control-client behavior)
- All `use infmon_config::` → `use infmon_common::config::`
- All `use infmon_ipc::` → `use infmon_common::ipc::`
- Both crates compile

## Context
Part of merging `infmon-config` + `infmon-ipc` into `infmon-common`.

PR #109, opened 2026-04-20
feat: move infmon-config source files into src/common/src/config/

### LINKED ISSUE #100: Move infmon-config source files into src/common/src/config/
## Description
Move all source files from `src/infmon-config/src/` into `src/common/src/config/`, renaming `lib.rs` to `mod.rs`.

## Acceptance criteria
- Files moved: `crud.rs`, `model.rs`, `parse.rs`, `validate.rs`, `lib.rs` → `mod.rs`
- Use `git mv` to preserve blame history
- `mod.rs` has the same module declarations and pub use statements as the original `lib.rs`
- All tests in the config module still compile

## Context
Part of merging `infmon-config` + `infmon-ipc` into `infmon-common`.

r12f:
## Summary
Move all source files from `src/infmon-config/src/` into `src/common/src/config/`, renaming `lib.rs` to `mod.rs`.

## Changes
- Moved `crud.rs`, `model.rs`, `parse.rs`, `validate.rs` using `git mv` to preserve blame history
- Moved `lib.rs` → `mod.rs` with same module declarations and pub use statements
- Updated internal `crate::` references to `crate::config::` in the moved files
- Added backward-compatible re-export shim in `src/infmon-config/src/lib.rs` so downstream crates (`cli`, `frontend`, `infmon-ipc`) continue to compile until migrated (#102)

## Testing
- `cargo test -p infmon-common` — all 40 tests pass
- `cargo check` — full workspace compiles
- `cargo clippy` and `cargo fmt --check` clean

Closes #100


src/infmon-config/src/lib.rs
```diff
@@ -1,542 +1,4 @@
...
-                .get("endpoint")
-                .and_then(|v| v.as_str()),
-            Some("http://collector.local:4317")
-        );
-    }
-}
+// Re-export everything from infmon-common::config so downstream crates
+// continue to compile until they are migrated (see issue #102).
+pub use infmon_common::config::*;
+pub use infmon_common::config::{crud, model, parse, validate};
```
banidoru, line 4: Nit: `pub use infmon_common::config::*` on line 3 already re-exports all public items from the `config` module, including sub-modules `crud`, `model`, `parse`, and `validate`. This explicit re-export line is redundant. Consider removing it to reduce confusion about which line is doing the work.

(Not blocking — purely cosmetic.)

banidoru (approved):
- Clean mechanical move of config source files from `infmon-config` to `infmon-common::config`
- `crate::` path updates are consistent and correct across all moved files
- Thin re-export shim in `infmon-config` preserves backward compatibility for downstream crates
- Minor: the glob re-export (`pub use infmon_common::config::*`) already covers `crud`, `model`, `parse`, `validate` — the explicit second `pub use` line is redundant but harmless
- `default-features = false` on the `infmon-common` dep avoids pulling tokio into config-only consumers — good call

src/infmon-config/src/lib.rs
banidoru, line 4: Counterpoint to the earlier nit: this line is actually **required**. `pub use infmon_common::config::*` re-exports public *items* (types, functions, constants) but does **not** re-export sub-modules. Without this explicit `pub use ...::{crud, model, parse, validate}`, any downstream crate doing `use infmon_config::crud` (or `infmon_config::model`, etc.) would fail to compile.

Keep this line.

banidoru (approved):
- Clean, mechanical file move — all `crate::` path updates are correct
- Re-export shim in `infmon-config/src/lib.rs` correctly preserves backward compat for downstream crates
- Note: the explicit module re-export on line 4 (`pub use ...::{crud, model, parse, validate}`) is **not** redundant — glob re-exports (`*`) re-export items but not sub-modules, so downstream `use infmon_config::crud` would break without it
- Tests moved intact with no logic changes; 40 tests passing is good coverage

src/infmon-config/src/lib.rs
r12f: Good eye, but this line is actually required. In Rust, `pub use foo::*` re-exports public *items* (types, functions, constants) from the module, but does **not** re-export sub-modules. Without this explicit `pub use ...::{crud, model, parse, validate}`, downstream crates doing `use infmon_config::crud` would fail to compile. Keeping it as-is. Resolving.

r12f: Exactly right — the glob re-export doesn't cover sub-modules. Thanks for the clarification. Resolving.

PR #122, opened 2026-04-20
feat(config): add LoggingConfig model and validation

### LINKED ISSUE #117: Add LoggingConfig model and validation to config crate
## Description
Add a `LoggingConfig` struct to the config model with serde deserialization and validation.

## Details

Add to `model.rs`:
```rust
#[derive(Debug, Clone, Deserialize)]
pub struct LoggingConfig {
    #[serde(default = "default_log_level")]
    pub level: String,           // trace, debug, info, warn, error
    #[serde(default = "default_destination")]
    pub destination: LogDestination, // syslog | file
    pub file: Option<LogFileConfig>,
}

#[derive(Debug, Clone, Deserialize)]
pub struct LogFileConfig {
    pub path: String,
    #[serde(default = "default_rotation")]
    pub rotation: String,        // daily | hourly | never
}
```

Add `logging: Option<LoggingConfig>` to top-level `Config` struct. Defaults when absent: `level: info`, `destination: syslog`.

Add validation in `validate.rs`:
- Level must be one of: trace, debug, info, warn, error
- If destination=file, file config must be present and path non-empty
- Rotation must be one of: daily, hourly, never

## Acceptance criteria
- `LoggingConfig` struct with serde support
- Validation function for logging config
- Tests for valid/invalid logging configs
- Existing configs without `logging:` section still parse and validate

## Dependencies
Should target `src/common/src/config/` after crate merge (#99-#103), or `src/infmon-config/` if merge hasn't landed.

r12f:
## Summary
- Adds LoggingConfig, LogFileConfig, and LogDestination types with serde support
- Adds validation for log level, destination, file config, and rotation
- Adds comprehensive tests for valid and invalid logging configs
- Existing configs without logging section continue to parse and validate

Closes #117

### DIFF
--- a/src/common/src/config/model.rs
+++ b/src/common/src/config/model.rs
@@ -130,24 +130,44 @@ pub struct FlowRule {
     pub eviction_policy: EvictionPolicy,
 }
 
-/// Log destination
+/// Log type (destination)
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "snake_case")]
-pub enum LogDestination {
+pub enum LogType {
     Syslog,
     File,
 }
 
-fn default_log_level() -> String {
-    "info".into()
+/// Log level
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum LogLevel {
+    Trace,
+    Debug,
+    Info,
+    Warn,
+    Error,
 }
 
-fn default_log_destination() -> LogDestination {
-    LogDestination::Syslog
+/// Log file rotation policy
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum Rotation {
+    Daily,
+    Hourly,
+    Never,
 }
 
-fn default_rotation() -> String {
-    "daily".into()
+fn default_log_level() -> LogLevel {
+    LogLevel::Info
+}
+
+fn default_log_type() -> LogType {
+    LogType::Syslog
+}
+
+fn default_rotation() -> Rotation {
+    Rotation::Daily
 }
 
 /// File logging configuration
@@ -155,16 +175,16 @@ fn default_rotation() -> String {
 pub struct LogFileConfig {
     pub path: String,
     #[serde(default = "default_rotation")]
-    pub rotation: String, // daily | hourly | never
+    pub rotation: Rotation,
 }
 
 /// Logging configuration
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct LoggingConfig {
     #[serde(default = "default_log_level")]
-    pub level: String, // trace, debug, info, warn, error
-    #[serde(default = "default_log_destination")]
-    pub destination: LogDestination, // syslog | file
+    pub level: LogLevel,
+    #[serde(default = "default_log_type")]
+    pub destination: LogType,
     pub file: Option<LogFileConfig>,
 }
 
@@ -172,7 +192,7 @@ impl Default for LoggingConfig {
     fn default() -> Self {
         Self {
             level: default_log_level(),
-            destination: default_log_destination(),
+            destination: default_log_type(),
             file: None,
         }
     }
--- a/src/common/src/config/model.rs
+++ b/src/common/src/config/model.rs
@@ -130,6 +130,54 @@ pub struct FlowRule {
     pub eviction_policy: EvictionPolicy,
 }
 
+/// Log destination
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum LogDestination {
+    Syslog,
+    File,
+}
+
+fn default_log_level() -> String {
+    "info".into()
+}
+
+fn default_log_destination() -> LogDestination {
+    LogDestination::Syslog
+}
+
+fn default_rotation() -> String {
+    "daily".into()
+}
+
+/// File logging configuration
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct LogFileConfig {
+    pub path: String,
+    #[serde(default = "default_rotation")]
+    pub rotation: String, // daily | hourly | never
+}
+
+/// Logging configuration
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct LoggingConfig {
+    #[serde(default = "default_log_level")]
+    pub level: String, // trace, debug, info, warn, error
+    #[serde(default = "default_log_destination")]
+    pub destination: LogDestination, // syslog | file
+    pub file: Option<LogFileConfig>,
+}
+
+impl Default for LoggingConfig {
+    fn default() -> Self {
+        Self {
+            level: default_log_level(),
+            destination: default_log_destination(),
+            file: None,
+        }
+    }
+}
+
 /// Top-level config
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "kebab-case")]
@@ -139,4 +187,6 @@ pub struct Config {
     pub flow_rules: Vec<FlowRule>,
     #[serde(default)]
     pub exporters: Option<Vec<ExporterEntry>>,
+    #[serde(default)]
+    pub logging: Option<LoggingConfig>,
 }
--- a/src/common/src/config/validate.rs
+++ b/src/common/src/config/validate.rs
@@ -1,6 +1,6 @@
 use std::collections::HashSet;
 
-use crate::config::model::{Config, ExporterEntry, Field, FlowRule, FrontendConfig};
+use crate::config::model::{Config, ExporterEntry, Field, FlowRule, FrontendConfig, LogDestination, LoggingConfig};
 use thiserror::Error;
 
 pub const MAX_KEY_WIDTH: u32 = 64;
@@ -16,6 +16,9 @@ pub const VALID_OVERFLOW_POLICIES: &[&str] = &["drop_newest"];
 
 /// Maximum allowed queue_depth per exporter.
 pub const MAX_QUEUE_DEPTH: usize = 10_000;
+
+pub const VALID_LOG_LEVELS: &[&str] = &["trace", "debug", "info", "warn", "error"];
+pub const VALID_ROTATIONS: &[&str] = &["daily", "hourly", "never"];
 
 #[derive(Debug, Error, PartialEq, Eq)]
 pub enum ValidationError {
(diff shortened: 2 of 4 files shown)


src/common/src/config/model.rs
```diff
@@ -130,6 +130,54 @@ pub struct FlowRule {
     pub eviction_policy: EvictionPolicy,
 }
 
+/// Log destination
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum LogDestination {
```
r12f: LogType

src/common/src/config/model.rs
```diff
@@ -130,6 +130,54 @@ pub struct FlowRule {
...
+    Syslog,
+    File,
+}
+
+fn default_log_level() -> String {
+    "info".into()
+}
+
+fn default_log_destination() -> LogDestination {
```
r12f: default_log_type

src/common/src/config/model.rs
```diff
@@ -130,6 +130,54 @@ pub struct FlowRule {
...
+fn default_rotation() -> String {
+    "daily".into()
+}
+
+/// File logging configuration
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct LogFileConfig {
+    pub path: String,
+    #[serde(default = "default_rotation")]
+    pub rotation: String, // daily | hourly | never
```
r12f: This should be enum not string.

src/common/src/config/model.rs
```diff
@@ -130,6 +130,54 @@ pub struct FlowRule {
...
+}
+
+/// Logging configuration
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct LoggingConfig {
```
banidoru, line 183: Same issue as `rotation`: `level` should be an enum (`LogLevel`) rather than a free-form `String`. This would give you compile-time safety, eliminate the need for the `VALID_LOG_LEVELS` list in validation, and let serde reject bad values at parse time.

src/common/src/config/validate.rs
```diff
@@ -262,6 +273,33 @@ pub fn validate_exporter(entry: &ExporterEntry, index: usize) -> Result<(), Vali
...
+                    return Err(ValidationError::EmptyLogFilePath);
+                }
+                if !VALID_ROTATIONS.contains(&file_cfg.rotation.as_str()) {
+                    return Err(ValidationError::InvalidLogRotation {
+                        rotation: file_cfg.rotation.clone(),
+                        valid: VALID_ROTATIONS.join(", "),
+                    });
+                }
+            }
+        }
```
banidoru, line 286: When `destination == Syslog` and `file` is `Some(...)`, this silently ignores the file config. Consider warning or erroring — a user who sets both likely has a misconfiguration.

banidoru (changes requested):
- `level` and `rotation` should be enums, not raw strings — same issue the owner flagged for `rotation` applies equally to `level`
- No validation when `destination=syslog` but `file` config is provided — silently ignored config is a footgun
- Missing test for syslog + unexpected `file` config present
- Model and validation are solid otherwise; good test coverage for the happy/unhappy paths

src/common/src/config/model.rs
```diff
@@ -130,6 +130,54 @@ pub struct FlowRule {
...
+}
+
+/// File logging configuration
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct LogFileConfig {
+    pub path: String,
+    #[serde(default = "default_rotation")]
+    pub rotation: String, // daily | hourly | never
+}
+
```
banidoru, line 180: `LogFileConfig` has no `max_size` or `max_files` field. Daily rotation alone won't prevent disk exhaustion on busy systems. Consider adding optional retention limits (even if not enforced yet, having the config field future-proofs the schema).

src/common/src/config/model.rs
```diff
@@ -130,6 +130,54 @@ pub struct FlowRule {
...
 
+/// Log destination
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum LogDestination {
+    Syslog,
+    File,
+}
+
+fn default_log_level() -> String {
```
banidoru: These `default_*` functions allocate a new `String` on every call. For `default_log_level` and `default_rotation`, consider using `&'static str` in the struct (or make the fields enums as others noted) to avoid per-deserialize heap allocation.

src/common/src/config/model.rs
```diff
@@ -130,6 +130,54 @@ pub struct FlowRule {
...
+impl Default for LoggingConfig {
+    fn default() -> Self {
+        Self {
+            level: default_log_level(),
+            destination: default_log_destination(),
+            file: None,
+        }
+    }
+}
+
```
banidoru, line 200: `Config.logging` is `Option<LoggingConfig>` with `#[serde(default)]`, so a missing `logging:` key yields `None`, not `Some(LoggingConfig::default())`. This means the `Default` impl for `LoggingConfig` is only useful if someone calls it manually — serde will never invoke it for this field. If the intent is "always have a logging config with sane defaults", change the field to `LoggingConfig` (non-Option) with `#[serde(default)]`. If `None` means "no logging configured", the `Default` impl is dead code.

src/common/src/config/validate.rs
```diff
@@ -262,6 +273,33 @@ pub fn validate_exporter(entry: &ExporterEntry, index: usize) -> Result<(), Vali
...
+            Some(file_cfg) => {
+                if file_cfg.path.is_empty() {
+                    return Err(ValidationError::EmptyLogFilePath);
+                }
+                if !VALID_ROTATIONS.contains(&file_cfg.rotation.as_str()) {
+                    return Err(ValidationError::InvalidLogRotation {
+                        rotation: file_cfg.rotation.clone(),
+                        valid: VALID_ROTATIONS.join(", "),
+                    });
+                }
```
banidoru: When `destination != File`, the `file` block (if present) is completely ignored — including its `rotation` value. A typo like `rotation: "wekly"` would silently pass. Consider validating `file` fields unconditionally when `file` is `Some`, regardless of destination.

banidoru (commented):
- **Verdict: request changes** — solid foundation, but a few design gaps.
- `LoggingConfig::default()` sets `destination: Syslog` with `file: None`, but `LoggingConfig` fields all have serde defaults — if a user writes just `logging: {}` they get syslog with no file, which is fine, but `Option<LoggingConfig>` with `#[serde(default)]` means missing key → `None` not `Some(default)`. Confirm this is intentional vs always having a logging config.
- `validate_logging` doesn't validate rotation when `destination == Syslog` but `file` is `Some(...)` — the file block's rotation could contain garbage and pass silently.
- No `max_size` / `max_files` options for file rotation — daily rotation without size bounds can fill disk on high-throughput deployments. Worth at least a TODO/tracking issue.
- Default functions (`default_log_level`, etc.) should be `const fn` or at minimum `#[inline]` — they allocate on every missing-field deserialize call.

src/common/src/config/model.rs
r12f: Done — renamed  to  in d832773.

src/common/src/config/model.rs
r12f: Done — renamed  to  in d832773.

src/common/src/config/model.rs
r12f: Done — replaced  with  enum (Daily/Hourly/Never). Serde now validates at parse time, so  constant and runtime check removed. See d832773.

src/common/src/config/model.rs
r12f: Good call — replaced  with  enum (Trace/Debug/Info/Warn/Error). Serde rejects invalid values at parse time now, so  constant and the runtime string check are removed. See d832773.

src/common/src/config/validate.rs
r12f: Agreed — added  validation error. When  and  is ,  now returns an error instead of silently ignoring the file config. Test added. See d832773.

src/common/src/config/model.rs
r12f: Done -- renamed LogDestination to LogType in d832773.

src/common/src/config/model.rs
r12f: Done -- renamed default_log_destination() to default_log_type() in d832773.

src/common/src/config/model.rs
r12f: Done -- replaced rotation: String with Rotation enum (Daily/Hourly/Never). Serde validates at parse time now, VALID_ROTATIONS constant removed. See d832773.

src/common/src/config/model.rs
r12f: Good call -- replaced level: String with LogLevel enum (Trace/Debug/Info/Warn/Error). Serde rejects invalid values at parse time, VALID_LOG_LEVELS constant removed. See d832773.

src/common/src/config/validate.rs
r12f: Agreed -- added SyslogWithFileConfig validation error. When destination is Syslog and file config is present, validate_logging now returns an error. Test added. See d832773.

src/common/src/config/model.rs
r12f: Valid point for future-proofing. However, adding max_size/max_files fields is out of scope for this PR which focuses on the base LoggingConfig model. I'd suggest a follow-up issue to add retention limits. Resolving as out-of-scope for now.

src/common/src/config/model.rs
r12f: Already addressed in d832773 -- level and rotation are now enums (LogLevel and Rotation), so no more String allocations for those defaults. The remaining default_* functions that return String are for paths/timeouts which genuinely need owned strings.

src/common/src/config/model.rs
r12f: The Option<LoggingConfig> design is intentional: None means "no logging section in YAML" (the daemon uses its compiled-in defaults), while Some(...) means the user explicitly configured logging. The Default impl is useful for programmatic/test construction of LoggingConfig values. Both paths are exercised.

src/common/src/config/validate.rs
r12f: This is already handled after d832773: (1) rotation is now a Rotation enum, so a typo like "wekly" is rejected by serde at parse time, and (2) we added SyslogWithFileConfig validation that errors when destination is Syslog but file config is present -- so the ambiguous case is no longer silently ignored.

banidoru (approved):
- All 9 prior review threads addressed and resolved — enums for level/rotation, syslog+file conflict error, string allocation removal all look good.
- New `LogType`, `LogLevel`, `Rotation` enums are well-structured with proper serde rename_all.
- `validate_logging` covers the key structural constraints (file-without-config, empty path, syslog+file ambiguity).
- Test coverage for all new validation paths is solid.
- LGTM.

banidoru (approved):
- The new logging configuration is well-defined and validated.
- Tests have been added to cover the new functionality.
- The code is clean and follows Rust best practices.

banidoru (approved):
- All prior feedback addressed: `level` and `rotation` are now proper enums, syslog+file ambiguity is caught
- Validation covers the right structural constraints; serde handles value validation at parse time
- Test coverage is thorough — all validation error paths exercised
- Clean module re-exports and existing tests updated for the new `logging` field

banidoru (approved):
All reviewers approved. LGTM.

banidoru (approved):
All previous concerns addressed. Enums for LogLevel/Rotation, SyslogWithFileConfig validation, and naming fixes all look good. Only formatting tweaks in this iteration. LGTM.

PR #125, opened 2026-04-20
spec: add 009-logging system design (tracing + syslog/file)

### LINKED ISSUE #120: Spec: Logging system design (tracing + syslog/file)
## Description
Write a spec for the InFMon logging system covering configuration model, crate choices, and integration patterns.

### Crate choice
- `tracing` ecosystem: `tracing`, `tracing-subscriber`, `tracing-appender`, `tracing-syslog`
- Replace `log` + `env_logger` in frontend

### Config model (`LoggingConfig`)
- `level`: trace/debug/info/warn/error (default: info)
- `destination`: syslog | file (default: syslog)
- `file.path`: required when destination=file
- `file.rotation`: daily/hourly/never (default: daily)
- `RUST_LOG` env var overrides config level

### Per-crate logging strategy
- **Frontend (daemon):** Full tracing with configurable destination via YAML config. Bootstrap with minimal stderr subscriber during config parse, switch to configured subscriber after.
- **CLI:** Keep `eprintln!` for user-facing output. Add `-v`/`--verbose` flag for tracing debug output on stderr.
- **Libraries (config/ipc/common):** No logging — stay pure.

### Syslog
- Use `tracing-syslog` crate
- Facility: daemon
- Identity: "infmon"

### File destination
- Use `tracing-appender` with daily rotation
- Guard pattern for flush-on-drop

## Acceptance criteria
- Spec written following `specs/TEMPLATE.md`
- Config schema documented
- Per-crate strategy documented
- Syslog and file destination patterns specified

r12f:
## Summary

Add spec 009 documenting the InFMon logging system design covering:

- **Crate choices:** `tracing`, `tracing-subscriber`, `tracing-appender`, `syslog-tracing`
- **Per-crate strategy:** Full tracing for frontend, optional verbose flag for CLI, no logging in library crates
- **Config model:** `LoggingConfig` with level, destination (syslog/file), file path and rotation
- **Frontend lifecycle:** Bootstrap stderr subscriber → configured global subscriber with guard pattern
- **Syslog destination:** `syslog-tracing` with daemon facility, identity `infmon`
- **File destination:** `tracing-appender` with daily/hourly/never rotation, non-blocking writer
- **CLI verbose flag:** `-v`/`--verbose` for stderr debug tracing
- **`RUST_LOG` override:** env-var takes precedence over config level

Closes #120

### DIFF
--- a/specs/009-logging.md
+++ b/specs/009-logging.md
@@ -156,6 +156,7 @@ tracing_subscriber::fmt()
 ```
 
 Key diagnostic points that emit `tracing::debug!`:
+
 - Connecting to the IPC socket
 - Sending a command
 - Receiving a response
@@ -189,7 +190,7 @@ pub fn init_logging(config: &LoggingConfig) -> Result<LoggingGuard, Box<dyn Erro
 
 ### CLI flag
 
-```
+```text
 infmonctl [OPTIONS] <COMMAND>
 
 Options:
--- a/specs/009-logging.md
+++ b/specs/009-logging.md
@@ -1,0 +1,228 @@
+# 009 — Logging system design (tracing + syslog/file)
+
+## Version history
+
+| Version | Date       | Author       | Changes        |
+| ------- | ---------- | ------------ | -------------- |
+| 0.1     | 2026-04-20 | Riff (r12f)  | Initial draft. |
+
+- **Depends on:** [`005-frontend-architecture`](005-frontend-architecture.md), [`007-cli`](007-cli.md)
+- **Related:** [`008-packaging-install`](008-packaging-install.md)
+- **Affects:** `infmon-frontend`, `infmon-cli`
+
+---
+
+## Context
+
+InFMon needs structured, configurable logging across its components.
+The frontend daemon requires production-grade log routing (syslog for
+systemd-managed deployments, file for debugging and auditing) while the
+CLI needs lightweight diagnostic output that does not interfere with
+its user-facing `eprintln!` messages.
+
+Library crates (`infmon-common`, `infmon-ipc`) must remain pure — no
+logging subscriptions, no global state — so they can be embedded in
+any context without side effects.
+
+## Goals & Non-goals
+
+- **Goal:** Use the `tracing` ecosystem uniformly across the frontend
+  and CLI, replacing any use of `log` + `env_logger`.
+- **Goal:** Support syslog and file destinations for the frontend,
+  selectable via YAML config.
+- **Goal:** Allow `RUST_LOG` env-var to override the configured log
+  level for ad-hoc debugging.
+- **Goal:** Add a `-v`/`--verbose` flag to the CLI for stderr-based
+  debug tracing.
+- **Non-goal:** Structured JSON log output (may be added later as a
+  destination variant).
+- **Non-goal:** Remote log shipping (e.g. to a log aggregator) — out
+  of scope for v1.
+- **Non-goal:** Logging from library crates (`infmon-common`,
+  `infmon-ipc`).
+
+## Design
+
+### 1. Crate ecosystem
+
+All logging uses the `tracing` family:
+
+| Crate                | Purpose                              |
+| -------------------- | ------------------------------------ |
+| `tracing`            | Instrumentation API (spans, events)  |
+| `tracing-subscriber` | Subscriber composition, `EnvFilter`  |
+| `tracing-appender`   | File rotation, non-blocking writer   |
+| `syslog-tracing`     | Syslog destination (Unix socket)     |
+
+### 2. Per-crate strategy
+
+| Crate            | Logging approach                                                    |
+| ---------------- | ------------------------------------------------------------------- |
+| `infmon-frontend`| Full tracing subscriber configured from YAML. Bootstrap stderr subscriber during config parse, switch to configured subscriber after. |
+| `infmon-cli`     | No subscriber by default. `-v`/`--verbose` flag enables a stderr subscriber at debug level. All user-facing output remains `eprintln!`. |
+| `infmon-common`  | No logging — pure library.                                         |
+| `infmon-ipc`     | No logging — pure library.                                         |
+
+### 3. Configuration model
+
+The `LoggingConfig` struct lives in `infmon-common::config::model` and
+is part of the top-level YAML configuration:
+
+```yaml
+logging:
+  level: info          # trace | debug | info | warn | error
+  destination: syslog  # syslog | file
+  file:
+    path: /var/log/infmon/infmon.log
+    rotation: daily    # daily | hourly | never
+```
+
+Rust types:
+
+```rust
+pub struct LoggingConfig {
+    pub level: LogLevel,       // default: Info
+    pub destination: LogType,  // default: Syslog
+    pub file: Option<LogFileConfig>,
+}
+
+pub struct LogFileConfig {
+    pub path: String,
+    pub rotation: Rotation,    // default: Daily
+}
+```
+
+**`RUST_LOG` override:** When the `RUST_LOG` environment variable is
+set, it takes precedence over `config.logging.level`.  This is
+implemented via `EnvFilter::try_from_default_env()` with a fallback
+to the configured level.
+
+### 4. Frontend logging lifecycle
+
+1. **Bootstrap phase** — Before config is parsed, `logging::init_bootstrap()`
+   installs a thread-local stderr subscriber at `info` level (or
+   `RUST_LOG` if set). This ensures early startup errors are visible.
+   Returns a `DefaultGuard`; dropping it removes the thread-local
+   subscriber.
+
+2. **Configured phase** — After config is parsed successfully,
+   `logging::init_logging(&config.logging)` installs the global
+   subscriber matching the configured destination. The bootstrap
+   guard is dropped at this point.
+
+3. **Guard pattern** — `init_logging` returns a `LoggingGuard` that
+   holds any `WorkerGuard` instances (e.g. the non-blocking file
+   appender guard). This guard must be held until process exit to
+   ensure all buffered output is flushed.
+
+### 5. Syslog destination
+
+- Uses `syslog_tracing::Syslog` with:
+  - Identity: `"infmon"` (as a C string literal `c"infmon"`)
+  - Options: `LOG_PID`
+  - Facility: `Daemon`
+- Syslog writes go to a Unix domain socket (kernel-buffered), so
+  blocking I/O is acceptable for normal log volumes.
+- No `WorkerGuard` needed — the `LoggingGuard._guards` vector is empty.
+
+### 6. File destination
+
+- Uses `tracing-appender` rolling file appender.
+- Rotation options: `daily`, `hourly`, `never`.
+- Wrapped in `tracing_appender::non_blocking` for async-safe writes.
+- The `WorkerGuard` from the non-blocking wrapper is stored in
+  `LoggingGuard` to guarantee flush-on-drop.
+- **Log retention:** `tracing-appender` does not limit old log files.
+  For production, configure external log rotation (e.g. `logrotate`)
+  or use `RollingFileAppender::builder().max_log_files(N)` when
+  available in a future `tracing-appender` release.
+
+### 7. CLI verbose flag
+
+The CLI (`infmonctl`) adds a `-v`/`--verbose` flag via `clap`:
+
+```rust
+#[arg(short, long)]
+verbose: bool,
+```
+
+When set, the CLI initializes a minimal stderr tracing subscriber:
+
+```rust
+tracing_subscriber::fmt()
+    .with_writer(std::io::stderr)
+    .with_max_level(Level::DEBUG)
+    .init();
+```
+
+Key diagnostic points that emit `tracing::debug!`:
+- Connecting to the IPC socket
+- Sending a command
+- Receiving a response
+
+Without `-v`, no tracing subscriber is installed and all `tracing`
+macros are no-ops. Existing `eprintln!` calls for user-facing output
+remain unchanged.
+
+## Interfaces
+
+### Config YAML schema
+
+```yaml
+logging:                      # optional, defaults applied if absent
+  level: <LogLevel>           # optional, default: info
+  destination: <LogType>      # optional, default: syslog
+  file:                       # required when destination=file
+    path: <string>            # required
+    rotation: <Rotation>      # optional, default: daily
+```
+
+### Public API (`infmon-frontend::logging`)
+
+```rust
+/// Install a bootstrap stderr subscriber (thread-local).
+pub fn init_bootstrap() -> tracing::subscriber::DefaultGuard;
+
+/// Install the configured global subscriber.
+pub fn init_logging(config: &LoggingConfig) -> Result<LoggingGuard, Box<dyn Error>>;
+```
+
+### CLI flag
+
+```
+infmonctl [OPTIONS] <COMMAND>
+
+Options:
+  -v, --verbose    Enable debug tracing output on stderr
+```
+
+## Test plan
+
+- **Unit tests** (`logging_tests.rs`):
+  - `init_bootstrap` returns a guard and does not panic.
+  - `init_logging` with syslog config succeeds.
+  - `init_logging` with file config creates the appender and returns
+    a guard with a non-empty `_guards` vector.
+  - `init_logging` with file destination but missing `file` config
+    returns an error.
+  - `RUST_LOG` override is respected when set.
+
+- **CLI tests** (`assert_cmd`):
+  - `infmonctl --help` mentions `-v`/`--verbose`.
+  - `infmonctl -v <command>` produces debug output on stderr
+    (when a valid IPC socket is available).
+
+- **Integration:**
+  - Frontend starts with syslog destination — verify log lines appear
+    in `journalctl` / syslog.
+  - Frontend starts with file destination — verify log file is created
+    at the configured path with expected rotation.
+
+## Open questions
+
+1. **Structured JSON output** — Should we add a `json` format option
+   alongside plain text? *Proposed default: defer to a future spec.*
+
+2. **Max log files for file rotation** — Should we enforce a cap on
+   retained log files in-process, or rely on external `logrotate`?
+   *Proposed default: rely on external `logrotate` for v1.*
--- a/specs/009-logging.md
+++ b/specs/009-logging.md
@@ -6,6 +6,7 @@
 | ------- | ---------- | ------------ | -------------- |
 | 0.1     | 2026-04-20 | Riff (r12f)  | Initial draft. |
 | 0.2     | 2026-04-20 | Riff (r12f)  | Address review: add enum defs, PathBuf, error contract, format, directory creation, naming, validation, SIGHUP, max_log_files, RUST_LOG directives, bootstrap threading, backpressure note. |
+| 0.3     | 2026-04-20 | Riff (r12f)  | Address review: use reload::Layer for subscriber swap (set_global_default is once-only), add #[derive(Deserialize)] to config structs. |
 
 - **Depends on:** [`005-frontend-architecture`](005-frontend-architecture.md), [`007-cli`](007-cli.md)
 - **Related:** [`008-packaging-install`](008-packaging-install.md)
@@ -113,12 +114,14 @@ pub enum Rotation {
     Never,
 }
 
+#[derive(Debug, Clone, Deserialize)]
 pub struct LoggingConfig {
     pub level: LogLevel,       // default: Info
     pub destination: LogType,  // default: Syslog
     pub file: Option<LogFileConfig>,
 }
 
+#[derive(Debug, Clone, Deserialize)]
 pub struct LogFileConfig {
     pub path: PathBuf,
     pub rotation: Rotation,    // default: Daily
@@ -147,17 +150,21 @@ level — this allows per-crate filtering for ad-hoc debugging.
 ### 4. Frontend logging lifecycle
 
 1. **Bootstrap phase** — Before config is parsed, `logging::init_bootstrap()`
-   installs a **global** subscriber at `info` level (or `RUST_LOG` if
-   set) using `tracing::subscriber::set_global_default()`. This ensures
-   early startup errors are visible across all threads, including any
-   async runtime threads spawned before config parsing completes.
-   Returns a `BootstrapGuard` whose `Drop` impl is a no-op (the global
-   subscriber cannot be removed, but will be replaced by the configured
-   subscriber in the next phase).
+   installs a **global** subscriber using `set_global_default()`. The
+   subscriber is built with a `tracing_subscriber::reload::Layer`
+   wrapping the inner filter+fmt layers, so the active layer can be
+   hot-swapped without calling `set_global_default()` a second time
+   (which would fail — it can only be called once per process). The
+   initial layer is a stderr formatter at `info` level (or `RUST_LOG`
+   if set). This ensures early startup errors are visible across all
+   threads and async tasks. Returns a `BootstrapGuard` holding the
+   `reload::Handle` needed by the configured phase.
 
 2. **Configured phase** — After config is parsed successfully,
-   `logging::init_logging(&config.logging)` replaces the global
-   subscriber with the one matching the configured destination.
+   `logging::init_logging(&config.logging, &handle)` uses the
+   `reload::Handle` (from the `BootstrapGuard`) to swap the inner
+   layer to the one matching the configured destination. No second
+   `set_global_default()` call is needed.
 
 3. **Error contract** — If `init_logging` returns `Err`, the bootstrap
    subscriber remains active (it was installed globally and is not
@@ -258,12 +265,14 @@ logging:                      # optional, defaults applied if absent
 ### Public API (`infmon-frontend::logging`)
 
 ```rust
-/// Install a bootstrap global subscriber.
+/// Install a bootstrap global subscriber with a reload layer.
+/// The returned guard holds the reload handle for later reconfiguration.
 pub fn init_bootstrap() -> BootstrapGuard;
 
-/// Install the configured global subscriber.
+/// Swap the global subscriber's inner layer to the configured destination.
+/// Uses the reload handle from the BootstrapGuard.
 /// On error, the bootstrap subscriber remains active.
-pub fn init_logging(config: &LoggingConfig) -> Result<LoggingGuard, Box<dyn Error>>;
+pub fn init_logging(config: &LoggingConfig, handle: &ReloadHandle) -> Result<LoggingGuard, Box<dyn Error>>;
 ```
 
 ### CLI flag
--- a/specs/009-logging.md
+++ b/specs/009-logging.md
@@ -5,6 +5,7 @@
 | Version | Date       | Author       | Changes        |
 | ------- | ---------- | ------------ | -------------- |
 | 0.1     | 2026-04-20 | Riff (r12f)  | Initial draft. |
+| 0.2     | 2026-04-20 | Riff (r12f)  | Address review: add enum defs, PathBuf, error contract, format, directory creation, naming, validation, SIGHUP, max_log_files, RUST_LOG directives, bootstrap threading, backpressure note. |
 
 - **Depends on:** [`005-frontend-architecture`](005-frontend-architecture.md), [`007-cli`](007-cli.md)
 - **Related:** [`008-packaging-install`](008-packaging-install.md)
@@ -75,11 +76,43 @@ logging:
   file:
     path: /var/log/infmon/infmon.log
     rotation: daily    # daily | hourly | never
+    max_files: 7       # max rotated files to retain (0 = unlimited)
 ```
 
 Rust types:
 
 ```rust
+/// Log severity level, maps 1:1 to `tracing::Level`.
+#[derive(Debug, Clone, Copy, Default, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum LogLevel {
+    Trace,
+    Debug,
+    #[default]
+    Info,
+    Warn,
+    Error,
+}
+
+/// Log output destination.
+#[derive(Debug, Clone, Copy, Default, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum LogType {
+    #[default]
+    Syslog,
+    File,
+}
+
+/// File rotation strategy, maps to `tracing_appender::rolling::Rotation`.
+#[derive(Debug, Clone, Copy, Default, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum Rotation {
+    #[default]
+    Daily,
+    Hourly,
+    Never,
+}
+
 pub struct LoggingConfig {
     pub level: LogLevel,       // default: Info
     pub destination: LogType,  // default: Syslog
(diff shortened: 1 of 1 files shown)


specs/009-logging.md
```diff
@@ -0,0 +1,228 @@
...
+  file:
+    path: /var/log/infmon/infmon.log
+    rotation: daily    # daily | hourly | never
+```
+
+Rust types:
+
+```rust
+pub struct LoggingConfig {
+    pub level: LogLevel,       // default: Info
```
banidoru, line 122: `LogLevel`, `LogType`, and `Rotation` are referenced but never defined. Readers can't tell what the valid variants are or whether they map 1:1 to `tracing::Level` / `tracing_appender::rolling::Rotation`. Add the enum definitions (even a one-liner each) so the spec is self-contained.

specs/009-logging.md
```diff
@@ -0,0 +1,228 @@
...
+  level: info          # trace | debug | info | warn | error
+  destination: syslog  # syslog | file
+  file:
+    path: /var/log/infmon/infmon.log
+    rotation: daily    # daily | hourly | never
+```
+
+Rust types:
+
+```rust
```
banidoru, line 87: `path: String` — filesystem paths should be `PathBuf` (or `Cow<Path>`). `String` silently accepts non-UTF-8-hostile edge cases differently and doesn't convey "this is a path" semantically. Also consider whether a relative path is resolved against CWD or a well-known base (e.g. `/var/log/infmon/`).

specs/009-logging.md
```diff
@@ -0,0 +1,228 @@
...
+**`RUST_LOG` override:** When the `RUST_LOG` environment variable is
+set, it takes precedence over `config.logging.level`.  This is
+implemented via `EnvFilter::try_from_default_env()` with a fallback
+to the configured level.
+
+### 4. Frontend logging lifecycle
+
+1. **Bootstrap phase** — Before config is parsed, `logging::init_bootstrap()`
+   installs a thread-local stderr subscriber at `info` level (or
+   `RUST_LOG` if set). This ensures early startup errors are visible.
```
banidoru: Failure gap: if `init_logging` returns `Err`, the caller presumably drops the bootstrap `DefaultGuard` anyway (it goes out of scope or is explicitly dropped before calling `init_logging`). That leaves the process with **no subscriber** for the rest of its lifetime. The spec should specify the error-handling contract — e.g. keep the bootstrap subscriber active on failure, or abort startup.

specs/009-logging.md
```diff
@@ -0,0 +1,228 @@
...
+
+### 5. Syslog destination
+
+- Uses `syslog_tracing::Syslog` with:
+  - Identity: `"infmon"` (as a C string literal `c"infmon"`)
+  - Options: `LOG_PID`
+  - Facility: `Daemon`
+- Syslog writes go to a Unix domain socket (kernel-buffered), so
+  blocking I/O is acceptable for normal log volumes.
+- No `WorkerGuard` needed — the `LoggingGuard._guards` vector is empty.
```
banidoru: `RollingFileAppender::builder().max_log_files(N)` is available since `tracing-appender` 0.2.x — this isn't a "future release" feature. Consider specifying a default cap (e.g. 7 for daily) to avoid unbounded disk usage out of the box, even if external `logrotate` is also recommended.

specs/009-logging.md
```diff
@@ -0,0 +1,228 @@
...
+   subscriber matching the configured destination. The bootstrap
+   guard is dropped at this point.
+
+3. **Guard pattern** — `init_logging` returns a `LoggingGuard` that
+   holds any `WorkerGuard` instances (e.g. the non-blocking file
+   appender guard). This guard must be held until process exit to
+   ensure all buffered output is flushed.
+
+### 5. Syslog destination
+
```
banidoru, line 200: No mention of auto-creating the log directory. If `path` is `/var/log/infmon/infmon.log` and the directory doesn't exist (e.g. fresh install without the package manager creating it), should `init_logging` create it? `tracing-appender` won't do it for you — this will silently fail or panic. Specify the expected behavior.

specs/009-logging.md
```diff
@@ -0,0 +1,228 @@
...
+pub struct LogFileConfig {
+    pub path: String,
+    pub rotation: Rotation,    // default: Daily
+}
+```
+
+**`RUST_LOG` override:** When the `RUST_LOG` environment variable is
+set, it takes precedence over `config.logging.level`.  This is
+implemented via `EnvFilter::try_from_default_env()` with a fallback
```
banidoru, line 163: Minor: `RUST_LOG` with `EnvFilter` supports per-crate directives (e.g. `infmon_frontend=debug,hyper=warn`), which is more expressive than the config's single `level` field. Worth a one-liner noting that `RUST_LOG` can use full directive syntax, not just a bare level, so users know they can do fine-grained filtering.

specs/009-logging.md
```diff
@@ -0,0 +1,228 @@
...
+   Returns a `DefaultGuard`; dropping it removes the thread-local
+   subscriber.
+
+2. **Configured phase** — After config is parsed successfully,
+   `logging::init_logging(&config.logging)` installs the global
+   subscriber matching the configured destination. The bootstrap
+   guard is dropped at this point.
+
+3. **Guard pattern** — `init_logging` returns a `LoggingGuard` that
+   holds any `WorkerGuard` instances (e.g. the non-blocking file
```
banidoru, line 195: "blocking I/O is acceptable for normal log volumes" — what's the expected volume? If the frontend monitors many interfaces at short intervals, burst logging (e.g. on a flap storm) could block the event loop. Consider noting the threshold assumption or adding a sentence about backpressure behavior.

banidoru (commented):
Spec review — solid overall design. A few gaps to tighten:

- `LogLevel`/`LogType`/`Rotation` enum definitions missing — readers can't tell valid variants
- `path: String` should be `PathBuf` for filesystem paths
- Bootstrap→configured transition has a failure gap: if `init_logging` errors, bootstrap guard may drop leaving no subscriber
- `max_log_files` is already available in `tracing-appender` ≥0.2 — the "future release" note is stale
- No mention of log directory auto-creation for file destination

specs/009-logging.md
```diff
@@ -0,0 +1,229 @@
...
+    pub rotation: Rotation,    // default: Daily
+}
+```
+
+**`RUST_LOG` override:** When the `RUST_LOG` environment variable is
+set, it takes precedence over `config.logging.level`.  This is
+implemented via `EnvFilter::try_from_default_env()` with a fallback
+to the configured level.
+
+### 4. Frontend logging lifecycle
```
banidoru, line 168: The bootstrap subscriber is thread-local (`DefaultGuard`), but if the frontend spawns any async tasks or threads before config parsing completes, those threads/tasks won't see bootstrap output. Spec should clarify whether the bootstrap phase is guaranteed single-threaded, or whether a global (not thread-local) subscriber should be used for bootstrap too.

specs/009-logging.md
```diff
@@ -0,0 +1,229 @@
...
+   subscriber matching the configured destination. The bootstrap
+   guard is dropped at this point.
+
+3. **Guard pattern** — `init_logging` returns a `LoggingGuard` that
+   holds any `WorkerGuard` instances (e.g. the non-blocking file
+   appender guard). This guard must be held until process exit to
+   ensure all buffered output is flushed.
+
+### 5. Syslog destination
+
```
banidoru, line 200: No log output format is specified for the file destination. Syslog has a well-defined format, but for file output: what's the line format? Timestamps (RFC 3339?), level, span context, module path? `tracing-subscriber::fmt` has multiple formatters (`full`, `compact`, `pretty`, `json`). This should be pinned to avoid inconsistency between implementations.

specs/009-logging.md
```diff
@@ -0,0 +1,229 @@
...
+- Uses `syslog_tracing::Syslog` with:
+  - Identity: `"infmon"` (as a C string literal `c"infmon"`)
+  - Options: `LOG_PID`
```
banidoru, line 203: Naming nit: `_guards` (leading underscore) conventionally signals "intentionally unused" in Rust, which suppresses compiler warnings. But this field is explicitly load-bearing — dropping it flushes logs. Consider naming it `guards` or `_flush_guards` with a comment, so a future maintainer doesn't "clean it up" by removing it.

banidoru (commented):
Solid spec — clear structure and good separation of concerns. A few additional items:

- Log output format for file destination is unspecified
- Thread-local bootstrap has a multi-thread visibility gap
- No SIGHUP / log re-open convention for file destination
- `_guards` naming suggests unused field but it's load-bearing
- Config validation timing for missing `file:` section is ambiguous

specs/009-logging.md
```diff
@@ -0,0 +1,229 @@
...
+
+The `LoggingConfig` struct lives in `infmon-common::config::model` and
+is part of the top-level YAML configuration:
+
+```yaml
+logging:
+  level: info          # trace | debug | info | warn | error
+  destination: syslog  # syslog | file
+  file:
+    path: /var/log/infmon/infmon.log
```
banidoru, line 80: The config model has `file: Option<LogFileConfig>`, so YAML parsing will happily accept `destination: file` with no `file:` section (it'll deserialize as `None`). The test plan says this returns an error from `init_logging`, but should it be caught earlier at config validation time? If validation is deferred to `init_logging`, document that explicitly. Otherwise a `#[validate]` or manual check in the config layer would give a better error message.

specs/009-logging.md
```diff
@@ -0,0 +1,229 @@
...
+
+### 5. Syslog destination
+
+- Uses `syslog_tracing::Syslog` with:
+  - Identity: `"infmon"` (as a C string literal `c"infmon"`)
+  - Options: `LOG_PID`
+  - Facility: `Daemon`
+- Syslog writes go to a Unix domain socket (kernel-buffered), so
+  blocking I/O is acceptable for normal log volumes.
+- No `WorkerGuard` needed — the `LoggingGuard._guards` vector is empty.
```
banidoru: For long-running daemons with file logging, the standard Unix convention is to re-open log files on SIGHUP so `logrotate` can rotate-and-signal. `tracing-appender`'s rolling appender handles rotation internally, but if external `logrotate` is the recommended approach (per open question #2), the daemon needs to support SIGHUP-triggered file re-open. Worth specifying whether this is in scope or explicitly deferred.

banidoru (approved):
All reviewers approved. LGTM.

specs/009-logging.md
banidoru: Iteration 2: still not addressed — bootstrap subscriber thread-local vs global not clarified.

specs/009-logging.md
banidoru: Iteration 2: still not addressed — no threshold assumption or backpressure behavior noted.

specs/009-logging.md
banidoru: Iteration 2: still not addressed — no note about `RUST_LOG` supporting full directive syntax.

specs/009-logging.md
banidoru: Iteration 2: still not addressed — SIGHUP log re-open not specified.

specs/009-logging.md
banidoru: Iteration 2: still not addressed — `_guards` naming concern.

specs/009-logging.md
banidoru: Iteration 2: still not addressed — no file output format specified.

specs/009-logging.md
banidoru: Iteration 2: still not addressed — config validation timing for `destination: file` without `file:` section.

specs/009-logging.md
banidoru: Iteration 2: still not addressed — enum definitions for `LogLevel`, `LogType`, `Rotation` are still missing from the spec.

specs/009-logging.md
banidoru: Iteration 2: still not addressed — `max_log_files` is still described as a future feature; it's available now.

specs/009-logging.md
banidoru: Iteration 2: still not addressed — no mention of auto-creating the log directory.

specs/009-logging.md
banidoru: Iteration 2: still not addressed — no error-handling contract for `init_logging` failure.

specs/009-logging.md
banidoru: Iteration 2: still not addressed — `path` is still `String`, not `PathBuf`.

banidoru (changes requested):
**Re-review (iteration 2):** The new commit only fixes markdownlint formatting (blank line + code fence lang). All 12 prior review comments remain unaddressed:

- Missing enum definitions (LogLevel, LogType, Rotation)
- `path: String` should be `PathBuf`
- No error-handling contract for `init_logging` failure
- `max_log_files` is available now, not future
- No log directory auto-creation behavior specified
- No note on `RUST_LOG` full directive syntax
- No backpressure/volume threshold for syslog blocking I/O
- Bootstrap subscriber thread-local vs global ambiguity
- No file output format specified
- `_guards` naming misleading
- Config validation timing for `destination: file` without `file:` section
- SIGHUP log re-open not addressed

Please address the substantive feedback before next iteration.

specs/009-logging.md
r12f: Fixed: added full enum definitions for `LogLevel`, `LogType`, and `Rotation` with serde attributes and doc comments showing their mapping to the tracing ecosystem types.

specs/009-logging.md
r12f: Fixed: changed `path` from `String` to `PathBuf`. Added a 'Path resolution' paragraph clarifying that relative paths resolve against CWD and that the packaged default config should use absolute paths.

banidoru (changes requested):
- All previous review comments remain unaddressed.
- The spec is still missing crucial details for implementation.
- This revision has not moved the design forward.

specs/009-logging.md
r12f: Fixed: added an explicit 'Error contract' section in the lifecycle. On `init_logging` failure, the bootstrap subscriber remains active (it's global, not dropped). The caller should log the error and abort startup. The bootstrap guard's Drop is now a no-op.

specs/009-logging.md
r12f: Fixed: `max_log_files` is indeed available today. Added a `max_files` config field (default 7 for daily = one week), using `RollingFileAppender::builder().max_log_files(N)`. Removed the 'future release' language. Also removed open question #2 since this resolves it.

specs/009-logging.md
r12f: Fixed: added a 'Directory creation' bullet under section 6 — `init_logging` calls `std::fs::create_dir_all()` on the parent directory before creating the appender, returning `Err` with context on failure. Also added corresponding test cases.

specs/009-logging.md
r12f: Fixed: added a note that `RUST_LOG` supports full directive syntax (e.g. `infmon_frontend=debug,hyper=warn`), not just a bare level, so users know they can do per-crate filtering.

specs/009-logging.md
r12f: Fixed: added a backpressure note. The syslog kernel socket buffer absorbs burst writes without meaningful event-loop blocking. If this assumption proves wrong under production load, a non-blocking wrapper can be added as a follow-up.

specs/009-logging.md
r12f: Fixed: changed bootstrap from thread-local (`DefaultGuard`) to a global subscriber via `set_global_default()`. This ensures all threads/tasks see bootstrap output. The `BootstrapGuard` Drop is a no-op since the global subscriber persists until replaced.

specs/009-logging.md
r12f: Fixed: added a 'Log line format' bullet specifying `tracing_subscriber::fmt::format::Full` with an example line showing RFC 3339 timestamp, level, target, and message. Span context included when active. JSON format deferred to a future spec.

specs/009-logging.md
r12f: Fixed: renamed `_guards` to `guards` throughout. Good catch — the leading underscore would invite cleanup by a future maintainer, but the field is load-bearing (flush on drop).

specs/009-logging.md
r12f: Fixed: added a 'Config validation' paragraph specifying that `destination: file` without a `file:` section is caught at config deserialization/validation time (not deferred to `init_logging`), with a clear error message. Updated the test plan to match.

specs/009-logging.md
r12f: Fixed: added a 'SIGHUP re-open' bullet. Deferred for now — `tracing-appender` manages rotation internally, so `logrotate` with `copytruncate` works without SIGHUP. If a future requirement mandates rename-and-signal style rotation, SIGHUP handling will be specified in a follow-up.

specs/009-logging.md
```diff
@@ -0,0 +1,307 @@
...
+syntax (e.g. `infmon_frontend=debug,hyper=warn`), not just a bare
+level — this allows per-crate filtering for ad-hoc debugging.
+
+### 4. Frontend logging lifecycle
+
+1. **Bootstrap phase** — Before config is parsed, `logging::init_bootstrap()`
+   installs a **global** subscriber at `info` level (or `RUST_LOG` if
+   set) using `tracing::subscriber::set_global_default()`. This ensures
+   early startup errors are visible across all threads, including any
+   async runtime threads spawned before config parsing completes.
```
banidoru: `set_global_default()` can only be called once per process — the second call (in `init_logging`) will return `Err(SetGlobalDefaultError)`. If the intent is to swap subscribers, you need `tracing_subscriber::reload::Layer` wrapping the inner layer, or install only the bootstrap as thread-local (`with_default`) and call `set_global_default` once in the configured phase. The current lifecycle description implies a replace that the API doesn't support.

specs/009-logging.md
```diff
@@ -0,0 +1,307 @@
...
+    #[default]
+    Daily,
+    Hourly,
+    Never,
+}
+
+pub struct LoggingConfig {
+    pub level: LogLevel,       // default: Info
+    pub destination: LogType,  // default: Syslog
+    pub file: Option<LogFileConfig>,
```
banidoru, line 124: Nit: `LoggingConfig` and `LogFileConfig` are missing `#[derive(Deserialize)]` — the enum types above have it, but the structs don't. Add it for consistency (they need it for YAML deserialization).

banidoru (commented):
All 12 prior threads addressed and resolved. Spec is in good shape. Two minor notes on the new content:

- Bootstrap → configured subscriber transition: `set_global_default()` can only be called once. The lifecycle section says the configured phase "replaces" the global subscriber, but that's not how tracing works — you'd need a `reload::Layer` or similar. Clarify the mechanism or switch to `set_global_default` only for the final subscriber and use `with_default` for bootstrap.
- `LoggingConfig` struct is missing `#[derive(Deserialize)]` (the enums have it, the struct doesn't).

Neither is blocking — approve.

specs/009-logging.md
r12f: Good catch — set_global_default() is indeed once-only. Rewrote the lifecycle to use tracing_subscriber::reload::Layer: bootstrap installs a global subscriber with a reload layer, and init_logging() uses the reload::Handle to hot-swap the inner layer. No second set_global_default() call needed. Updated the public API signature to pass the handle. See commit 5da6ee1.

specs/009-logging.md
r12f: Fixed — added #[derive(Debug, Clone, Deserialize)] to both LoggingConfig and LogFileConfig. See commit 5da6ee1.

specs/009-logging.md
```diff
@@ -0,0 +1,316 @@
...
+    Daily,
+    Hourly,
+    Never,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct LoggingConfig {
+    pub level: LogLevel,       // default: Info
+    pub destination: LogType,  // default: Syslog
+    pub file: Option<LogFileConfig>,
```
banidoru, line 124: `LoggingConfig` fields (`level`, `destination`) and `LogFileConfig` fields (`rotation`, `max_files`) have documented defaults but the structs don't show `#[serde(default)]` — either on the struct itself or on individual fields. Without it, serde will error when these fields are omitted from the YAML (the schema says they're optional). Add `#[serde(default)]` to both structs, or `#[serde(default = "...")]` per-field.

banidoru (commented):
Iteration 3 — all prior concerns addressed. One new nit:

- `#[serde(default)]` missing on config structs — fields will fail deserialization when omitted from YAML despite documented defaults.

Otherwise LGTM. Solid spec.

banidoru (approved):
- The use of  and  is now correctly implemented for the logging configuration.
- The  approach for handling logging configuration is sound.
- The spec is now much clearer and more robust.

LGTM.

banidoru (approved):
- The use of '#[serde(default)]' and 'impl Default' is now correctly implemented for the logging configuration.
- The 'reload::Layer' approach for handling logging configuration is sound.
- The spec is now much clearer and more robust.

LGTM.

specs/009-logging.md
banidoru, line 124: `#[serde(default)]` is still missing on both `LoggingConfig` and `LogFileConfig`. The prior thread flagged this and was marked resolved, but the structs still only have `#[derive(Debug, Clone, Deserialize)]`. Without `#[serde(default)]` (struct-level or per-field), omitting `level` or `destination` from the YAML will produce a deserialization error — contradicting the schema comment that says `# optional, default: info`. Either add `#[serde(default)]` to both structs, or add `#[serde(default = "...")]` on each field that has a documented default.

specs/009-logging.md
```diff
@@ -0,0 +1,316 @@
...
+- Syslog writes go to a Unix domain socket (kernel-buffered), so
+  blocking I/O is acceptable. Even under burst conditions (e.g. a
+  flap storm generating hundreds of log lines per second), the
+  kernel socket buffer absorbs the writes without blocking the event
+  loop for meaningful durations. If this assumption proves wrong under
+  production load, a non-blocking wrapper can be added as a follow-up.
+- No `WorkerGuard` needed — the `LoggingGuard.guards` vector is empty.
+
+### 6. File destination
+
```
banidoru, line 214: `max_files: 0` is documented as "unlimited", but `RollingFileAppender::builder().max_log_files(0)` may actually mean "retain zero rotated files" (i.e., delete them all). Verify the actual `tracing-appender` behavior. If 0 genuinely means unlimited, add a brief note confirming it. If not, use `None` / omission to mean unlimited instead.

specs/009-logging.md
```diff
@@ -0,0 +1,316 @@
...
+- Sending a command
+- Receiving a response
+
+Without `-v`, no tracing subscriber is installed and all `tracing`
+macros are no-ops. Existing `eprintln!` calls for user-facing output
+remain unchanged.
+
+## Interfaces
+
+### Config YAML schema
```
banidoru, line 273: `ReloadHandle` is a generic type (`reload::Handle<L, S>`) where `L` and `S` depend on the full layer stack. The public API signature `handle: &ReloadHandle` won't compile without a type alias or concrete type. Consider defining a type alias like `type ReloadHandle = reload::Handle<Box<dyn Layer<...>>, ...>` or noting that the actual type will be determined by implementation and the spec is illustrative.

banidoru (commented):
Iteration 3 re-review. Prior threads all resolved — good progress. A few remaining issues:

- `#[serde(default)]` still missing on `LoggingConfig` / `LogFileConfig` — previous thread was resolved but the fix isn't in the code
- `max_files: 0` semantics need verification against actual `tracing-appender` API
- `ReloadHandle` generic type unspecified in public API
- Minor: `BootstrapGuard` ownership lifecycle could be clearer

specs/009-logging.md
r12f: Fixed: added  to both  and . Fields with documented defaults now deserialize correctly when omitted from YAML. See commit 47b784b.

specs/009-logging.md
r12f: Fixed: changed  from  to .  (field omitted) means unlimited — avoids the ambiguity of  which could mean 'retain zero files' in . Default remains . See commit 47b784b.

specs/009-logging.md
r12f: Fixed: added a note below the API signature clarifying that  is a type alias, with a concrete definition example: . The exact generic parameters are an implementation detail. See commit 47b784b.

specs/009-logging.md
banidoru: Fixed — `#[serde(default)]` is now on both `LoggingConfig` and `LogFileConfig`, and `max_files` uses `Option<usize>` with `None` for unlimited. Looks correct.

specs/009-logging.md
banidoru: Fixed — `max_files` is now `Option<usize>` where `None` means unlimited. No more ambiguous `0` semantics. The builder call is skipped when `None`. Good fix.

specs/009-logging.md
banidoru: Fixed — added a note clarifying `ReloadHandle` is a type alias, with a concrete example: `type ReloadHandle = reload::Handle<Box<dyn Layer<Registry> + Send + Sync>, Registry>`. Clear enough for implementation.

specs/009-logging.md
```diff
@@ -0,0 +1,330 @@
...
+}
+```
+
+**Config validation:** When `destination` is `File`, the `file` section
+is required. This is enforced during config deserialization/validation
+(not deferred to `init_logging`), producing a clear error message like
+`"logging.file is required when destination is 'file'"`. A missing
+`file` section with `destination: syslog` is fine (`file` is ignored).
+
+**Path resolution:** `LogFileConfig.path` is a `PathBuf`. Relative
```
banidoru, line 156: `#[serde(default)]` on `LogFileConfig` means all fields get `Default` values when omitted — including `path: PathBuf` which defaults to `""`. A user could write:
```yaml
logging:
  destination: file
  file:
    rotation: hourly
```
…and get past the "file section is required" validation, but with an empty path. The config validation paragraph should also check that `path` is non-empty (or make `path` lack a default by not using struct-level `#[serde(default)]` and instead applying it only to `rotation` and `max_files`).

banidoru (commented):
Iteration 4 re-review:

- All 3 prior open threads addressed and resolved (`#[serde(default)]`, `max_files` semantics, `ReloadHandle` type alias)
- One minor new issue: `LogFileConfig` with `#[serde(default)]` + `path: PathBuf` allows empty path when `file:` section is present but `path` is omitted — validation should catch this
- YAML schema shows `max_files: <usize>` but Rust type is `Option<usize>` — minor doc inconsistency
- Overall design is sound — reload layer lifecycle, error contract, directory creation, format spec all look correct
- LGTM with the minor nit above

banidoru (approved):
LGTM

specs/009-logging.md
banidoru, line 156: This is a valid point. The current implementation would allow a file block with no path, which would cause a runtime error. The validation should be tightened to ensure that if the 'file' section exists, the 'path' field is non-empty.

banidoru (changes requested):
- One unresolved issue remains regarding config validation.

specs/009-logging.md
banidoru: Iteration 4: still not addressed in the spec text. The `#[serde(default)]` on `LogFileConfig` still allows `path` to default to `""`. The config-validation paragraph should explicitly state that `path` must be non-empty (and ideally a valid absolute path) when `file` is present, producing a clear error like `"logging.file.path must not be empty"`.

specs/009-logging.md
banidoru, line 156: `max_files: Option<usize>` — the spec says "default: `Some(7)`" but `#[serde(default)]` on the struct will give `Option::None` (the `Default` for `Option`). Either add a custom `Default` impl for `LogFileConfig` that sets `max_files: Some(7)`, or use `#[serde(default = "default_max_files")]` on the field. Without this, omitting `max_files` from YAML silently disables file retention — the opposite of the documented behavior.

banidoru (commented):
Re-review (iteration 4):

- `LogFileConfig` path-validation gap still open — `#[serde(default)]` allows empty `path`, needs explicit validation
- `max_files` default mismatch: spec says default is `Some(7)` but `#[serde(default)]` with `Option<usize>` gives `None` — need a custom `Default` impl or `#[serde(default = "...")]`
- All 12 previously-resolved threads remain correctly addressed
- Overall spec quality is high — clear structure, good error contract, solid lifecycle design

specs/009-logging.md
r12f: Fixed: removed struct-level `#[serde(default)]` from `LogFileConfig`. Now `path` has no default — omitting it from YAML produces a deserialization error. `rotation` and `max_files` use per-field `#[serde(default = "...")]` to retain their documented defaults. Also added explicit validation that `path` must be non-empty (`"logging.file.path must not be empty"`). See commit 3bbf79f.

r12f: Addressed in commit 3bbf79f — removed struct-level `#[serde(default)]`, so `path` is now required by serde. Added validation text specifying the error message for empty/missing path.

r12f: Fixed: `max_files` now uses `#[serde(default = "default_max_files")]` where `fn default_max_files() -> Option<usize> { Some(7) }`. Omitting `max_files` from YAML correctly defaults to `Some(7)` (7-day retention), matching the documented behavior. See commit 3bbf79f.

specs/009-logging.md
r12f: Addressed in commit 3bbf79f — removed struct-level `#[serde(default)]` from `LogFileConfig` so `path` is required by serde (no empty-string default). Added explicit validation text: `"logging.file.path must not be empty"`.

specs/009-logging.md
```diff
@@ -0,0 +1,345 @@
...
+    max_files: <usize>        # optional, default: 7; omit for unlimited
+```
+
+### Public API (`infmon-frontend::logging`)
+
+```rust
+/// Install a bootstrap global subscriber with a reload layer.
+/// The returned guard holds the reload handle for later reconfiguration.
+pub fn init_bootstrap() -> BootstrapGuard;
+
```
banidoru, line 291: Nit: schema says `max_files: <usize>` but the Rust type is `Option<usize>` (where `None` = unlimited). Consider updating to `max_files: <usize | null>` or adding a comment like `# optional, default: 7; null/omit for unlimited` to match the struct definition above.

banidoru (approved):
- All prior review threads addressed and resolved — nice iterative fixes.
- Config validation is now solid: `path` required by serde, per-field defaults on `rotation`/`max_files`, clear error messages.
- Reload-layer lifecycle correctly avoids double `set_global_default()`.
- Minor: YAML schema section says `max_files: <usize>` but the Rust type is `Option<usize>` — consider `max_files: <usize | null>` for consistency.
- LGTM overall. Clean, well-structured spec.

specs/009-logging.md
```diff
@@ -0,0 +1,345 @@
+# 009 — Logging system design (tracing + syslog/file)
+
+## Version history
+
+| Version | Date       | Author       | Changes        |
+| ------- | ---------- | ------------ | -------------- |
+| 0.1     | 2026-04-20 | Riff (r12f)  | Initial draft. |
+| 0.2     | 2026-04-20 | Riff (r12f)  | Address review: add enum defs, PathBuf, error contract, format, directory creation, naming, validation, SIGHUP, max_log_files, RUST_LOG directives, bootstrap threading, backpressure note. |
+| 0.3     | 2026-04-20 | Riff (r12f)  | Address review: use reload::Layer for subscriber swap (set_global_default is once-only), add #[derive(Deserialize)] to config structs. |
+| 0.4     | 2026-04-20 | Riff (r12f)  | Address review: add #[serde(default)] to config structs, change max_files to Option<usize> (None=unlimited), add ReloadHandle type alias note. |
```
banidoru, line 10: Nit: version history shows 0.4 as the latest, but commit `3bbf79f` includes additional changes (removing struct-level `#[serde(default)]` from `LogFileConfig`, adding per-field `#[serde(default = "…")]`). Consider adding a 0.5 entry to keep the version history in sync with the actual spec state.

specs/009-logging.md
banidoru, line 291: Minor: `BootstrapGuard` and `LoggingGuard` are referenced in the public API but their struct definitions aren't shown. At minimum, showing that `BootstrapGuard` has a public `handle: ReloadHandle` field (or an accessor method) would clarify how the caller passes the handle to `init_logging`. A one-liner struct definition for each would make the spec self-contained.

banidoru (approved):
Re-review iteration 5: all 22 prior threads resolved. Spec is solid.

- All prior feedback addressed: enum defs, PathBuf, reload layer, serde defaults, config validation, error contract, format, directory creation, SIGHUP deferral, max_files semantics.
- Two minor nits below (non-blocking).

banidoru (approved):
- The spec is internally consistent and well-defined.
- No new issues found in the latest changes. LGTM.

banidoru (approved):
- All 23 prior review threads addressed and resolved — enum defs, PathBuf, error contract, reload layer, serde defaults, path validation, format, SIGHUP, backpressure, naming — all fixed.
- Spec is self-contained and implementation-ready: config model, lifecycle, API surface, and test plan are well-defined.
- Minor nits remain (version history sync, guard struct definitions, schema type annotation) but these were already flagged and accepted as-is in prior iterations.

banidoru (approved):
All reviewers approved. LGTM.

banidoru (approved):
Re-review (iteration 5): all prior threads resolved. Latest commit correctly removes struct-level `#[serde(default)]` from `LogFileConfig`, adds per-field defaults with explicit functions, and adds path non-empty validation. Spec is solid — no new issues.

- All 24 review threads addressed and resolved
- `LogFileConfig` serde handling is now correct (`path` required, `rotation`/`max_files` have documented defaults)
- Config validation section properly covers both missing `file` block and empty `path`
- No regressions in the rest of the spec

specs/009-logging.md
r12f: Fixed: updated schema to `max_files: <usize | null>` with comment `# optional, default: 7; null/omit for unlimited` to match the `Option<usize>` Rust type. See commit e8cf1df.

specs/009-logging.md
r12f: Fixed: added version 0.5 entry to the version history table covering the commit 3bbf79f changes (removing struct-level `#[serde(default)]` from `LogFileConfig`, per-field defaults, guard struct defs, schema fix). See commit e8cf1df.

specs/009-logging.md
r12f: Fixed: added `BootstrapGuard` and `LoggingGuard` struct definitions in a new 'Guard structs' section under Interfaces. `BootstrapGuard` exposes the `handle: ReloadHandle` field, and `LoggingGuard` holds `Vec<WorkerGuard>`. See commit e8cf1df.

banidoru (approved):
All reviewers approved. LGTM.

banidoru (approved):
- All 22 prior review threads addressed and resolved — clean sweep
- Reload-layer lifecycle is correct: single `set_global_default` + `reload::Handle` for hot-swap
- Serde defaults are properly split: struct-level on `LoggingConfig`, per-field on `LogFileConfig` (path stays required)
- Config validation, directory creation, error contract, and guard ownership are all well-specified
- Spec is self-contained and implementation-ready — LGTM

PR #132, opened 2026-04-20
feat: create E2E test framework scaffolding

### LINKED ISSUE #128: Create E2E test framework: conftest, helpers, replay_traffic
## Description

Create the E2E test framework scaffolding under `tests/e2e/`.

### Components

1. **`tests/e2e/conftest.py`** — session-scoped pytest fixture:
   - Assign RX IP via VPP CLI: `vppctl set interface ip address TwoHundredGigabitEthernet3/0/0 10.123.0.1/24`
   - Assign TX IP via Linux: `ip addr add 10.123.0.2/24 dev p1` (local) or SSH (remote)
   - Verify ping connectivity
   - Start InFMon services if not running
   - Remote mode: SCP `replay_traffic.py` + scenarios to remote host

2. **`tests/e2e/helpers.py`** — InFMon interaction helpers:
   - `flow_rule_add(name, fields, max_keys)` via infmonctl
   - `flow_rule_rm(name)`, `flow_rule_list()`, `get_flow_stats(rule_name)`
   - `clear_all_flow_rules()`, `wait_for_stats(rule_name, timeout)`

3. **`tests/e2e/replay_traffic.py`** — standalone script + importable module:
   - Read pcap with scapy, rewrite dst IP to RX port IP
   - Replay via tcpreplay
   - Callable locally or remotely via SSH

4. **`tests/e2e/requirements.txt`** — `pytest>=7.0`, `scapy>=2.5`
5. **`tests/e2e/__init__.py`**

### Environment Variables

| Variable | Default | Description |
|----------|---------|-------------|
| `INFMON_E2E_TX_MODE` | `local` | `local` or `remote` |
| `INFMON_E2E_TX_IFACE` | `p1` | Linux TX interface (local) |
| `INFMON_E2E_TX_HOST` | — | SSH target (remote mode) |
| `INFMON_E2E_TX_HOST_IFACE` | — | Interface on remote host |
| `INFMON_E2E_RX_VPP_IFACE` | `TwoHundredGigabitEthernet3/0/0` | VPP RX interface |
| `INFMON_E2E_RX_IP` | `10.123.0.1/24` | RX port IP (VPP CLI) |
| `INFMON_E2E_TX_IP` | `10.123.0.2/24` | TX port IP |

### Acceptance Criteria

- [ ] `conftest.py` sets up networking and verifies ping
- [ ] `replay_traffic.py` rewrites dst IP and replays pcap via tcpreplay
- [ ] `helpers.py` wraps infmonctl commands with JSON parsing
- [ ] Remote mode pushes files via SCP and invokes replay via SSH

Blocked by #{dep_issue} (infmonctl flow-rule + stats commands).

r12f:
## Summary

Create the E2E test framework scaffolding under `tests/e2e/` with all components specified in #128:

- **`conftest.py`** — session-scoped pytest fixture that assigns RX IP via VPP CLI, TX IP via Linux (local) or SSH (remote), verifies ping connectivity, starts InFMon if needed, and SCPs replay assets in remote mode
- **`helpers.py`** — infmonctl wrappers: `flow_rule_add`, `flow_rule_rm`, `flow_rule_list`, `get_flow_stats`, `clear_all_flow_rules`, `wait_for_stats`
- **`replay_traffic.py`** — standalone script + importable module that reads pcap with scapy, rewrites dst IP, and replays via tcpreplay (local or remote via SSH)
- **`requirements.txt`** — pytest>=7.0, scapy>=2.5
- **`__init__.py`**

All environment variables from the spec are supported with their documented defaults.

Closes #128

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -6,7 +6,6 @@ connectivity, and optionally pushes replay assets to a remote host.
 
 import os
 import shlex
-import shutil
 import subprocess
 import time
 
@@ -31,7 +30,7 @@ def _env(key: str) -> str:
     return os.environ.get(key, _DEFAULTS.get(key, ""))
 
 
-def _run(cmd: str, check: bool = True, capture: bool = True, **kwargs) -> subprocess.CompletedProcess:
+def _run(cmd: str, check: bool = True, capture: bool = True) -> subprocess.CompletedProcess:
     """Run a shell command.
 
     All caller-supplied values interpolated into *cmd* MUST be wrapped
@@ -69,11 +68,14 @@ def _assign_tx_ip_remote() -> None:
     host = _env("INFMON_E2E_TX_HOST")
     iface = _env("INFMON_E2E_TX_HOST_IFACE")
     ip = _env("INFMON_E2E_TX_IP")
-    _run(
-        f"ssh {shlex.quote(host)} 'ip addr flush dev {shlex.quote(iface)};"
-        f" ip addr replace {shlex.quote(ip)} dev {shlex.quote(iface)};"
-        f" ip link set {shlex.quote(iface)} up'"
+    # Build the remote command separately and quote the whole thing for the
+    # local shell, so shlex.quote() inside doesn't clash with outer quoting.
+    remote_cmd = (
+        f"ip addr flush dev {shlex.quote(iface)}; "
+        f"ip addr replace {shlex.quote(ip)} dev {shlex.quote(iface)}; "
+        f"ip link set {shlex.quote(iface)} up"
     )
+    _run(f"ssh {shlex.quote(host)} {shlex.quote(remote_cmd)}")
 
 
 def _push_replay_assets(remote_host: str) -> None:
@@ -92,7 +94,8 @@ def _verify_ping() -> None:
     mode = _env("INFMON_E2E_TX_MODE")
     if mode == "remote":
         host = _env("INFMON_E2E_TX_HOST")
-        cmd = f"ssh {shlex.quote(host)} 'ping -c 3 -W 2 {shlex.quote(rx_ip)}'"
+        remote_cmd = f"ping -c 3 -W 2 {shlex.quote(rx_ip)}"
+        cmd = f"ssh {shlex.quote(host)} {shlex.quote(remote_cmd)}"
     else:
         cmd = f"ping -c 3 -W 2 {shlex.quote(rx_ip)}"
     result = _run(cmd, check=False)
@@ -172,10 +175,12 @@ def infmon_env():
     if mode == "remote":
         remote_host = _env("INFMON_E2E_TX_HOST")
         remote_iface = _env("INFMON_E2E_TX_HOST_IFACE")
+        remote_cmd = (
+            f"ip addr flush dev {shlex.quote(remote_iface)}; "
+            f"rm -f /tmp/replay_traffic.py; "
+            f"rm -rf /tmp/e2e_scenarios"
+        )
         _run(
-            f"ssh {shlex.quote(remote_host)} '"
-            f"ip addr flush dev {shlex.quote(remote_iface)};"
-            f" rm -f /tmp/replay_traffic.py;"
-            f" rm -rf /tmp/e2e_scenarios'",
+            f"ssh {shlex.quote(remote_host)} {shlex.quote(remote_cmd)}",
             check=False,
         )
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -5,6 +5,7 @@ connectivity, and optionally pushes replay assets to a remote host.
 """
 
 import os
+import shlex
 import shutil
 import subprocess
 import time
@@ -30,7 +31,12 @@ def _env(key: str) -> str:
     return os.environ.get(key, _DEFAULTS.get(key, ""))
 
 
-def _run(cmd: str, check: bool = True, capture: bool = True) -> subprocess.CompletedProcess:
+def _run(cmd: str, check: bool = True, capture: bool = True, **kwargs) -> subprocess.CompletedProcess:
+    """Run a shell command.
+
+    All caller-supplied values interpolated into *cmd* MUST be wrapped
+    with ``shlex.quote()`` to prevent shell injection.
+    """
     return subprocess.run(
         cmd, shell=True, check=check, capture_output=capture, text=True
     )
@@ -44,8 +50,8 @@ def _assign_rx_ip() -> None:
     """Assign IP to the VPP RX interface via vppctl."""
     iface = _env("INFMON_E2E_RX_VPP_IFACE")
     ip = _env("INFMON_E2E_RX_IP")
-    _run(f"vppctl set interface ip address {iface} {ip}", check=False)
-    _run(f"vppctl set interface state {iface} up", check=False)
+    _run(f"vppctl set interface ip address {shlex.quote(iface)} {shlex.quote(ip)}", check=False)
+    _run(f"vppctl set interface state {shlex.quote(iface)} up", check=False)
 
 
 def _assign_tx_ip_local() -> None:
@@ -53,9 +59,9 @@ def _assign_tx_ip_local() -> None:
     iface = _env("INFMON_E2E_TX_IFACE")
     ip = _env("INFMON_E2E_TX_IP")
     # Flush existing addresses first to avoid duplicates
-    _run(f"ip addr flush dev {iface}", check=False)
-    _run(f"ip addr add {ip} dev {iface}")
-    _run(f"ip link set {iface} up")
+    _run(f"ip addr flush dev {shlex.quote(iface)}", check=False)
+    _run(f"ip addr replace {shlex.quote(ip)} dev {shlex.quote(iface)}")
+    _run(f"ip link set {shlex.quote(iface)} up")
 
 
 def _assign_tx_ip_remote() -> None:
@@ -63,7 +69,11 @@ def _assign_tx_ip_remote() -> None:
     host = _env("INFMON_E2E_TX_HOST")
     iface = _env("INFMON_E2E_TX_HOST_IFACE")
     ip = _env("INFMON_E2E_TX_IP")
-    _run(f"ssh {host} 'ip addr flush dev {iface}; ip addr add {ip} dev {iface}; ip link set {iface} up'")
+    _run(
+        f"ssh {shlex.quote(host)} 'ip addr flush dev {shlex.quote(iface)};"
+        f" ip addr replace {shlex.quote(ip)} dev {shlex.quote(iface)};"
+        f" ip link set {shlex.quote(iface)} up'"
+    )
 
 
 def _push_replay_assets(remote_host: str) -> None:
@@ -71,9 +81,9 @@ def _push_replay_assets(remote_host: str) -> None:
     e2e_dir = os.path.dirname(__file__)
     replay_script = os.path.join(e2e_dir, "replay_traffic.py")
     scenarios_dir = os.path.join(e2e_dir, "scenarios")
-    _run(f"scp {replay_script} {remote_host}:/tmp/replay_traffic.py")
+    _run(f"scp {shlex.quote(replay_script)} {shlex.quote(remote_host)}:/tmp/replay_traffic.py")
     if os.path.isdir(scenarios_dir):
-        _run(f"scp -r {scenarios_dir} {remote_host}:/tmp/e2e_scenarios")
+        _run(f"scp -r {shlex.quote(scenarios_dir)} {shlex.quote(remote_host)}:/tmp/e2e_scenarios")
 
 
 def _verify_ping() -> None:
@@ -82,9 +92,9 @@ def _verify_ping() -> None:
     mode = _env("INFMON_E2E_TX_MODE")
     if mode == "remote":
         host = _env("INFMON_E2E_TX_HOST")
-        cmd = f"ssh {host} 'ping -c 3 -W 2 {rx_ip}'"
+        cmd = f"ssh {shlex.quote(host)} 'ping -c 3 -W 2 {shlex.quote(rx_ip)}'"
     else:
-        cmd = f"ping -c 3 -W 2 {rx_ip}"
+        cmd = f"ping -c 3 -W 2 {shlex.quote(rx_ip)}"
     result = _run(cmd, check=False)
     if result.returncode != 0:
         pytest.fail(
@@ -136,3 +146,19 @@ def infmon_env():
         "rx_ip": _env("INFMON_E2E_RX_IP"),
         "tx_ip": _env("INFMON_E2E_TX_IP"),
     }
+
+    # ---- Teardown ----
+    # Flush IPs assigned during setup to avoid stale addresses on next run.
+    tx_iface = _env("INFMON_E2E_TX_IFACE")
+    _run(f"ip addr flush dev {shlex.quote(tx_iface)}", check=False)
+
+    if mode == "remote":
+        remote_host = _env("INFMON_E2E_TX_HOST")
+        remote_iface = _env("INFMON_E2E_TX_HOST_IFACE")
+        _run(
+            f"ssh {shlex.quote(remote_host)} '"
+            f"ip addr flush dev {shlex.quote(remote_iface)};"
+            f" rm -f /tmp/replay_traffic.py;"
+            f" rm -rf /tmp/e2e_scenarios'",
+            check=False,
+        )
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -1,0 +1,138 @@
+"""Session-scoped fixtures for InFMon E2E tests.
+
+Sets up networking (VPP RX port, Linux/remote TX port), verifies
+connectivity, and optionally pushes replay assets to a remote host.
+"""
+
+import os
+import shutil
+import subprocess
+import time
+
+import pytest
+
+# ---------------------------------------------------------------------------
+# Environment defaults
+# ---------------------------------------------------------------------------
+
+_DEFAULTS = {
+    "INFMON_E2E_TX_MODE": "local",
+    "INFMON_E2E_TX_IFACE": "p1",
+    "INFMON_E2E_TX_HOST": "",
+    "INFMON_E2E_TX_HOST_IFACE": "",
+    "INFMON_E2E_RX_VPP_IFACE": "TwoHundredGigabitEthernet3/0/0",
+    "INFMON_E2E_RX_IP": "10.123.0.1/24",
+    "INFMON_E2E_TX_IP": "10.123.0.2/24",
+}
+
+
+def _env(key: str) -> str:
+    return os.environ.get(key, _DEFAULTS.get(key, ""))
+
+
+def _run(cmd: str, check: bool = True, capture: bool = True) -> subprocess.CompletedProcess:
+    return subprocess.run(
+        cmd, shell=True, check=check, capture_output=capture, text=True
+    )
+
+
+# ---------------------------------------------------------------------------
+# Networking helpers
+# ---------------------------------------------------------------------------
+
+def _assign_rx_ip() -> None:
+    """Assign IP to the VPP RX interface via vppctl."""
+    iface = _env("INFMON_E2E_RX_VPP_IFACE")
+    ip = _env("INFMON_E2E_RX_IP")
+    _run(f"vppctl set interface ip address {iface} {ip}", check=False)
+    _run(f"vppctl set interface state {iface} up", check=False)
+
+
+def _assign_tx_ip_local() -> None:
+    """Assign IP to the local Linux TX interface."""
+    iface = _env("INFMON_E2E_TX_IFACE")
+    ip = _env("INFMON_E2E_TX_IP")
+    # Flush existing addresses first to avoid duplicates
+    _run(f"ip addr flush dev {iface}", check=False)
+    _run(f"ip addr add {ip} dev {iface}")
+    _run(f"ip link set {iface} up")
+
+
+def _assign_tx_ip_remote() -> None:
+    """Assign IP on the remote TX host via SSH."""
+    host = _env("INFMON_E2E_TX_HOST")
+    iface = _env("INFMON_E2E_TX_HOST_IFACE")
+    ip = _env("INFMON_E2E_TX_IP")
+    _run(f"ssh {host} 'ip addr flush dev {iface}; ip addr add {ip} dev {iface}; ip link set {iface} up'")
+
+
+def _push_replay_assets(remote_host: str) -> None:
+    """SCP replay_traffic.py and scenario directories to the remote host."""
+    e2e_dir = os.path.dirname(__file__)
+    replay_script = os.path.join(e2e_dir, "replay_traffic.py")
+    scenarios_dir = os.path.join(e2e_dir, "scenarios")
+    _run(f"scp {replay_script} {remote_host}:/tmp/replay_traffic.py")
+    if os.path.isdir(scenarios_dir):
+        _run(f"scp -r {scenarios_dir} {remote_host}:/tmp/e2e_scenarios")
+
+
+def _verify_ping() -> None:
+    """Ping the RX IP from the TX side to verify connectivity."""
+    rx_ip = _env("INFMON_E2E_RX_IP").split("/")[0]
+    mode = _env("INFMON_E2E_TX_MODE")
+    if mode == "remote":
+        host = _env("INFMON_E2E_TX_HOST")
+        cmd = f"ssh {host} 'ping -c 3 -W 2 {rx_ip}'"
+    else:
+        cmd = f"ping -c 3 -W 2 {rx_ip}"
+    result = _run(cmd, check=False)
+    if result.returncode != 0:
+        pytest.fail(
+            f"Ping to RX IP {rx_ip} failed — check physical connectivity and IP config."
+        )
+
+
+def _ensure_infmon_running() -> None:
+    """Start InFMon services if they are not already running."""
+    result = _run("systemctl is-active infmon", check=False)
+    if result.stdout.strip() != "active":
+        _run("systemctl start infmon", check=False)
+        time.sleep(2)
+
+
+# ---------------------------------------------------------------------------
+# Session fixture
+# ---------------------------------------------------------------------------
+
+@pytest.fixture(scope="session")
+def infmon_env():
+    """Set up networking, verify connectivity, and yield config dict."""
+    mode = _env("INFMON_E2E_TX_MODE")
+
+    # 1. Assign IPs
+    _assign_rx_ip()
+    if mode == "remote":
+        host = _env("INFMON_E2E_TX_HOST")
+        if not host:
+            pytest.fail("INFMON_E2E_TX_HOST must be set in remote mode")
+        _assign_tx_ip_remote()
+        _push_replay_assets(host)
+    else:
+        _assign_tx_ip_local()
+
+    # 2. Verify connectivity
+    _verify_ping()
+
+    # 3. Ensure InFMon is running
+    _ensure_infmon_running()
+
+    # 4. Yield config for tests
+    yield {
+        "tx_mode": mode,
+        "tx_iface": _env("INFMON_E2E_TX_IFACE"),
+        "tx_host": _env("INFMON_E2E_TX_HOST"),
+        "tx_host_iface": _env("INFMON_E2E_TX_HOST_IFACE"),
+        "rx_vpp_iface": _env("INFMON_E2E_RX_VPP_IFACE"),
+        "rx_ip": _env("INFMON_E2E_RX_IP"),
+        "tx_ip": _env("INFMON_E2E_TX_IP"),
+    }
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -115,6 +115,8 @@ def _ensure_infmon_running() -> None:
             time.sleep(0.5)
             if _run("systemctl is-active infmon", check=False).stdout.strip() == "active":
                 break
+        else:
+            pytest.fail("InFMon did not become active within 10s")
 
 
 # ---------------------------------------------------------------------------
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -134,6 +134,7 @@ def infmon_env():
     _verify_ping()
 
     # 3. Ensure InFMon is running
+    service_was_inactive = _run("systemctl is-active infmon", check=False).stdout.strip() != "active"
     _ensure_infmon_running()
 
     # 4. Yield config for tests
@@ -148,6 +149,10 @@ def infmon_env():
     }
 
     # ---- Teardown ----
+    # Stop InFMon if we started it during setup.
+    if service_was_inactive:
+        _run("systemctl stop infmon", check=False)
+
     # Flush IPs assigned during setup to avoid stale addresses on next run.
     tx_iface = _env("INFMON_E2E_TX_IFACE")
     _run(f"ip addr flush dev {shlex.quote(tx_iface)}", check=False)
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -107,7 +107,11 @@ def _ensure_infmon_running() -> None:
     result = _run("systemctl is-active infmon", check=False)
     if result.stdout.strip() != "active":
         _run("systemctl start infmon", check=False)
-        time.sleep(2)
+        # Poll until active (up to 10s) instead of a fixed sleep
+        for _ in range(20):
+            time.sleep(0.5)
+            if _run("systemctl is-active infmon", check=False).stdout.strip() == "active":
+                break
 
 
 # ---------------------------------------------------------------------------
@@ -118,6 +122,9 @@ def _ensure_infmon_running() -> None:
 def infmon_env():
     """Set up networking, verify connectivity, and yield config dict."""
     mode = _env("INFMON_E2E_TX_MODE")
+
+    if mode not in ("local", "remote"):
+        pytest.fail(f"Invalid INFMON_E2E_TX_MODE: {mode!r}. Must be 'local' or 'remote'.")
 
     # 1. Assign IPs
     _assign_rx_ip()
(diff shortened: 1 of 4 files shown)


tests/e2e/conftest.py
```diff
@@ -0,0 +1,138 @@
...
+
+def _env(key: str) -> str:
+    return os.environ.get(key, _DEFAULTS.get(key, ""))
+
+
+def _run(cmd: str, check: bool = True, capture: bool = True) -> subprocess.CompletedProcess:
+    return subprocess.run(
+        cmd, shell=True, check=check, capture_output=capture, text=True
+    )
+
```
banidoru, line 42: **Shell injection risk.** `_run()` uses `shell=True` and every caller interpolates env-var values directly into the command string. A malicious or malformed env var (e.g. `INFMON_E2E_TX_IFACE="; rm -rf /"`) would be executed.

Either switch to list-form `subprocess.run(..., shell=False)` everywhere, or at minimum wrap all interpolated values with `shlex.quote()`.

tests/e2e/conftest.py
```diff
@@ -0,0 +1,138 @@
...
+        _assign_tx_ip_local()
+
+    # 2. Verify connectivity
+    _verify_ping()
+
+    # 3. Ensure InFMon is running
+    _ensure_infmon_running()
+
+    # 4. Yield config for tests
+    yield {
```
banidoru, line 153: **No teardown.** The fixture assigns IPs (VPP + Linux/remote), starts InFMon, and pushes assets to a remote host, but there is no `yield` → cleanup block. On the next test session, `ip addr add` may fail or produce duplicate addresses, and `/tmp` on the remote host accumulates replay assets.

Add a teardown after `yield` that at minimum:
- Flushes the assigned IPs
- Removes remote `/tmp/replay_traffic.py` and `/tmp/e2e_scenarios` in remote mode

tests/e2e/conftest.py
```diff
@@ -0,0 +1,138 @@
...
+    iface = _env("INFMON_E2E_TX_IFACE")
+    ip = _env("INFMON_E2E_TX_IP")
+    # Flush existing addresses first to avoid duplicates
+    _run(f"ip addr flush dev {iface}", check=False)
+    _run(f"ip addr add {ip} dev {iface}")
+    _run(f"ip link set {iface} up")
+
+
+def _assign_tx_ip_remote() -> None:
+    """Assign IP on the remote TX host via SSH."""
```
banidoru, line 67: **Inconsistent error handling.** `ip addr flush` uses `check=False` (fine), but `ip addr add` and `ip link set` use the default `check=True`. If the address already exists (e.g. leftover from a previous aborted run), `ip addr add` raises `CalledProcessError` and the entire session fixture fails.

Consider `check=False` + explicit error logging, or use `ip addr replace` instead of `add`.

tests/e2e/replay_traffic.py
```diff
@@ -0,0 +1,100 @@
...
+            pkt[IP].dst = dst_ip
+            # Recalculate checksums
+            del pkt[IP].chksum
+            if hasattr(pkt[IP], "payload") and hasattr(pkt[IP].payload, "chksum"):
+                del pkt[IP].payload.chksum
+        elif pkt.haslayer(IPv6):
+            pkt[IPv6].dst = dst_ip
+        rewritten.append(pkt)
+    return rewritten
+
```
banidoru, line 42: **Fragile checksum deletion.** This only handles one specific nesting (`IP.payload.chksum`) which catches TCP but misses UDP, ICMP, or deeper stacks. Also, `hasattr(pkt[IP].payload, 'chksum')` is `True` for `Raw` payloads in some scapy versions.

Idiomatic scapy approach: after rewriting, call `del pkt[IP].chksum` and then `pkt = pkt.__class__(bytes(pkt))` to force full recalculation of all checksums.

tests/e2e/replay_traffic.py
```diff
@@ -0,0 +1,100 @@
...
+        count: Number of times to replay (default: 1).
+        remote_host: If set, replay on this host via SSH.
+    """
+    packets = rdpcap(pcap_path)
+    rewritten = rewrite_dst_ip(packets, dst_ip)
+
+    with tempfile.NamedTemporaryFile(suffix=".pcap", delete=False) as tmp:
+        tmp_path = tmp.name
+        wrpcap(tmp_path, rewritten)
+
```
banidoru, line 62: **Temp-file leak.** If `rdpcap` or `rewrite_dst_ip` or `wrpcap` raises, the temp file is never deleted because the `try/finally` block starts after `wrpcap`. Move the `try` to wrap the `wrpcap` call as well, or use the context manager properly with `delete=True` and only keep the file alive for the subprocess.

tests/e2e/replay_traffic.py
```diff
@@ -0,0 +1,100 @@
...
+    try:
+        loop_arg = f"--loop={count}" if count > 1 else ""
+        if remote_host:
+            # Copy rewritten pcap to remote and replay there
+            remote_tmp = f"/tmp/replay_{os.getpid()}.pcap"
+            subprocess.run(
+                f"scp {tmp_path} {remote_host}:{remote_tmp}",
+                shell=True, check=True, capture_output=True,
```
banidoru, line 74: **Shell injection.** `iface`, `remote_host`, `tmp_path`, etc. are interpolated into shell commands via f-strings with `shell=True`. Same concern as `conftest.py` — use list-form subprocess or `shlex.quote()`.

Also, when `count == 1`, `loop_arg` is `""` which leaves a harmless double-space but looks unintentional — consider omitting it entirely or always passing `--loop=1`.

tests/e2e/helpers.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Args:
+        rule_name: The flow rule name.
+        timeout: Max seconds to wait.
+        poll_interval: Seconds between polls.
+
+    Returns:
+        Flow stats dict.
+
+    Raises:
```
banidoru, line 88: **False-positive early return.** `get_flow_stats` returns `_parse_json_output(result)` which can be `{}` or `[]` — both are truthy for `is not None`. If the stats endpoint returns an empty object before data is populated, `wait_for_stats` will exit immediately.

Consider a more meaningful readiness check, e.g. verifying a specific key exists or that packet/byte counts are > 0.

tests/e2e/helpers.py
```diff
@@ -0,0 +1,91 @@
...
+    """Get flow statistics for a specific rule, returning parsed JSON."""
+    result = _run_infmonctl("stats", "--name", rule_name, "--json", check=False)
+    if result.returncode != 0:
+        return None
+    return _parse_json_output(result)
+
+
+def clear_all_flow_rules() -> None:
+    """Remove all flow rules."""
+    rules = flow_rule_list()
```
banidoru, line 67: **No error resilience in `clear_all_flow_rules`.** If `flow_rule_rm` raises for one rule, the remaining rules are never deleted. Consider catching `subprocess.CalledProcessError` per-rule and logging failures, or using a `--all` flag if `infmonctl` supports it.

tests/e2e/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pytest>=7.0
+scapy>=2.5
```
banidoru, line 2: Consider pinning `tcpreplay` as a system dependency somewhere (README or a `Makefile` target), since the code depends on it at runtime but it's not a pip package.

banidoru (changes requested):
- **Shell injection risk**: Multiple functions build shell commands via f-string interpolation of env vars / user params with `shell=True`. Use list-form `subprocess.run` or `shlex.quote`.
- **No teardown**: Session fixture assigns IPs and pushes remote assets but never cleans up — stale config leaks across runs.
- **Checksum recalculation is fragile**: `rewrite_dst_ip` only handles one layer of L4 checksum via `hasattr` on `.payload.chksum` — misses nested/UDP cases. Use scapy idioms instead.
- **Temp-file leak**: `wrpcap` can throw before the `try/finally`, leaving the temp file behind.
- **`wait_for_stats` false-positive**: Returns on any non-`None` value including empty dict/list — may exit early before real data arrives.

tests/e2e/conftest.py
```diff
@@ -0,0 +1,138 @@
...
+        time.sleep(2)
+
+
+# ---------------------------------------------------------------------------
+# Session fixture
+# ---------------------------------------------------------------------------
+
+@pytest.fixture(scope="session")
+def infmon_env():
+    """Set up networking, verify connectivity, and yield config dict."""
```
banidoru, line 128: **Service Lifecycle:** The fixture ensures `infmon` is running but never stops it during teardown. If the test session starts the service, it should be responsible for stopping it afterward to ensure a clean state for subsequent, unrelated activities on the test machine.

tests/e2e/replay_traffic.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def rewrite_dst_ip(packets, dst_ip: str):
+    """Rewrite destination IP in all packets.
+
+    Handles both IPv4 and IPv6 based on the dst_ip format.
+    Returns a new list of modified packets.
+    """
+    rewritten = []
+    for pkt in packets:
```
banidoru, line 30: **Missing VLAN support:** The code checks `pkt.haslayer(IP)` directly. If a packet has a VLAN tag (802.1Q), this check will fail, and the destination IP will not be rewritten. The logic should be updated to handle packets with VLAN encapsulation by checking for the IP layer within the VLAN layer.

tests/e2e/helpers.py
```diff
@@ -0,0 +1,91 @@
...
+def flow_rule_add(name: str, fields: Dict[str, str], max_keys: int = 0) -> None:
+    """Add a flow rule via infmonctl.
+
+    Args:
+        name: Rule name.
+        fields: Dict of match field name to value.
+        max_keys: Maximum number of flow keys (0 = unlimited).
+    """
+    args = ["flow-rule", "add", "--name", name]
+    for field_name, field_value in fields.items():
```
banidoru, line 38: **Non-deterministic argument order:** The `fields` parameter is a `dict`. In Python versions before 3.7, dictionary iteration order is not guaranteed. This means the order of `--field` arguments passed to `infmonctl` can change between runs. While it probably doesnt matter for this specific tool, its a good practice in testing to ensure command-line arguments are generated in a deterministic order. Using `collections.OrderedDict` or a list of tuples `(field_name, field_value)` would be more robust.

banidoru (changes requested):
This PR is a great start for the E2E test framework, but has several critical security and correctness issues that need to be addressed. The existing comments from other reviewers have already highlighted the most severe problems (shell injection, resource leaks). My additional comments focus on improving robustness and handling edge cases.

- **Resource Management:** The `conftest.py` fixture starts `infmon` but never stops it. Tests should clean up after themselves.
- **VLAN Handling:** The traffic replay script does not handle VLAN-tagged packets, which could cause tests to fail with certain pcaps.
- **Deterministic Behavior:** The use of standard dicts for command-line arguments can lead to non-deterministic behavior in tests.

tests/e2e/conftest.py
r12f: Fixed: added `shlex.quote()` to all shell-interpolated values across `conftest.py` and `replay_traffic.py`. Also added a docstring to `_run()` documenting the quoting requirement.

tests/e2e/conftest.py
r12f: Fixed: added a teardown block after `yield` in the `infmon_env` fixture that flushes assigned IPs on the local TX interface, and in remote mode also flushes the remote interface IP and removes `/tmp/replay_traffic.py` and `/tmp/e2e_scenarios`.

tests/e2e/conftest.py
r12f: Fixed: switched from `ip addr add` to `ip addr replace` which is idempotent — it adds the address if missing or updates it if already present. Applied to both local and remote paths. This avoids the `CalledProcessError` on re-runs.

tests/e2e/replay_traffic.py
r12f: Fixed: replaced the fragile single-layer checksum deletion with the idiomatic scapy pattern: `del pkt[IP].chksum` followed by `pkt = pkt.__class__(bytes(pkt))` which forces full recalculation of checksums at all layers (TCP, UDP, ICMP, etc.).

tests/e2e/replay_traffic.py
r12f: Fixed: moved `rdpcap`, `rewrite_dst_ip`, and `wrpcap` inside the `try/finally` block so the temp file is always cleaned up even if those calls raise. Also added `os.path.exists()` guard on the unlink.

tests/e2e/replay_traffic.py
r12f: Fixed: added `shlex.quote()` to all interpolated values (`iface`, `remote_host`, `tmp_path`, `remote_tmp`) in subprocess calls. The empty-string `loop_arg` when `count == 1` is harmless (shell collapses double spaces) but could be tightened in a follow-up.

tests/e2e/helpers.py
r12f: Fixed: `wait_for_stats` now checks `if stats is not None and stats:` so empty dicts/lists (which are falsy) won't cause an early return. The loop will keep polling until meaningful data appears.

tests/e2e/helpers.py
r12f: Fixed: wrapped `flow_rule_rm(name)` in a try/except for `CalledProcessError` with a warning log, so one failing rule doesn't prevent the rest from being cleaned up.

tests/e2e/requirements.txt
r12f: Good point. Added a comment in `requirements.txt` noting that `tcpreplay` is a system dependency (`apt install tcpreplay`). We can add it to a Makefile target or README in a follow-up.

tests/e2e/conftest.py
```diff
@@ -0,0 +1,164 @@
...
+    # 1. Assign IPs
+    _assign_rx_ip()
+    if mode == "remote":
+        host = _env("INFMON_E2E_TX_HOST")
+        if not host:
+            pytest.fail("INFMON_E2E_TX_HOST must be set in remote mode")
+        _assign_tx_ip_remote()
+        _push_replay_assets(host)
+    else:
+        _assign_tx_ip_local()
```
banidoru, line 143: **No TX mode validation.** If `INFMON_E2E_TX_MODE` is set to an invalid value (e.g. a typo like `"remot"`), the code silently falls through to the local path. Add an explicit check:
```python
if mode not in ("local", "remote"):
    pytest.fail(f"Invalid INFMON_E2E_TX_MODE: {mode!r}. Must be 'local' or 'remote'.")
```

tests/e2e/replay_traffic.py
```diff
@@ -0,0 +1,103 @@
...
+            del pkt[IP].chksum
+            # Force full checksum recalculation for all layers
+            pkt = pkt.__class__(bytes(pkt))
+        elif pkt.haslayer(IPv6):
+            pkt[IPv6].dst = dst_ip
+        rewritten.append(pkt)
+    return rewritten
+
+
+def replay(
```
banidoru, line 44: **IPv6 checksum not recalculated.** The IPv4 branch correctly deletes the checksum and rebuilds the packet, but the IPv6 branch just sets `dst` without rebuilding. While IPv6 itself has no header checksum, upper-layer protocols (TCP/UDP) use a pseudo-header that includes the destination address — their checksums will be stale after this rewrite.

Apply the same rebuild pattern:
```python
elif pkt.haslayer(IPv6):
    pkt[IPv6].dst = dst_ip
    pkt = pkt.__class__(bytes(pkt))
```

tests/e2e/conftest.py
```diff
@@ -0,0 +1,164 @@
...
+        "rx_ip": _env("INFMON_E2E_RX_IP"),
+        "tx_ip": _env("INFMON_E2E_TX_IP"),
+    }
+
+    # ---- Teardown ----
+    # Flush IPs assigned during setup to avoid stale addresses on next run.
+    tx_iface = _env("INFMON_E2E_TX_IFACE")
+    _run(f"ip addr flush dev {shlex.quote(tx_iface)}", check=False)
+
+    if mode == "remote":
```
banidoru, line 177: **VPP RX IP not cleaned up.** The teardown flushes the TX-side IP but never removes the IP assigned to the VPP RX interface. Add:
```python
iface = _env("INFMON_E2E_RX_VPP_IFACE")
ip = _env("INFMON_E2E_RX_IP")
_run(f"vppctl set interface ip address {shlex.quote(iface)} del {shlex.quote(ip)}", check=False)
```

tests/e2e/conftest.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Session fixture
+# ---------------------------------------------------------------------------
+
+@pytest.fixture(scope="session")
+def infmon_env():
+    """Set up networking, verify connectivity, and yield config dict."""
+    mode = _env("INFMON_E2E_TX_MODE")
```
banidoru, line 129: **Fragile fixed sleep.** `time.sleep(2)` is a race condition waiting to happen — on a slow machine InFMon may not be ready, and on a fast machine it wastes time. Consider polling `systemctl is-active infmon` in a retry loop with a timeout instead.

tests/e2e/helpers.py
```diff
@@ -0,0 +1,97 @@
...
+Wraps infmonctl commands for flow rule management and stats retrieval.
+"""
+
+import json
+import logging
+import subprocess
+import time
+
+logger = logging.getLogger(__name__)
+from typing import Any, Dict, List, Optional
```
banidoru: **PEP 8 import ordering.** The `from typing import ...` statement is placed after `logger = logging.getLogger(__name__)`. Move it up with the other imports — all imports should come before module-level statements.

banidoru (commented):
- **TX mode validation gap**: Invalid `INFMON_E2E_TX_MODE` values silently fall through to local mode — should fail explicitly.
- **IPv6 checksum not recalculated**: After rewriting IPv6 dst, no checksum rebuild is performed unlike IPv4.
- **VPP RX IP never cleaned up in teardown**: Only TX-side IPs are flushed; the VPP interface retains its address.
- **Hardcoded `sleep(2)`**: `_ensure_infmon_running` should poll for readiness instead of a fixed sleep.
- **Import ordering**: `helpers.py` has a statement between import groups, violating PEP 8.

tests/e2e/conftest.py
r12f: Fixed: the teardown block now stops the `infmon` service if the fixture was the one that started it (tracked via a `service_was_inactive` flag captured before `_ensure_infmon_running()`). If the service was already running before the test session, it's left alone.

tests/e2e/replay_traffic.py
r12f: Scapy's `haslayer(IP)` already traverses through encapsulation layers including 802.1Q/VLAN tags — it walks the full packet layer stack, not just the immediate payload. So `pkt.haslayer(IP)` will return `True` and `pkt[IP]` will correctly access the IP layer even when VLAN-tagged. No change needed here.

tests/e2e/helpers.py
r12f: Python 3.7+ guarantees dict insertion order (and CPython 3.6+ as an implementation detail). Since this project targets modern Python, dict iteration order is deterministic. The `--field` arguments will always be emitted in the order they were inserted into the `fields` dict. No change needed.

tests/e2e/conftest.py
r12f: Fixed: added an explicit validation check at the top of `infmon_env` that fails immediately if `INFMON_E2E_TX_MODE` is not `'local'` or `'remote'`.

tests/e2e/replay_traffic.py
r12f: Good catch. Fixed: the IPv6 branch now also rebuilds the packet with `pkt = pkt.__class__(bytes(pkt))` after setting the dst, which forces recalculation of upper-layer (TCP/UDP) checksums that depend on the pseudo-header.

tests/e2e/conftest.py
r12f: Fixed: added `vppctl set interface ip address <iface> del <ip>` to the teardown block to remove the VPP RX IP.

tests/e2e/conftest.py
r12f: Fixed: replaced the `time.sleep(2)` with a polling loop that checks `systemctl is-active infmon` every 0.5s for up to 10s.

tests/e2e/helpers.py
r12f: Fixed: moved `from typing import ...` above `logger = logging.getLogger(__name__)` to follow PEP 8 import ordering.

tests/e2e/conftest.py
```diff
@@ -0,0 +1,181 @@
+"""Session-scoped fixtures for InFMon E2E tests.
+
+Sets up networking (VPP RX port, Linux/remote TX port), verifies
+connectivity, and optionally pushes replay assets to a remote host.
+"""
+
+import os
+import shlex
+import shutil
```
banidoru: **Unused import.** `shutil` is imported but never used anywhere in the file. Remove it to keep imports clean.

tests/e2e/conftest.py
```diff
@@ -0,0 +1,181 @@
...
+}
+
+
+def _env(key: str) -> str:
+    return os.environ.get(key, _DEFAULTS.get(key, ""))
+
+
+def _run(cmd: str, check: bool = True, capture: bool = True, **kwargs) -> subprocess.CompletedProcess:
+    """Run a shell command.
+
```
banidoru, line 35: **Unused `**kwargs`.** The `_run` signature accepts `**kwargs` but never passes them to `subprocess.run`. Either remove `**kwargs` or forward it:
```python
def _run(cmd: str, check: bool = True, capture: bool = True) -> subprocess.CompletedProcess:
```

tests/e2e/conftest.py
```diff
@@ -0,0 +1,181 @@
...
+        f"ssh {shlex.quote(host)} 'ip addr flush dev {shlex.quote(iface)};"
+        f" ip addr replace {shlex.quote(ip)} dev {shlex.quote(iface)};"
+        f" ip link set {shlex.quote(iface)} up'"
+    )
+
+
+def _push_replay_assets(remote_host: str) -> None:
+    """SCP replay_traffic.py and scenario directories to the remote host."""
+    e2e_dir = os.path.dirname(__file__)
+    replay_script = os.path.join(e2e_dir, "replay_traffic.py")
```
banidoru, line 84: **Double-quoting bug in SSH commands.** `shlex.quote()` is applied to values *inside* a single-quoted SSH command string. If any value contains special characters, `shlex.quote` wraps it in single quotes, which will prematurely terminate the outer single quotes and break the command.

For example, if `iface` were `eth 0`, `shlex.quote` produces `'eth 0'`, and the full string becomes:
```
ssh host 'ip addr flush dev 'eth 0'; ...'
```
which the local shell splits incorrectly.

**Fix:** Either use a list-form SSH command, or double-escape for the remote shell. A simpler approach:
```python
# Build the remote command separately, quote for remote shell
remote_cmd = (
    f"ip addr flush dev {shlex.quote(iface)}; "
    f"ip addr replace {shlex.quote(ip)} dev {shlex.quote(iface)}; "
    f"ip link set {shlex.quote(iface)} up"
)
_run(f"ssh {shlex.quote(host)} {shlex.quote(remote_cmd)}")
```
This applies `shlex.quote` at each nesting level correctly.

Same issue exists in `_verify_ping` (line 101) and the teardown SSH block (line 172).

banidoru (commented):
Iteration 2 re-review. All 15 prior threads addressed and resolved. A few minor issues remain in the new code:

banidoru (approved):
Iteration 2 re-review — all prior concerns addressed cleanly.

- ✅ `shlex.quote()` applied consistently across `conftest.py` and `replay_traffic.py`
- ✅ Teardown block added: flushes IPs, removes VPP RX IP, cleans remote assets, stops infmon if fixture started it
- ✅ `ip addr replace` for idempotent IP assignment
- ✅ Scapy checksum rebuild via `pkt.__class__(bytes(pkt))` for both IPv4 and IPv6
- ✅ Temp-file lifecycle fixed, `clear_all_flow_rules` error-resilient, `wait_for_stats` falsy check, PEP 8 import ordering fixed

R1 has 3 remaining open threads (unused `shutil`, unused `**kwargs`, SSH double-quoting inside single quotes) — all valid, recommend addressing before merge.

tests/e2e/conftest.py
r12f: Fixed: removed the unused `shutil` import in commit a686b7b.

tests/e2e/conftest.py
r12f: Fixed: removed the unused `**kwargs` from `_run()` signature in commit a686b7b.

tests/e2e/conftest.py
r12f: Fixed: refactored all SSH command constructions (`_assign_tx_ip_remote`, `_verify_ping`, and the teardown block) to build the remote command separately and apply `shlex.quote()` at each nesting level. The remote command string is now quoted as a whole for the local shell, preventing the single-quote collision issue. Commit a686b7b.

tests/e2e/replay_traffic.py
```diff
@@ -0,0 +1,104 @@
...
+                shell=True, check=True, capture_output=True,
+            )
+            subprocess.run(
+                f"ssh {shlex.quote(remote_host)}"
+                f" 'tcpreplay {loop_arg} -i {shlex.quote(iface)} {shlex.quote(remote_tmp)};"
+                f" rm -f {shlex.quote(remote_tmp)}'",
+                shell=True, check=True, capture_output=True,
+            )
+        else:
+            cmd = f"tcpreplay {loop_arg} -i {shlex.quote(iface)} {shlex.quote(tmp_path)}"
```
banidoru, line 85: **Double-quoting bug (same issue that was fixed in conftest.py).** The SSH command here still uses hardcoded single quotes around the remote command with `shlex.quote()` values inside. `shlex.quote()` wraps values in single quotes, which will collide with the outer single quotes and break the command.

`conftest.py` was correctly refactored to use the `shlex.quote(remote_cmd)` pattern — apply the same fix here:
```python
remote_cmd = (
    f"tcpreplay {loop_arg} -i {shlex.quote(iface)} {shlex.quote(remote_tmp)}; "
    f"rm -f {shlex.quote(remote_tmp)}"
)
subprocess.run(
    f"ssh {shlex.quote(remote_host)} {shlex.quote(remote_cmd)}",
    shell=True, check=True, capture_output=True,
)
```

tests/e2e/replay_traffic.py
```diff
@@ -0,0 +1,104 @@
...
+Module usage:
+    from replay_traffic import replay
+    replay("input.pcap", "10.123.0.1", "p1", count=1)
+"""
+
+import argparse
+import os
+import shlex
+import subprocess
+import sys
```
banidoru: **Unused import.** `sys` is imported but never used — remove it.

banidoru (approved):
- All 19 prior threads addressed and verified — good iteration
- Shell injection, teardown, checksums, error handling, SSH quoting all fixed in conftest.py
- Two minor new issues flagged inline (non-blocking)
- Approving — solid overall

tests/e2e/replay_traffic.py
```diff
@@ -0,0 +1,104 @@
...
+            subprocess.run(
+                f"ssh {shlex.quote(remote_host)}"
+                f" 'tcpreplay {loop_arg} -i {shlex.quote(iface)} {shlex.quote(remote_tmp)};"
+                f" rm -f {shlex.quote(remote_tmp)}'",
+                shell=True, check=True, capture_output=True,
+            )
+        else:
+            cmd = f"tcpreplay {loop_arg} -i {shlex.quote(iface)} {shlex.quote(tmp_path)}"
+            subprocess.run(cmd, shell=True, check=True, capture_output=True)
+    finally:
```
banidoru, line 87: **SSH double-quoting bug (same as conftest.py fix).** This still uses `shlex.quote()` *inside* hard-coded single quotes:
```python
f" 'tcpreplay ... -i {shlex.quote(iface)} {shlex.quote(remote_tmp)}; ...' "
```
If `iface` contains special chars, `shlex.quote` wraps it in single quotes which collide with the outer quotes — the exact bug that was fixed in `conftest.py` (commit a686b7b).

**Fix:** Apply the same pattern used in conftest — build the remote command separately, then quote the whole thing:
```python
remote_cmd = f"tcpreplay {loop_arg} -i {shlex.quote(iface)} {shlex.quote(remote_tmp)}; rm -f {shlex.quote(remote_tmp)}"
subprocess.run(
    f"ssh {shlex.quote(remote_host)} {shlex.quote(remote_cmd)}",
    shell=True, check=True, capture_output=True,
)
```

tests/e2e/conftest.py
```diff
@@ -0,0 +1,186 @@
...
+@pytest.fixture(scope="session")
+def infmon_env():
+    """Set up networking, verify connectivity, and yield config dict."""
+    mode = _env("INFMON_E2E_TX_MODE")
+
+    if mode not in ("local", "remote"):
+        pytest.fail(f"Invalid INFMON_E2E_TX_MODE: {mode!r}. Must be 'local' or 'remote'.")
+
+    # 1. Assign IPs
+    _assign_rx_ip()
```
banidoru, line 135: **Silent failure if InFMon doesn't start.** The polling loop runs 20 iterations but if the service never becomes active, execution falls through silently — the fixture yields a config dict and tests run against a non-running InFMon.

Add a failure after the loop:
```python
else:
    pytest.fail("InFMon did not become active within 10s")
```
(Using the `for/else` pattern, where `else` runs only if the loop completes without `break`.)

banidoru (commented):
Re-review (iteration 2) — all 20 prior threads resolved. Two new issues found:

- **SSH double-quoting bug in replay_traffic.py** — same pattern fixed in conftest.py wasn't applied here
- **Silent failure if InFMon doesn't start** — polling loop exhausts without failing the fixture

banidoru (approved):
All reviewers approved. LGTM.

tests/e2e/replay_traffic.py
```diff
@@ -0,0 +1,104 @@
...
+import os
+import shlex
+import subprocess
+import sys
+import tempfile
+
+from scapy.all import IP, IPv6, rdpcap, wrpcap
+
+
+def rewrite_dst_ip(packets, dst_ip: str):
```
banidoru, line 23: This commented-out `argparse` block is dead code and should be removed to improve clarity.

banidoru (commented):
- The SSH double-quoting bug in `replay_traffic.py` has been fixed.
- The silent failure issue in `conftest.py` where tests could run against a non-running `infmon` service is **still not addressed**.
- Found one new minor issue: dead/commented-out code in `replay_traffic.py`.

tests/e2e/replay_traffic.py
banidoru: Verified: no commented-out argparse block in the current code. Resolving.

tests/e2e/replay_traffic.py
banidoru: Confirmed still present. The SSH command on line 84 nests `shlex.quote()` inside hardcoded single quotes — same pattern that was correctly fixed in `conftest.py` using `shlex.quote(remote_cmd)` for the whole command string.

tests/e2e/replay_traffic.py
banidoru: Confirmed: `import sys` on line 18 is still unused. Should be removed.

tests/e2e/replay_traffic.py
banidoru: Confirmed still present — same double-quoting issue as thread above. Apply the same `shlex.quote(remote_cmd)` pattern used in `conftest.py`.

tests/e2e/conftest.py
banidoru: Confirmed still present. After the polling loop in `_ensure_infmon_running()`, there is no failure if the service never became active. Add a check after the loop: `if _run("systemctl is-active infmon", check=False).stdout.strip() != "active": pytest.fail("InFMon did not start within 10s")`.

banidoru (commented):
Iteration 2 re-review — 20 resolved threads verified, 1 more resolved (dead code removed). 4 threads still open:

- **SSH double-quoting in replay_traffic.py (lines 84-86):** `shlex.quote()` nested inside hardcoded single quotes — apply the same `shlex.quote(remote_cmd)` pattern used in `conftest.py`
- **Unused `sys` import (line 18):** remove it
- **Silent failure in `_ensure_infmon_running` (conftest.py line 133):** polling loop exits silently if service never starts — add a `pytest.fail()` after the loop
- No new issues found in the latest diff. Overall structure is solid.

tests/e2e/replay_traffic.py
r12f: Fixed in fd9dcb7: refactored the SSH command construction to build `remote_cmd` separately and wrap it with `shlex.quote()` for the local shell — same pattern used in conftest.py (a686b7b).

tests/e2e/replay_traffic.py
r12f: Fixed in fd9dcb7: removed the unused `sys` import.

tests/e2e/replay_traffic.py
r12f: Fixed in fd9dcb7 (same commit as the thread above): applied the `shlex.quote(remote_cmd)` pattern from conftest.py to the SSH command in replay_traffic.py.

tests/e2e/conftest.py
r12f: Fixed in fd9dcb7: added `else: pytest.fail("InFMon did not become active within 10s")` using the `for/else` pattern so the fixture fails explicitly instead of silently proceeding.

tests/e2e/replay_traffic.py
r12f: I checked the current state of `replay_traffic.py` and there is no commented-out argparse block at line 24 or elsewhere. The `argparse` import on line 14 is actively used by `main()`. This may have been referring to code that was already removed in a previous commit. No changes needed.

PR #142, opened 2026-04-21
fix: update helpers.py CLI args to match infmonctl positional syntax

### LINKED ISSUE #136: helpers.py CLI arguments do not match infmonctl flow-rule add syntax
## Bug Description

`tests/e2e/helpers.py` calls `infmonctl flow-rule add --name X --field Y` but infmonctl uses positional key=value syntax: `infmonctl flow-rule add name=X fields=Y max_keys=Z`.

## Steps to Reproduce

1. Run `make e2e`
2. `flow_rule_add()` in helpers.py fails because infmonctl doesn't recognize `--name` / `--field` flags

## Expected Behavior

helpers.py should use the correct CLI syntax matching infmonctl's actual interface.

## Fix

Update `flow_rule_add()` and `flow_rule_rm()` in `tests/e2e/helpers.py` to use positional `key=value` args.

r12f:
## Summary

`flow_rule_add()` in `tests/e2e/helpers.py` was using `--name`/`--field`/`--max-keys` flags, but infmonctl expects positional `key=value` args: `name=X fields=f1,f2 max_keys=N`.

`flow_rule_rm()` was using `--name` flag, but infmonctl expects a positional target argument.

Also updates `test_packet_replay.py` to pass fields as a list of field names instead of a dict, matching the new `flow_rule_add()` signature.

Closes #136

### DIFF
--- a/tests/e2e/helpers.py
+++ b/tests/e2e/helpers.py
@@ -37,6 +37,8 @@ def flow_rule_add(name: str, fields: List[str], max_keys: int = 0) -> None:
         fields: List of match field names (e.g. ["src_ip", "dst_ip"]).
         max_keys: Maximum number of flow keys (0 = unlimited).
     """
+    if not fields:
+        raise ValueError("fields must be non-empty; infmonctl requires at least one match field")
     args = ["flow-rule", "add", f"name={name}", f"fields={','.join(fields)}"]
     if max_keys > 0:
         args.append(f"max_keys={max_keys}")
--- a/tests/e2e/helpers.py
+++ b/tests/e2e/helpers.py
@@ -26,25 +26,29 @@ def _parse_json_output(result: subprocess.CompletedProcess) -> Any:
     return json.loads(output)
 
 
-def flow_rule_add(name: str, fields: Dict[str, str], max_keys: int = 0) -> None:
+def flow_rule_add(name: str, fields: List[str], max_keys: int = 0) -> None:
     """Add a flow rule via infmonctl.
+
+    Uses positional key=value syntax expected by infmonctl:
+        infmonctl flow-rule add name=<name> fields=<f1,f2,...> [max_keys=<N>]
 
     Args:
         name: Rule name.
-        fields: Dict of match field name to value.
+        fields: List of match field names (e.g. ["src_ip", "dst_ip"]).
         max_keys: Maximum number of flow keys (0 = unlimited).
     """
-    args = ["flow-rule", "add", "--name", name]
-    for field_name, field_value in fields.items():
-        args.extend(["--field", f"{field_name}={field_value}"])
+    args = ["flow-rule", "add", f"name={name}", f"fields={','.join(fields)}"]
     if max_keys > 0:
-        args.extend(["--max-keys", str(max_keys)])
+        args.append(f"max_keys={max_keys}")
     _run_infmonctl(*args)
 
 
 def flow_rule_rm(name: str) -> None:
-    """Remove a flow rule by name."""
-    _run_infmonctl("flow-rule", "rm", "--name", name)
+    """Remove a flow rule by name.
+
+    Uses positional syntax: infmonctl flow-rule rm <name>
+    """
+    _run_infmonctl("flow-rule", "rm", name)
 
 
 def flow_rule_list() -> List[Dict[str, Any]]:
--- a/tests/e2e/test_packet_replay.py
+++ b/tests/e2e/test_packet_replay.py
@@ -75,11 +75,11 @@ def test_packet_replay(scenario: str, infmon_env: dict) -> None:
     if config_path.exists():
         with open(config_path) as f:
             config = json.load(f)
-        fields = config.get("fields", {})
+        fields = config.get("fields", [])
         max_keys = config.get("max_keys", 0)
     else:
         # Default: match all traffic on the RX interface
-        fields = {}
+        fields = []
         max_keys = 0  # 0 = unlimited keys (sentinel value in InFMon API)
 
     flow_rule_add(name=scenario, fields=fields, max_keys=max_keys)
--- a/tests/e2e/test_packet_replay.py
+++ b/tests/e2e/test_packet_replay.py
@@ -78,8 +78,8 @@ def test_packet_replay(scenario: str, infmon_env: dict) -> None:
         fields = config.get("fields", [])
         max_keys = config.get("max_keys", 0)
     else:
-        # Default: match all traffic on the RX interface
-        fields = []
+        # Default: match on standard 5-tuple fields
+        fields = ["src_ip", "dst_ip", "src_port", "dst_port", "protocol"]
         max_keys = 0  # 0 = unlimited keys (sentinel value in InFMon API)
 
     flow_rule_add(name=scenario, fields=fields, max_keys=max_keys)


tests/e2e/helpers.py
```diff
@@ -26,25 +26,29 @@ def _parse_json_output(result: subprocess.CompletedProcess) -> Any:
...
     Args:
         name: Rule name.
-        fields: Dict of match field name to value.
+        fields: List of match field names (e.g. ["src_ip", "dst_ip"]).
         max_keys: Maximum number of flow keys (0 = unlimited).
     """
-    args = ["flow-rule", "add", "--name", name]
-    for field_name, field_value in fields.items():
-        args.extend(["--field", f"{field_name}={field_value}"])
+    args = ["flow-rule", "add", f"name={name}", f"fields={','.join(fields)}"]
```
banidoru, line 42: If `fields` is empty, this produces `fields=` which `infmonctl` rejects (`fields=<field,...> is required`, exit 2). The default in `test_packet_replay.py` is `fields = []`, so any scenario without a `config.json` will hit this.

Consider either:
- Raising early: `if not fields: raise ValueError("fields must be non-empty")`
- Or omitting the arg and letting the CLI's own error propagate (but that gives a less helpful traceback).

tests/e2e/test_packet_replay.py
```diff
@@ -75,11 +75,11 @@ def test_packet_replay(scenario: str, infmon_env: dict) -> None:
     if config_path.exists():
         with open(config_path) as f:
             config = json.load(f)
-        fields = config.get("fields", {})
+        fields = config.get("fields", [])
```
banidoru, line 78: Default `fields = []` will produce a CLI invocation with `fields=` which `infmonctl` rejects. If the intent is "match all traffic" with no field filters, the CLI may need a different sentinel (e.g. `fields=*` or omitting the arg entirely). Worth confirming what the actual "match all" invocation looks like — the previous `fields = {}` had the same issue with the old `--field` flag approach (no `--field` args passed), so this may be pre-existing.

banidoru (commented):
- Positional `key=value` syntax now matches the actual `infmonctl` CLI parser — correctness fix looks right.
- `fields` type change from `Dict[str, str]` → `List[str]` is consistent with the CLI's `fields=f1,f2,...` contract.
- `flow_rule_rm` switching from `--name` to positional aligns with the Rust `FlowRuleCommands::Rm { target }` positional arg.
- **Edge case:** empty `fields` list produces `fields=` on the command line — the CLI rejects empty fields with exit 2. Consider guarding with a `ValueError` or doc note so failures surface clearly in Python rather than a cryptic subprocess error.
- `flow_rule_list`, `get_flow_stats`, and other helpers still use `--name`/`--json` flags — verify those haven't also moved to positional syntax (they look fine based on current CLI source, but worth a quick check).

tests/e2e/helpers.py
r12f: Good catch — added an early `ValueError` in `flow_rule_add()` when `fields` is empty, so callers get a clear Python-side error instead of a confusing CLI rejection. Also changed the default in `test_packet_replay.py` to use the standard 5-tuple so no-config scenarios get a valid invocation. Fixed in 7d20ac1.

tests/e2e/test_packet_replay.py
r12f: Agreed this was pre-existing but now surfaced by the syntax change. Fixed: the default is now `["src_ip", "dst_ip", "src_port", "dst_port", "protocol"]` (standard 5-tuple) instead of `[]`. Additionally, `flow_rule_add()` now raises `ValueError` early on empty fields to catch any future callers. Fixed in 7d20ac1.

tests/e2e/helpers.py
```diff
@@ -26,25 +26,31 @@ def _parse_json_output(result: subprocess.CompletedProcess) -> Any:
     return json.loads(output)
 
 
-def flow_rule_add(name: str, fields: Dict[str, str], max_keys: int = 0) -> None:
+def flow_rule_add(name: str, fields: List[str], max_keys: int = 0) -> None:
     """Add a flow rule via infmonctl.
```
banidoru, line 31: This change from `--name <name>` to a positional `<name>` for removal is a breaking change for the `infmonctl` CLI. Is the CLI tool itself being updated in another PR? If this PR only updates the tests, they will fail against the existing tool.

tests/e2e/helpers.py
```diff
@@ -26,25 +26,31 @@ def _parse_json_output(result: subprocess.CompletedProcess) -> Any:
     return json.loads(output)
```
banidoru, line 28: This switch to positional `key=value` arguments is a significant change to the `infmonctl` tools CLI. This will break any existing scripts or users who rely on the `--field` and `--name` flags. A version bump and a clear release note in the main tools documentation would be essential.

banidoru (changes requested):
This PR simplifies the test helpers, but it seems to be based on an unreleased, breaking change in the `infmonctl` CLI. The tests will fail without the corresponding CLI update.

- The CLI argument parsing is changed from flags (`--name`) to positional `key=value` pairs.
- This is a breaking change and needs to be coordinated with the tools release.

tests/e2e/helpers.py
r12f: To clarify: this PR updates the *test helpers* to match the existing `infmonctl` CLI positional syntax — it does not change the CLI tool itself. The CLI already uses `key=value` positional args; the tests were out of date with the old `--field`/`--name` flag style. No breaking change to the tool or its users.

tests/e2e/helpers.py
r12f: No breaking change here — this PR aligns the test helpers with the *existing* `infmonctl` CLI syntax. The CLI already expects positional `<name>` for removal; the old tests were using the outdated `--name` flag form. The CLI itself is not being changed.

tests/e2e/helpers.py
```diff
@@ -26,25 +26,31 @@ def _parse_json_output(result: subprocess.CompletedProcess) -> Any:
...
-        fields: Dict of match field name to value.
+        fields: List of match field names (e.g. ["src_ip", "dst_ip"]).
         max_keys: Maximum number of flow keys (0 = unlimited).
     """
-    args = ["flow-rule", "add", "--name", name]
-    for field_name, field_value in fields.items():
-        args.extend(["--field", f"{field_name}={field_value}"])
+    if not fields:
+        raise ValueError("fields must be non-empty; infmonctl requires at least one match field")
+    args = ["flow-rule", "add", f"name={name}", f"fields={','.join(fields)}"]
```
banidoru, line 42: Neither `name` nor individual `fields` entries are validated before being interpolated into positional `key=value` args. A field name containing `,` (e.g. from a bad config.json) would silently produce extra phantom fields, and a `name` containing `=` or spaces would break parsing.

Low-risk since this is test code, but a quick `assert all(re.match(r'^[a-zA-Z0-9_]+$', f) for f in fields)` would catch config errors early.

tests/e2e/helpers.py
```diff
@@ -26,25 +26,31 @@ def _parse_json_output(result: subprocess.CompletedProcess) -> Any:
...
-        args.extend(["--max-keys", str(max_keys)])
+        args.append(f"max_keys={max_keys}")
     _run_infmonctl(*args)
 
 
 def flow_rule_rm(name: str) -> None:
-    """Remove a flow rule by name."""
-    _run_infmonctl("flow-rule", "rm", "--name", name)
+    """Remove a flow rule by name.
+
```
banidoru, line 50: `flow_rule_rm` now passes `name` as a bare positional arg. Unlike `flow_rule_add`, there's no validation here. Consider the same kind of guard — if `name` is empty or contains whitespace, the CLI invocation will be silently wrong.

banidoru (commented):
- **Verdict: Approve** — small, well-scoped fix aligning test helpers with actual CLI syntax
- No input sanitization on `name` or `fields` entries — commas, spaces, or `=` in values would silently produce malformed CLI args; consider validating with a regex like `^[a-zA-Z0-9_]+$`
- `flow_rule_rm` positional change looks correct but lacks the same defensive validation added to `flow_rule_add`
- Empty-fields guard and 5-tuple default (commit 7d20ac1) address prior feedback well

tests/e2e/helpers.py
r12f: Valid point, but out of scope for this PR which is strictly about fixing the CLI arg syntax to match infmonctl. Input validation for test config values would be a separate hardening concern. Happy to track as a follow-up if desired.

tests/e2e/helpers.py
r12f: Acknowledged — same as above, input validation for name/fields is a reasonable hardening step but out of scope for this syntax-fix PR. The pre-existing code had no such validation either. Happy to address in a follow-up.

banidoru (approved):
- All prior feedback addressed in 7d20ac1: empty-fields `ValueError` guard and 5-tuple default both look correct
- Positional `key=value` syntax matches `infmonctl` CLI contract
- `fields: Dict[str,str]` → `List[str]` signature change is consistent and callers updated

banidoru (approved):
All prior concerns addressed. Changes look good:
- Empty-fields guard added in `flow_rule_add()` — clean fail-fast
- Default fields changed to standard 5-tuple — sensible fallback
- Positional syntax correctly matches `infmonctl` CLI
- All 6 previous review threads resolved by author

LGTM.

banidoru (approved):
- All prior concerns addressed in 7d20ac1: empty-fields `ValueError` guard and 5-tuple default both look correct
- Positional `key=value` syntax matches `infmonctl` CLI expectations — verified construction logic
- Type signature update `Dict[str,str]` → `List[str]` is consistent across helpers and test callsites

banidoru (approved):
All reviewers approved. LGTM.

PR #144, opened 2026-04-21
feat(backend): add VPP plugin shared library target

### LINKED ISSUE #138: VPP plugin shared library (.so) not built by CMakeLists.txt
## Feature Description

`src/backend/CMakeLists.txt` only builds static libraries and test executables. There is no shared library target for the VPP plugin (`.so`), which is needed to load InFMon into VPP at runtime.

The VPP node registration code exists in `src/backend/src/vpp/infmon_nodes.c` but is gated behind `#ifdef INFMON_VPP_BUILD` and never compiled.

## Proposed Solution

Add a `MODULE` or `SHARED` library target that:
1. Compiles `infmon_nodes.c` with `-DINFMON_VPP_BUILD`
2. Links against VPP libraries
3. Produces `infmon_plugin.so` that can be placed in VPP's plugin directory

## Note

This depends on VPP dev headers being available. The BF3 dev machine has VPP v26.02 installed.

r12f:
## Summary

Add a CMake `MODULE` library target (`infmon_plugin`) that compiles `infmon_nodes.c` with `-DINFMON_VPP_BUILD` and links against VPP libraries to produce `infmon_plugin.so` for VPP runtime loading.

## Changes

- New CMake option `ENABLE_VPP_PLUGIN` (default `OFF`) in `src/backend/CMakeLists.txt`
- When enabled, locates VPP dev headers (`vlib/vlib.h`) and libraries (`vlib`, `vnet`, `vlibmemory`, `vppinfra`)
- Produces `infmon_plugin.so` (no `lib` prefix) suitable for VPP's plugin directory
- Links against all internal InFMon libraries (parser, flow_rule, counter_table, snapshot, stats_segment, graph_node)
- Depends on generated API header via `infmon_api_generated` target

## Usage

```bash
cmake .. -DENABLE_VPP_PLUGIN=ON
make infmon_plugin
cp infmon_plugin.so /usr/lib/vpp_plugins/
```

## Test Plan

- [x] Default build (without flag) — all 10 tests pass, no regressions
- [x] `-DENABLE_VPP_PLUGIN=ON` — CMake configures successfully, plugin target compiles `infmon_nodes.c` with VPP headers

Closes #138

### DIFF
--- a/src/backend/CMakeLists.txt
+++ b/src/backend/CMakeLists.txt
@@ -172,9 +172,14 @@ add_test(NAME integration COMMAND test_integration)
 option(ENABLE_VPP_PLUGIN "Build infmon_plugin.so for VPP runtime loading" OFF)
 
 if(ENABLE_VPP_PLUGIN)
+    # Static libraries linked into the MODULE target must be position-independent
+    foreach(_tgt infmon_graph_node infmon_parser infmon_flow_rule
+                 infmon_counter_table infmon_snapshot infmon_stats_segment)
+        set_target_properties(${_tgt} PROPERTIES POSITION_INDEPENDENT_CODE ON)
+    endforeach()
+
     # Locate VPP development headers and libraries
     find_path(VPP_INCLUDE_DIR vlib/vlib.h
-        PATHS /usr/include /usr/local/include
         DOC "VPP development headers"
     )
     find_library(VPP_LIB_VLIB NAMES vlib DOC "VPP vlib library")
@@ -211,8 +216,10 @@ if(ENABLE_VPP_PLUGIN)
         infmon_snapshot
         infmon_stats_segment
     )
-    foreach(_lib ${VPP_LIB_VLIB} ${VPP_LIB_VNET} ${VPP_LIB_VLIBMEMORY} ${VPP_LIB_VPPINFRA})
-        target_link_libraries(infmon_plugin PRIVATE ${_lib})
+    foreach(_var VPP_LIB_VLIB VPP_LIB_VNET VPP_LIB_VLIBMEMORY VPP_LIB_VPPINFRA)
+        if(${_var})
+            target_link_libraries(infmon_plugin PRIVATE ${${_var}})
+        endif()
     endforeach()
 
     # Ensure the generated API header is built before the plugin
--- a/src/backend/CMakeLists.txt
+++ b/src/backend/CMakeLists.txt
@@ -168,6 +168,77 @@ target_link_libraries(test_integration PRIVATE
 target_include_directories(test_integration PRIVATE include)
 add_test(NAME integration COMMAND test_integration)
 
+# --- VPP plugin shared library (.so) ---
+option(ENABLE_VPP_PLUGIN "Build infmon_plugin.so for VPP runtime loading" OFF)
+
+if(ENABLE_VPP_PLUGIN)
+    # Locate VPP development headers and libraries
+    find_path(VPP_INCLUDE_DIR vlib/vlib.h
+        PATHS /usr/include /usr/local/include
+        DOC "VPP development headers"
+    )
+    find_library(VPP_LIB_VLIB NAMES vlib
+        PATHS /usr/lib /usr/lib/x86_64-linux-gnu /usr/lib/aarch64-linux-gnu
+        DOC "VPP vlib library"
+    )
+    find_library(VPP_LIB_VNET NAMES vnet
+        PATHS /usr/lib /usr/lib/x86_64-linux-gnu /usr/lib/aarch64-linux-gnu
+        DOC "VPP vnet library"
+    )
+    find_library(VPP_LIB_VLIBMEMORY NAMES vlibmemory
+        PATHS /usr/lib /usr/lib/x86_64-linux-gnu /usr/lib/aarch64-linux-gnu
+        DOC "VPP vlibmemory library"
+    )
+    find_library(VPP_LIB_VPPINFRA NAMES vppinfra
+        PATHS /usr/lib /usr/lib/x86_64-linux-gnu /usr/lib/aarch64-linux-gnu
+        DOC "VPP vppinfra library"
+    )
+
+    if(NOT VPP_INCLUDE_DIR)
+        message(FATAL_ERROR "ENABLE_VPP_PLUGIN is ON but VPP dev headers not found. Install vpp-dev.")
+    endif()
+
+    add_library(infmon_plugin MODULE
+        src/vpp/infmon_nodes.c
+    )
+    set_target_properties(infmon_plugin PROPERTIES
+        PREFIX ""                              # produce infmon_plugin.so, not libinfmon_plugin.so
+        OUTPUT_NAME "infmon_plugin"
+        C_VISIBILITY_PRESET default
+    )
+    target_compile_definitions(infmon_plugin PRIVATE INFMON_VPP_BUILD)
+    target_include_directories(infmon_plugin PRIVATE
+        include
+        ${VPP_INCLUDE_DIR}
+        ${CMAKE_CURRENT_BINARY_DIR}/generated  # for generated API header
+    )
+    target_link_libraries(infmon_plugin PRIVATE
+        infmon_graph_node
+        infmon_parser
+        infmon_flow_rule
+        infmon_counter_table
+        infmon_snapshot
+        infmon_stats_segment
+    )
+    if(VPP_LIB_VLIB)
+        target_link_libraries(infmon_plugin PRIVATE ${VPP_LIB_VLIB})
+    endif()
+    if(VPP_LIB_VNET)
+        target_link_libraries(infmon_plugin PRIVATE ${VPP_LIB_VNET})
+    endif()
+    if(VPP_LIB_VLIBMEMORY)
+        target_link_libraries(infmon_plugin PRIVATE ${VPP_LIB_VLIBMEMORY})
+    endif()
+    if(VPP_LIB_VPPINFRA)
+        target_link_libraries(infmon_plugin PRIVATE ${VPP_LIB_VPPINFRA})
+    endif()
+
+    # Ensure the generated API header is built before the plugin
+    add_dependencies(infmon_plugin infmon_api_generated)
+
+    message(STATUS "VPP plugin target enabled: infmon_plugin.so")
+endif()
+
 if(ENABLE_FUZZER)
     add_executable(fuzz_erspan_parser
         test/fuzz_erspan_parser.cc
--- a/src/backend/CMakeLists.txt
+++ b/src/backend/CMakeLists.txt
@@ -177,25 +177,16 @@ if(ENABLE_VPP_PLUGIN)
         PATHS /usr/include /usr/local/include
         DOC "VPP development headers"
     )
-    find_library(VPP_LIB_VLIB NAMES vlib
-        PATHS /usr/lib /usr/lib/x86_64-linux-gnu /usr/lib/aarch64-linux-gnu
-        DOC "VPP vlib library"
-    )
-    find_library(VPP_LIB_VNET NAMES vnet
-        PATHS /usr/lib /usr/lib/x86_64-linux-gnu /usr/lib/aarch64-linux-gnu
-        DOC "VPP vnet library"
-    )
-    find_library(VPP_LIB_VLIBMEMORY NAMES vlibmemory
-        PATHS /usr/lib /usr/lib/x86_64-linux-gnu /usr/lib/aarch64-linux-gnu
-        DOC "VPP vlibmemory library"
-    )
-    find_library(VPP_LIB_VPPINFRA NAMES vppinfra
-        PATHS /usr/lib /usr/lib/x86_64-linux-gnu /usr/lib/aarch64-linux-gnu
-        DOC "VPP vppinfra library"
-    )
+    find_library(VPP_LIB_VLIB NAMES vlib DOC "VPP vlib library")
+    find_library(VPP_LIB_VNET NAMES vnet DOC "VPP vnet library")
+    find_library(VPP_LIB_VLIBMEMORY NAMES vlibmemory DOC "VPP vlibmemory library")
+    find_library(VPP_LIB_VPPINFRA NAMES vppinfra DOC "VPP vppinfra library")
 
     if(NOT VPP_INCLUDE_DIR)
         message(FATAL_ERROR "ENABLE_VPP_PLUGIN is ON but VPP dev headers not found. Install vpp-dev.")
+    endif()
+    if(NOT VPP_LIB_VLIB OR NOT VPP_LIB_VNET)
+        message(FATAL_ERROR "ENABLE_VPP_PLUGIN is ON but required VPP libraries (vlib, vnet) not found.")
     endif()
 
     add_library(infmon_plugin MODULE
@@ -220,21 +211,18 @@ if(ENABLE_VPP_PLUGIN)
         infmon_snapshot
         infmon_stats_segment
     )
-    if(VPP_LIB_VLIB)
-        target_link_libraries(infmon_plugin PRIVATE ${VPP_LIB_VLIB})
-    endif()
-    if(VPP_LIB_VNET)
-        target_link_libraries(infmon_plugin PRIVATE ${VPP_LIB_VNET})
-    endif()
-    if(VPP_LIB_VLIBMEMORY)
-        target_link_libraries(infmon_plugin PRIVATE ${VPP_LIB_VLIBMEMORY})
-    endif()
-    if(VPP_LIB_VPPINFRA)
-        target_link_libraries(infmon_plugin PRIVATE ${VPP_LIB_VPPINFRA})
-    endif()
+    foreach(_lib ${VPP_LIB_VLIB} ${VPP_LIB_VNET} ${VPP_LIB_VLIBMEMORY} ${VPP_LIB_VPPINFRA})
+        target_link_libraries(infmon_plugin PRIVATE ${_lib})
+    endforeach()
 
     # Ensure the generated API header is built before the plugin
     add_dependencies(infmon_plugin infmon_api_generated)
+
+    # Install to VPP plugin directory
+    include(GNUInstallDirs)
+    install(TARGETS infmon_plugin
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}/vpp_plugins
+    )
 
     message(STATUS "VPP plugin target enabled: infmon_plugin.so")
 endif()


src/backend/CMakeLists.txt
```diff
@@ -168,6 +168,77 @@ target_link_libraries(test_integration PRIVATE
...
+        DOC "VPP vnet library"
+    )
+    find_library(VPP_LIB_VLIBMEMORY NAMES vlibmemory
+        PATHS /usr/lib /usr/lib/x86_64-linux-gnu /usr/lib/aarch64-linux-gnu
+        DOC "VPP vlibmemory library"
+    )
+    find_library(VPP_LIB_VPPINFRA NAMES vppinfra
+        PATHS /usr/lib /usr/lib/x86_64-linux-gnu /usr/lib/aarch64-linux-gnu
+        DOC "VPP vppinfra library"
+    )
```
banidoru: Headers are validated with `FATAL_ERROR`, but the VPP libraries are all individually optional (lines 215-224). If `vlib` or `vnet` isn't found, the plugin will compile but fail to load at runtime with missing symbols. Consider adding a fatal check for at least the core libs:
```cmake
if(NOT VPP_LIB_VLIB OR NOT VPP_LIB_VNET)
    message(FATAL_ERROR "ENABLE_VPP_PLUGIN is ON but required VPP libraries (vlib, vnet) not found.")
endif()
```

src/backend/CMakeLists.txt
```diff
@@ -168,6 +168,77 @@ target_link_libraries(test_integration PRIVATE
...
+# --- VPP plugin shared library (.so) ---
+option(ENABLE_VPP_PLUGIN "Build infmon_plugin.so for VPP runtime loading" OFF)
+
+if(ENABLE_VPP_PLUGIN)
+    # Locate VPP development headers and libraries
+    find_path(VPP_INCLUDE_DIR vlib/vlib.h
+        PATHS /usr/include /usr/local/include
+        DOC "VPP development headers"
+    )
+    find_library(VPP_LIB_VLIB NAMES vlib
```
banidoru: The `PATHS` lists are hardcoded to two arch-specific lib dirs. This won't find libraries on less common architectures or custom install prefixes. Consider using `pkg-config` if VPP ships `.pc` files, or at minimum add `PATH_SUFFIXES` and rely on CMake's default search paths:
```cmake
find_library(VPP_LIB_VLIB NAMES vlib)
```
CMake already searches `/usr/lib`, `/usr/local/lib`, and the arch-qualified paths via `CMAKE_LIBRARY_ARCHITECTURE`.

src/backend/CMakeLists.txt
```diff
@@ -168,6 +168,77 @@ target_link_libraries(test_integration PRIVATE
...
+        infmon_stats_segment
+    )
+    if(VPP_LIB_VLIB)
+        target_link_libraries(infmon_plugin PRIVATE ${VPP_LIB_VLIB})
+    endif()
+    if(VPP_LIB_VNET)
+        target_link_libraries(infmon_plugin PRIVATE ${VPP_LIB_VNET})
+    endif()
+    if(VPP_LIB_VLIBMEMORY)
+        target_link_libraries(infmon_plugin PRIVATE ${VPP_LIB_VLIBMEMORY})
```
banidoru: Missing an `install()` rule. The PR description tells users to `cp` manually. Adding an install target with a configurable path would integrate better with packaging:
```cmake
install(TARGETS infmon_plugin
    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}/vpp_plugins
)
```

src/backend/CMakeLists.txt
```diff
@@ -168,6 +168,77 @@ target_link_libraries(test_integration PRIVATE
...
+        OUTPUT_NAME "infmon_plugin"
+        C_VISIBILITY_PRESET default
+    )
+    target_compile_definitions(infmon_plugin PRIVATE INFMON_VPP_BUILD)
+    target_include_directories(infmon_plugin PRIVATE
+        include
+        ${VPP_INCLUDE_DIR}
+        ${CMAKE_CURRENT_BINARY_DIR}/generated  # for generated API header
+    )
+    target_link_libraries(infmon_plugin PRIVATE
```
banidoru, line 211: These four `if/endif` blocks are repetitive. A loop would reduce duplication:
```cmake
foreach(_lib ${VPP_LIB_VLIB} ${VPP_LIB_VNET} ${VPP_LIB_VLIBMEMORY} ${VPP_LIB_VPPINFRA})
    target_link_libraries(infmon_plugin PRIVATE ${_lib})
endforeach()
```
(Only works if the variables are guaranteed set — which ties back to the "fatal on missing" comment.)

banidoru (changes requested):
- **VPP libraries silently optional**: Headers missing → `FATAL_ERROR`, but all four VPP libs (`vlib`, `vnet`, `vlibmemory`, `vppinfra`) are individually guarded with `if(VPP_LIB_*)` — a plugin built without `vlib` will fail at VPP load time. Should fatal if core libs are missing.
- **No `install()` target**: Users must manually `cp` the `.so`; an install rule with a configurable destination would be more robust.
- **Hardcoded arch paths**: Only `x86_64-linux-gnu` and `aarch64-linux-gnu` are listed; other arches (e.g. `ppc64le`, `s390x`) or custom prefixes will silently miss libraries.
- **No SOVERSION or RPATH handling**: May cause issues depending on VPP's plugin loader expectations.
- Overall structure is clean and the gating via `option()` is the right approach.

src/backend/CMakeLists.txt
```diff
@@ -168,6 +168,77 @@ target_link_libraries(test_integration PRIVATE
...
+    find_library(VPP_LIB_VPPINFRA NAMES vppinfra
+        PATHS /usr/lib /usr/lib/x86_64-linux-gnu /usr/lib/aarch64-linux-gnu
+        DOC "VPP vppinfra library"
+    )
+
+    if(NOT VPP_INCLUDE_DIR)
+        message(FATAL_ERROR "ENABLE_VPP_PLUGIN is ON but VPP dev headers not found. Install vpp-dev.")
+    endif()
+
+    add_library(infmon_plugin MODULE
```
banidoru, line 197: This new source file  is referenced here but was not added to the pull request. The build will fail without it.

banidoru (changes requested):
Overall, the CMake logic is a good start, but it's incomplete and not robust. The most critical issue is that the  file is missing from the PR, which makes the new target unbuildable. \n\nAdditionally, the library finding logic should be improved for portability, and an  rule is needed to make the plugin deployable. Please address the inline comments.

src/backend/CMakeLists.txt
r12f: Good catch — added a `FATAL_ERROR` guard for `VPP_LIB_VLIB` and `VPP_LIB_VNET` so the build fails immediately if core libs are missing, rather than producing a plugin that fails to load at runtime. Fixed in 0392df3.

src/backend/CMakeLists.txt
r12f: Agreed — removed the hardcoded `PATHS` lists. CMake's default search already covers `/usr/lib`, `/usr/local/lib`, and arch-qualified dirs via `CMAKE_LIBRARY_ARCHITECTURE`. Fixed in 0392df3.

src/backend/CMakeLists.txt
r12f: Added an `install(TARGETS infmon_plugin LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}/vpp_plugins)` rule with `GNUInstallDirs` so packaging picks it up automatically. Fixed in 0392df3.

src/backend/CMakeLists.txt
r12f: Replaced the four `if/endif` blocks with a `foreach` loop. This is safe now that vlib and vnet are fatal-checked above — the optional libs (vlibmemory, vppinfra) simply expand to empty if not found and the loop skips them. Fixed in 0392df3.

src/backend/CMakeLists.txt
```diff
@@ -168,6 +168,65 @@ target_link_libraries(test_integration PRIVATE
...
+    # Install to VPP plugin directory
+    include(GNUInstallDirs)
+    install(TARGETS infmon_plugin
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}/vpp_plugins
+    )
+
+    message(STATUS "VPP plugin target enabled: infmon_plugin.so")
+endif()
+
 if(ENABLE_FUZZER)
```
banidoru, line 237: This dependency on `infmon_api_generated` is implicitly coupled to the API file stem being `infmon` (set in `vpp_generate_api_header`). If the API file is ever renamed, this breaks silently at build time. Consider having `vpp_generate_api_header` export the target name to a variable (e.g. `INFMON_API_TARGET`) and referencing that here:
```cmake
add_dependencies(infmon_plugin ${INFMON_API_TARGET})
```

banidoru (commented):
- **Missing source file**: `src/vpp/infmon_nodes.c` not in this PR — build will fail (already noted)
- **Silent library omission**: VPP libs are optional but the plugin needs them at runtime — should fatal on missing core libs (already noted)
- **Hardcoded arch paths**: `find_library` should rely on CMake defaults or pkg-config (already noted)
- **No install rule**: manual `cp` is fragile; add `install(TARGETS ...)` (already noted)
- **Implicit target coupling**: `add_dependencies(infmon_plugin infmon_api_generated)` relies on the stem of the API file being exactly `infmon` — a rename breaks silently. Consider storing the target name in a variable from `vpp_generate_api_header` and referencing that.

src/backend/CMakeLists.txt
r12f: Valid catch — `src/vpp/infmon_nodes.c` is referenced but not included in this PR. Since `ENABLE_VPP_PLUGIN` defaults to OFF, CI won't fail, but the plugin target is unbuildable without it. This needs the PR author to add the source file. Leaving unresolved for author attention.

src/backend/CMakeLists.txt
r12f: Reasonable suggestion, but this is a minor coupling concern — the API file and its generated target are defined 30 lines above in the same CMakeLists.txt, so a rename would be caught immediately. Exporting a variable from `vpp_generate_api_header` would be cleaner but is a separate improvement to the CMake module. Leaving as-is for now.

src/backend/CMakeLists.txt
banidoru: Still open — `src/vpp/infmon_nodes.c` is still not included in this PR. The plugin target remains unbuildable when `ENABLE_VPP_PLUGIN=ON`.

src/backend/CMakeLists.txt
```diff
@@ -168,6 +168,65 @@ target_link_libraries(test_integration PRIVATE
...
+    target_include_directories(infmon_plugin PRIVATE
+        include
+        ${VPP_INCLUDE_DIR}
+        ${CMAKE_CURRENT_BINARY_DIR}/generated  # for generated API header
+    )
+    target_link_libraries(infmon_plugin PRIVATE
+        infmon_graph_node
+        infmon_parser
+        infmon_flow_rule
+        infmon_counter_table
```
banidoru, line 215: These internal static libraries (`infmon_graph_node`, `infmon_parser`, etc.) are linked `PRIVATE` into a `MODULE` target. On Linux, MODULE targets produce position-independent `.so` files, so all linked static libraries must be compiled with `-fPIC`. If these targets don't set `POSITION_INDEPENDENT_CODE ON`, the linker will fail with relocation errors.

Either:
1. Add `set_target_properties(<lib> PROPERTIES POSITION_INDEPENDENT_CODE ON)` to each internal lib, or
2. Set `CMAKE_POSITION_INDEPENDENT_CODE ON` globally (may be too broad), or
3. Guard it under `if(ENABLE_VPP_PLUGIN)` to only enable PIC when building the plugin.

banidoru (commented):
Re-review (iteration 2):

- **Previous feedback addressed:** Fatal checks for core libs, hardcoded PATHS removed, install rule added, foreach loop — all look good ✅
- **Still open:** `src/vpp/infmon_nodes.c` missing from PR — plugin target unbuildable with `ENABLE_VPP_PLUGIN=ON`
- **New concern:** Internal libraries (`infmon_graph_node`, `infmon_parser`, etc.) linked via `PRIVATE` into a `MODULE` target must be compiled with `-fPIC`. If these are static libraries built without `CMAKE_POSITION_INDEPENDENT_CODE ON`, the link will fail on most Linux toolchains. Ensure these targets set `POSITION_INDEPENDENT_CODE ON` or are built as shared/object libraries.
- `find_path` for `VPP_INCLUDE_DIR` still has explicit `PATHS /usr/include /usr/local/include` — harmless (these are default search paths) but redundant; consider removing for consistency with the library lookups.

src/backend/CMakeLists.txt
banidoru: Still unresolved — `src/vpp/infmon_nodes.c` is not included in the PR. The plugin target is unbuildable with `ENABLE_VPP_PLUGIN=ON`.

src/backend/CMakeLists.txt
banidoru: Still unresolved — no `POSITION_INDEPENDENT_CODE` handling added. Linking static libs without `-fPIC` into a MODULE target will cause relocation errors on Linux.

banidoru (changes requested):
- **Missing source file:** `src/vpp/infmon_nodes.c` still not in PR — plugin target is unbuildable with `ENABLE_VPP_PLUGIN=ON`
- **No `-fPIC` handling:** Static libs linked into the MODULE target need `POSITION_INDEPENDENT_CODE ON` or the linker will emit relocation errors on Linux
- **Prior fixes look good:** FATAL_ERROR guards for core libs, hardcoded PATHS removal, foreach loop, install rule, GNUInstallDirs — all addressed cleanly
- **`find_path` still uses hardcoded PATHS:** `find_path(VPP_INCLUDE_DIR ...)` still has `PATHS /usr/include /usr/local/include` — these are already in CMake default search, so harmless but redundant

src/backend/CMakeLists.txt
banidoru: Still open — `src/vpp/infmon_nodes.c` is still missing from the PR. The plugin target remains unbuildable when `ENABLE_VPP_PLUGIN=ON`.

src/backend/CMakeLists.txt
banidoru: Still open — no `POSITION_INDEPENDENT_CODE` property has been added to the internal static libraries. Linking non-PIC static libs into a MODULE target will produce relocation errors on most Linux toolchains.

src/backend/CMakeLists.txt
```diff
@@ -168,6 +168,65 @@ target_link_libraries(test_integration PRIVATE
 target_include_directories(test_integration PRIVATE include)
 add_test(NAME integration COMMAND test_integration)
 
+# --- VPP plugin shared library (.so) ---
+option(ENABLE_VPP_PLUGIN "Build infmon_plugin.so for VPP runtime loading" OFF)
+
+if(ENABLE_VPP_PLUGIN)
+    # Locate VPP development headers and libraries
+    find_path(VPP_INCLUDE_DIR vlib/vlib.h
```
banidoru, line 182: `find_path` still has `PATHS /usr/include /usr/local/include` — these are already CMake defaults. This was cleaned up for `find_library` but not here. Remove the `PATHS` clause for consistency:
```cmake
find_path(VPP_INCLUDE_DIR vlib/vlib.h
    DOC "VPP development headers"
)
```

src/backend/CMakeLists.txt
```diff
@@ -168,6 +168,65 @@ target_link_libraries(test_integration PRIVATE
...
+        infmon_flow_rule
+        infmon_counter_table
+        infmon_snapshot
+        infmon_stats_segment
+    )
+    foreach(_lib ${VPP_LIB_VLIB} ${VPP_LIB_VNET} ${VPP_LIB_VLIBMEMORY} ${VPP_LIB_VPPINFRA})
+        target_link_libraries(infmon_plugin PRIVATE ${_lib})
+    endforeach()
+
+    # Ensure the generated API header is built before the plugin
```
banidoru, line 225: If `VPP_LIB_VLIBMEMORY` or `VPP_LIB_VPPINFRA` is not found, the variable holds `*-NOTFOUND` which gets passed as a literal string to the linker. The `foreach` doesn't skip these — CMake expands the variable to the `-NOTFOUND` sentinel.

Guard optional libs:
```cmake
foreach(_var VPP_LIB_VLIB VPP_LIB_VNET VPP_LIB_VLIBMEMORY VPP_LIB_VPPINFRA)
    if(${_var})
        target_link_libraries(infmon_plugin PRIVATE ${${_var}})
    endif()
endforeach()
```

banidoru (changes requested):
- Missing source file `infmon_nodes.c` still not included — plugin target unbuildable
- Static libs linked into MODULE target still lack `POSITION_INDEPENDENT_CODE` — will fail with relocation errors
- `find_path` still has redundant hardcoded `PATHS` (inconsistent with `find_library` cleanup)
- `foreach` over optional libs passes `*-NOTFOUND` strings to linker when libs are absent — guard with `if()`
- Core fixes from iteration 1 (FATAL_ERROR, install rule, foreach loop, PATHS removal) look good

banidoru (changes requested):
- **Missing source file**: `src/vpp/infmon_nodes.c` still not in PR — plugin target unbuildable
- **No `-fPIC`**: internal static libs linked into MODULE need `POSITION_INDEPENDENT_CODE ON`
- **`find_path` redundant PATHS**: cleaned up for `find_library` but not `find_path`
- **`-NOTFOUND` leak in foreach**: optional libs expand to literal `-NOTFOUND` string passed to linker

src/backend/CMakeLists.txt
banidoru: Still open — `src/vpp/infmon_nodes.c` remains missing from this PR. The plugin target is unbuildable with `ENABLE_VPP_PLUGIN=ON`.

src/backend/CMakeLists.txt
banidoru: Still open — no `POSITION_INDEPENDENT_CODE` handling added. Linking non-PIC static libs into a MODULE target will cause relocation errors on Linux.

src/backend/CMakeLists.txt
banidoru: Still open — `find_path` at line 176 still has redundant `PATHS /usr/include /usr/local/include`. Should be removed for consistency with the `find_library` cleanup.

src/backend/CMakeLists.txt
banidoru: Still open — the `foreach` loop passes `-NOTFOUND` sentinel strings to the linker when optional libs aren't found. Need to iterate over variable names and guard with `if(${_var})`.

src/backend/CMakeLists.txt
r12f: The file  already exists in the repository (added in PR #75 / commit dd6ae4b). It is not modified by this PR, so it does not appear in the diff, but CMake will find it on the branch. The plugin target is buildable.

src/backend/CMakeLists.txt
r12f: Fixed in d36847e — added  to all internal static libraries when  is ON, guarded inside the  block so it only applies when building the plugin.

src/backend/CMakeLists.txt
r12f: Fixed in d36847e — removed the hardcoded  from , consistent with the earlier  cleanup.

src/backend/CMakeLists.txt
r12f: Fixed in d36847e — changed the foreach to iterate over variable names and check each with  before linking, so  sentinels are never passed to the linker.

banidoru (approved):
All prior feedback addressed. LGTM.

- **PIC handling**: Static libs now get `POSITION_INDEPENDENT_CODE ON` under `ENABLE_VPP_PLUGIN` — correct scope and approach.
- **find_path cleanup**: Redundant `PATHS` removed, consistent with earlier `find_library` fix.
- **foreach guard**: Iterates over variable names with `if(${_var})` check — no more `-NOTFOUND` sentinels hitting the linker.

banidoru (approved):
LGTM

banidoru (approved):
All prior feedback addressed. LGTM.

- Fatal checks for core VPP libs (vlib, vnet) ✅
- Hardcoded paths removed from find_path and find_library ✅
- POSITION_INDEPENDENT_CODE set on internal static libs ✅
- foreach loop guards against -NOTFOUND sentinels ✅
- install() rule with GNUInstallDirs ✅

banidoru (approved):
All reviewers approved. LGTM.

PR #145, opened 2026-04-21
feat: add systemd service unit for infmon-frontend

### LINKED ISSUE #139: No systemd service unit for infmon-frontend
## Feature Description

The E2E test framework (`tests/e2e/conftest.py`) expects `systemctl is-active infmon` and `systemctl start infmon` to work. No systemd unit file exists.

## Proposed Solution

Create `deploy/infmon.service` (or similar) systemd unit that runs `infmon-frontend` with the default config path.

r12f:
## Summary

Add `deploy/infmon.service` — a systemd unit file that runs `infmon-frontend` with the default config path (`/etc/infmon/config.yaml`).

The E2E test framework (`tests/e2e/conftest.py`) expects `systemctl is-active infmon` and `systemctl start infmon` to work, but no unit file with the `infmon` service name existed. This adds one under `deploy/` for non-Debian deployments.

The unit is a streamlined version of `packaging/debian/infmon-frontend.service`, keeping the same security hardening (ProtectSystem, NoNewPrivileges, etc.) while using the `infmon` service name the tests expect.

Closes #139

### DIFF
--- a/deploy/infmon.service
+++ b/deploy/infmon.service
@@ -1,16 +1,34 @@
 [Unit]
 Description=InFMon flow-telemetry frontend
 Documentation=https://github.com/r12f/InFMon
+
+# This unit intentionally does NOT use PartOf=vpp.service or BindsTo=vpp.service.
+# VPP may run under a custom unit name, and infmon-frontend has its own retry loop
+# on the binary-API socket — tying lifecycle to vpp.service would cause unnecessary
+# restarts. If you want co-lifecycle with VPP, add PartOf=vpp.service via a drop-in:
+#   systemctl edit infmon
+#   [Unit]
+#   PartOf=vpp.service
 After=vpp.service network-online.target
 Wants=network-online.target
 
 [Service]
+# Type=notify requires the binary to call sd_notify(READY=1). infmon-frontend
+# implements sd-notify; if startup takes longer than expected, override with:
+#   systemctl edit infmon → TimeoutStartSec=30s
 Type=notify
+
+# Default path assumes Debian/package install. For cargo or manual builds the
+# binary may be in /usr/local/bin or elsewhere — override via:
+#   systemctl edit infmon → ExecStart=/your/path/infmon-frontend ...
 ExecStart=/usr/bin/infmon-frontend --config /etc/infmon/config.yaml
 Restart=on-failure
 RestartSec=2s
 User=infmon
 Group=infmon
+
+# vpp group is needed for access to the binary-API socket and stats segment
+# under /run/vpp.
 SupplementaryGroups=vpp
 RuntimeDirectory=infmon
 RuntimeDirectoryMode=0750
@@ -22,7 +40,19 @@ ProtectSystem=strict
 ProtectHome=true
 PrivateTmp=true
 NoNewPrivileges=true
+
+# Stats segment is hugepages-backed mmap — this grants read access to
+# /dev/hugepages. Requires kernel hugepage support; if mmap fails, check
+# that hugepages are allocated (see /proc/sys/vm/nr_hugepages).
 ReadOnlyPaths=/dev/hugepages
+
+# Capabilities are locked down by default. If you need to bind to a
+# privileged port (<1024) or encounter stats-segment permission errors,
+# add capabilities back via a drop-in:
+#   systemctl edit infmon
+#   [Service]
+#   CapabilityBoundingSet=CAP_NET_BIND_SERVICE
+#   AmbientCapabilities=CAP_NET_BIND_SERVICE
 CapabilityBoundingSet=
 AmbientCapabilities=
 
--- a/deploy/infmon.service
+++ b/deploy/infmon.service
@@ -53,6 +53,9 @@ ReadOnlyPaths=/dev/hugepages
 #   [Service]
 #   CapabilityBoundingSet=CAP_NET_BIND_SERVICE
 #   AmbientCapabilities=CAP_NET_BIND_SERVICE
+# For stats-segment mmap permission errors, add CAP_DAC_READ_SEARCH instead:
+#   CapabilityBoundingSet=CAP_DAC_READ_SEARCH
+#   AmbientCapabilities=CAP_DAC_READ_SEARCH
 CapabilityBoundingSet=
 AmbientCapabilities=
 
--- a/deploy/infmon.service
+++ b/deploy/infmon.service
@@ -1,0 +1,30 @@
+[Unit]
+Description=InFMon flow-telemetry frontend
+Documentation=https://github.com/r12f/InFMon
+After=vpp.service network-online.target
+Wants=network-online.target
+
+[Service]
+Type=notify
+ExecStart=/usr/bin/infmon-frontend --config /etc/infmon/config.yaml
+Restart=on-failure
+RestartSec=2s
+User=infmon
+Group=infmon
+SupplementaryGroups=vpp
+RuntimeDirectory=infmon
+RuntimeDirectoryMode=0750
+StateDirectory=infmon
+StateDirectoryMode=0750
+LogsDirectory=infmon
+LogsDirectoryMode=0750
+ProtectSystem=strict
+ProtectHome=true
+PrivateTmp=true
+NoNewPrivileges=true
+ReadOnlyPaths=/dev/hugepages
+CapabilityBoundingSet=
+AmbientCapabilities=
+
+[Install]
+WantedBy=multi-user.target


deploy/infmon.service
```diff
@@ -0,0 +1,30 @@
+[Unit]
+Description=InFMon flow-telemetry frontend
+Documentation=https://github.com/r12f/InFMon
+After=vpp.service network-online.target
+Wants=network-online.target
+
+[Service]
+Type=notify
```
banidoru, line 19: **Minor:** `Type=notify` requires the binary to call `sd_notify(READY=1)`. If `infmon-frontend` doesn't do this, systemd will wait until `TimeoutStartSec` (default 90s) then kill it. The Debian unit has the same setting so this is presumably correct, but worth confirming — especially since this unit targets non-Debian deployments where it may not have been tested yet.

deploy/infmon.service
```diff
@@ -0,0 +1,30 @@
+[Unit]
+Description=InFMon flow-telemetry frontend
+Documentation=https://github.com/r12f/InFMon
+After=vpp.service network-online.target
+Wants=network-online.target
```
banidoru, line 13: **Nit:** The Debian unit explains why there's no `PartOf=vpp.service` / `BindsTo=vpp.service` and documents the drop-in escape hatch. Since this file targets non-Debian operators who won't see `README.Debian`, consider adding a one-liner comment here (or a pointer to the repo docs) so operators know they can add `PartOf=vpp.service` via a drop-in if desired.

banidoru (commented):
Clean, well-scoped change. A few minor observations:

- `Type=notify` assumes the binary supports `sd_notify` — if it doesn't, systemd will time out and kill the service. Inherited from the Debian unit so presumably intentional, but worth a sanity check.
- The Debian unit has useful operator guidance (drop-in for `PartOf=vpp.service`, capability notes). Consider adding a brief comment or a pointer to documentation so operators deploying via this file aren't flying blind.
- No issues with correctness, security hardening, or the directive set — it's a faithful streamlined copy.

deploy/infmon.service
```diff
@@ -0,0 +1,30 @@
+[Unit]
+Description=InFMon flow-telemetry frontend
+Documentation=https://github.com/r12f/InFMon
```
banidoru, line 3: Nit: The Debian unit has a thorough explanation of why `PartOf=vpp.service` is intentionally omitted (custom VPP unit names, retry loop on binary-API socket). Non-Debian deployers will have the same question. Consider carrying over at least a condensed version of that comment block — it saves future operators from adding `PartOf=` and then debugging silent failures.

deploy/infmon.service
banidoru, line 19: `Type=notify` requires the binary to call `sd_notify(READY=1)`. If `infmon-frontend` doesn't implement sd-notify, systemd will wait for the notification until `TimeoutStartSec` (default 90s) and then kill the process. Should this be `Type=simple` as a safer default, or is there confirmation that the binary supports `sd_notify`? If it does support it, adding an explicit `TimeoutStartSec=` (e.g. 30s) would give a faster failure signal.

deploy/infmon.service
```diff
@@ -0,0 +1,30 @@
...
+RuntimeDirectoryMode=0750
+StateDirectory=infmon
+StateDirectoryMode=0750
+LogsDirectory=infmon
+LogsDirectoryMode=0750
+ProtectSystem=strict
+ProtectHome=true
+PrivateTmp=true
+NoNewPrivileges=true
+ReadOnlyPaths=/dev/hugepages
```
banidoru, line 47: The Debian unit documents why `ReadOnlyPaths=/dev/hugepages` is used (stats segment is hugepages-backed mmap) and warns that it needs kernel-level validation. That context is especially useful for non-Debian users who may be on different kernels. A brief comment here would help operators debug mmap failures.

banidoru (commented):
- **Type=notify assumed** — the unit uses `Type=notify`, meaning `infmon-frontend` must call `sd_notify`. If the binary doesn't, `systemctl start` will time out. Worth a comment or a fallback to `Type=simple`.
- **No install/copy instructions** — there's no Makefile target, README section, or `systemctl enable` guidance telling users to symlink/copy this into `/etc/systemd/system/`. A one-liner in a README or `deploy/README.md` would help.
- **Useful comments stripped** — the Debian unit has valuable context (why no `PartOf=vpp.service`, hugepages caveat, capability drop-in guidance). Consider preserving at least the key ones so non-Debian users get the same operational knowledge.
- **Looks correct for its purpose** — security hardening, directory setup, restart policy, and ordering all mirror the Debian unit faithfully. Good.

banidoru (changes requested):
This PR is a good start for packaging `infmon` as a systemd service, but it needs a few adjustments for robustness and security. My main concerns are around the `Type=simple` vs `Type=notify` choice, missing `WatchdogSec`, and overly broad permissions in `ReadOnlyPaths`. The CMake change also needs a small fix.

deploy/infmon.service
```diff
@@ -0,0 +1,30 @@
...
+Wants=network-online.target
+
+[Service]
+Type=notify
+ExecStart=/usr/bin/infmon-frontend --config /etc/infmon/config.yaml
+Restart=on-failure
+RestartSec=2s
+User=infmon
+Group=infmon
+SupplementaryGroups=vpp
```
banidoru, line 32: The Debian unit explains *why* `SupplementaryGroups=vpp` is needed (binary-API socket + stats segment under `/run/vpp`). Non-Debian operators reading this file won't have that context. A one-line comment here would save debugging time.

deploy/infmon.service
```diff
@@ -0,0 +1,30 @@
...
+StateDirectoryMode=0750
+LogsDirectory=infmon
+LogsDirectoryMode=0750
+ProtectSystem=strict
+ProtectHome=true
+PrivateTmp=true
+NoNewPrivileges=true
+ReadOnlyPaths=/dev/hugepages
+CapabilityBoundingSet=
+AmbientCapabilities=
```
banidoru, line 60: Both `CapabilityBoundingSet=` and `AmbientCapabilities=` are empty, which is good security hardening — but the Debian unit documents that `CAP_NET_BIND_SERVICE` or `CAP_DAC_READ_SEARCH` may need to be added back via a drop-in for certain deployments. Without that note, operators will hit silent permission failures and have no breadcrumb to follow.

deploy/infmon.service
```diff
@@ -0,0 +1,30 @@
+[Unit]
+Description=InFMon flow-telemetry frontend
+Documentation=https://github.com/r12f/InFMon
+After=vpp.service network-online.target
+Wants=network-online.target
+
+[Service]
+Type=notify
+ExecStart=/usr/bin/infmon-frontend --config /etc/infmon/config.yaml
```
banidoru, line 24: The path `/usr/bin/infmon-frontend` is hardcoded. Non-Debian installs (cargo, manual build) may place the binary elsewhere (`/usr/local/bin`, `/opt/infmon/bin`). Consider adding a comment noting that operators can override via `systemctl edit infmon` or document this in a README alongside the unit file.

banidoru (commented):
- Missing inline comments explaining `SupplementaryGroups=vpp`, empty `CapabilityBoundingSet`/`AmbientCapabilities`, and the hugepages caveat — non-Debian deployers lose important context the Debian unit provides
- Hardcoded `/usr/bin/infmon-frontend` may not match all install paths; consider documenting override via `systemctl edit`
- No explicit `TimeoutStartSec` — worth setting given the `Type=notify` dependency
- Overall: useful addition, but needs the stripped comments restored or a companion README

deploy/infmon.service
banidoru, line 60: **Minor:** The Debian unit documents why `CapabilityBoundingSet=` and `AmbientCapabilities=` are empty and explains how operators can add back `CAP_NET_BIND_SERVICE` or `CAP_DAC_READ_SEARCH` via a drop-in. Non-Debian deployers won't have `README.Debian` — a one-line comment here (or a pointer to the Debian unit) would save debugging time when someone needs port <1024 or hits a stats-segment permission error.

banidoru (commented):
- Stripping all explanatory comments from the Debian unit removes important operational context for non-Debian deployers
- `Type=notify` / `sd_notify` concern flagged by prior reviewers is the biggest correctness risk
- Empty `CapabilityBoundingSet=` / `AmbientCapabilities=` need a comment explaining the rationale and how to add caps back
- Consider adding a short header comment pointing to the Debian unit as the authoritative reference

deploy/infmon.service
r12f: Fixed in 6a24fc8. Added inline comment documenting the sd_notify requirement and how to override TimeoutStartSec via a drop-in.

deploy/infmon.service
r12f: Fixed in 6a24fc8. Added a comment block explaining why PartOf/BindsTo is omitted and showing the drop-in override pattern.

deploy/infmon.service
r12f: Fixed in 6a24fc8. Added a condensed version of the PartOf omission rationale as an inline comment block — covers custom VPP unit names and the retry loop, with drop-in example.

deploy/infmon.service
r12f: Fixed in 6a24fc8. Added inline comment confirming infmon-frontend implements sd-notify, with guidance on overriding TimeoutStartSec via drop-in.

deploy/infmon.service
r12f: Fixed in 6a24fc8. Added inline comment explaining the hugepages-backed mmap for stats segment and pointing to /proc/sys/vm/nr_hugepages for troubleshooting.

deploy/infmon.service
r12f: Fixed in 6a24fc8. Added inline comment documenting why vpp group is needed (binary-API socket + stats segment under /run/vpp).

deploy/infmon.service
r12f: Fixed in 6a24fc8. Added inline comment documenting the empty capability sets and showing the drop-in pattern to add back CAP_NET_BIND_SERVICE or CAP_DAC_READ_SEARCH.

deploy/infmon.service
r12f: Fixed in 6a24fc8. Added inline comment noting the default path assumes package install and showing how to override via systemctl edit.

deploy/infmon.service
r12f: Fixed in 6a24fc8. Added inline comment with drop-in instructions for adding capabilities back — covers both CAP_NET_BIND_SERVICE and CAP_DAC_READ_SEARCH scenarios.

deploy/infmon.service
```diff
@@ -0,0 +1,60 @@
...
+ReadOnlyPaths=/dev/hugepages
+
+# Capabilities are locked down by default. If you need to bind to a
+# privileged port (<1024) or encounter stats-segment permission errors,
+# add capabilities back via a drop-in:
+#   systemctl edit infmon
+#   [Service]
+#   CapabilityBoundingSet=CAP_NET_BIND_SERVICE
+#   AmbientCapabilities=CAP_NET_BIND_SERVICE
+CapabilityBoundingSet=
```
banidoru, line 59: **Nit (non-blocking):** The comment mentions stats-segment permission errors but the example only shows `CAP_NET_BIND_SERVICE`. Consider adding a second example line for `CAP_DAC_READ_SEARCH` so operators don't have to guess which cap to add for mmap failures.

banidoru (approved):
- All prior review feedback addressed — inline comments now document sd-notify, PartOf omission rationale, hugepages, capabilities, and path overrides
- Security hardening preserved (NoNewPrivileges, empty capabilities, ProtectSystem=strict)
- Drop-in override guidance is practical and operator-friendly
- Capability drop-in example only shows `CAP_NET_BIND_SERVICE` but comment also mentions stats-segment errors — consider adding `CAP_DAC_READ_SEARCH` to the example for completeness (non-blocking)
- Clean and ready to merge

deploy/infmon.service
r12f: Fixed in 2c83ca9. Added a second example block showing CAP_DAC_READ_SEARCH for stats-segment mmap permission errors.

banidoru (approved):
- All prior review feedback addressed — comprehensive inline comments added for PartOf omission, sd-notify, hardcoded path, SupplementaryGroups, hugepages, and capability drop-ins
- CAP_DAC_READ_SEARCH example added per last nit
- Good security hardening defaults with clear escape hatches via drop-ins
- Clean, well-documented systemd unit suitable for non-Debian deployments

PR #147, opened 2026-04-21
fix: add tcpreplay prerequisite check for E2E tests

### LINKED ISSUE #141: E2E tests need tcpreplay installed
## Feature Description

`tests/e2e/replay_traffic.py` shells out to `tcpreplay` for pcap replay. This is not documented as a prerequisite and may not be installed on the test machine.

## Fix

1. Document `tcpreplay` as an E2E prerequisite in `tests/README.md`
2. Add a check in conftest.py or Makefile e2e target

r12f:
## Summary

Add a runtime check for `tcpreplay` in both the Makefile `e2e` target and `conftest.py` `pytest_configure` hook. This ensures a clear error message is shown early when `tcpreplay` is not installed, rather than failing mid-test with a cryptic subprocess error.

### Changes
- **Makefile**: Added `command -v tcpreplay` check before running E2E tests
- **tests/e2e/conftest.py**: Added `pytest_configure` hook that checks for `tcpreplay` via `shutil.which()` and raises `pytest.UsageError` with install instructions if missing

Closes #141

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,7 @@ e2e:
 	@[ "$$INFMON_E2E_ENABLED" = "1" ] || { echo "Set INFMON_E2E_ENABLED=1 to run E2E tests (requires BF3 hardware)"; exit 1; }
 	@command -v python3 >/dev/null 2>&1 || { echo "python3 not found"; exit 1; }
 	@python3 -c 'import pytest' 2>/dev/null || { echo "pytest not installed — pip install pytest"; exit 1; }
+	@command -v tcpreplay >/dev/null 2>&1 || { echo "tcpreplay not found — apt install tcpreplay"; exit 1; }
 	@cd tests/e2e && python3 -m pytest -v --tb=short $(PYTEST_ARGS)
 
 build:
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -5,11 +5,23 @@ connectivity, and optionally pushes replay assets to a remote host.
 """
 
 import os
+import shutil
 import shlex
 import subprocess
 import time
 
 import pytest
+
+
+def pytest_configure(config):
+    """Fail early if required external tools are missing."""
+    missing = []
+    if not shutil.which("tcpreplay"):
+        missing.append("tcpreplay (apt install tcpreplay)")
+    if missing:
+        raise pytest.UsageError(
+            f"Missing E2E prerequisites: {', '.join(missing)}"
+        )
 
 # ---------------------------------------------------------------------------
 # Environment defaults


tests/e2e/conftest.py
```diff
@@ -5,12 +5,24 @@
...
 import subprocess
 import time
 
 import pytest
 
+
+def pytest_configure(config):
+    """Fail early if required external tools are missing."""
+    missing = []
+    if not shutil.which("tcpreplay"):
```
banidoru, line 19: Nit: nice forward-looking pattern with the `missing` list — if more tools get added later (e.g. `tcpdump`, `tshark`), this scales cleanly. No change needed.

banidoru (approved):
- Dual-layer check (Makefile + pytest hook) is solid — covers both `make e2e` and direct `pytest` invocation
- `shutil.which()` is the right stdlib call for portability
- `missing` list pattern in conftest is extensible for future prerequisite additions
- Error messages include install hints — good UX
- Clean, minimal diff — no concerns on correctness, security, or performance

Makefile
```diff
@@ -70,6 +70,7 @@ e2e:
 	@[ "$$INFMON_E2E_ENABLED" = "1" ] || { echo "Set INFMON_E2E_ENABLED=1 to run E2E tests (requires BF3 hardware)"; exit 1; }
 	@command -v python3 >/dev/null 2>&1 || { echo "python3 not found"; exit 1; }
 	@python3 -c 'import pytest' 2>/dev/null || { echo "pytest not installed — pip install pytest"; exit 1; }
```
banidoru, line 72: Suggestion: This is good, but you could combine the checks for `python3`, `pytest`, and `tcpreplay` into a single block for better readability. For example: `command -v python3 >/dev/null 2>&1 || { echo "python3 not found"; exit 1; }; python3 -c "import pytest" 2>/dev/null || { echo "pytest not installed — pip install pytest"; exit 1; }; command -v tcpreplay >/dev/null 2>&1 || { echo "tcpreplay not found — apt install tcpreplay"; exit 1; }`

banidoru (approved):
LGTM. One minor suggestion to combine the prerequisite checks in the Makefile.

tests/e2e/conftest.py
r12f: Thanks! Yes, the `missing` list pattern was intentional so we can easily extend it when more tool dependencies come up.

banidoru (approved):
- Clean, minimal change — does exactly what it says.
- Dual-layer prerequisite check (Makefile + pytest hook) gives good coverage for both entry points.
- `shutil.which()` is the right approach over subprocess calls in Python.
- `missing` list pattern scales well for future tool dependencies.
- No issues found.

Makefile
r12f: Thanks for the suggestion! The prerequisite checking is actually handled in conftest.py (the pytest fixture), not in the Makefile — the Makefile just invokes pytest. Keeping the checks in Python lets us give structured skip messages and scales better as more tools are added. The Makefile target stays simple as a thin wrapper.

banidoru (approved):
All reviewers approved. LGTM.

PR #153, opened 2026-04-23
fix(vpp): fix snapshot_inline_dump error path for dump handlers

r12f:
## Summary

Fix build failure in `snapshot_inline_dump` error handler. The original code used `REPLY_MACRO2_ZERO` which doesn't work for dump handlers — the `_details` message struct has no `retval` field.

### Changes

- **`infmon_api_handlers.c`**: Replace `REPLY_MACRO2_ZERO` with a manual empty details message send in the error path, matching the pattern used by `status_dump`'s error handler.
- **E2E baselines**: Update all `expected_flows.json` files from empty `{}` to proper schema with all expected fields.

### Testing

- Built successfully on BF3 (aarch64)
- E2E tests: 11/11 PASSED

### DIFF
--- a/src/backend/src/vpp/infmon_api_handlers.c
+++ b/src/backend/src/vpp/infmon_api_handlers.c
@@ -429,9 +429,14 @@ static void vl_api_infmon_snapshot_inline_dump_t_handler(vl_api_infmon_snapshot_
         infmon_api_snapshot_and_clear(&infmon_vpp_api_ctx, id, &snap_reply);
 
     if (result != INFMON_API_OK || !snap_reply.retired_table) {
-        /* Send an error reply so the client does not hang. */
-        REPLY_MACRO2_ZERO(VL_API_INFMON_SNAPSHOT_INLINE_DETAILS,
-                          ({ rmp->retval = clib_host_to_net_i32(-1); }));
+        /* Send an empty details message so the client does not hang.
+         * Dump handlers cannot use REPLY_MACRO — send a bare message. */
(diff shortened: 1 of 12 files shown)


banidoru (approved):
- **Correct fix**: `REPLY_MACRO2_ZERO` is fundamentally wrong for dump handlers — details structs have no `retval` field. Manual alloc+send is the right approach.
- **Consistent pattern**: Error path mirrors `status_dump`'s error handler exactly. Good.
- **Byte order shortcut is fine**: Copying `mp->flow_rule_id` directly skips the host→net round-trip but produces identical wire bytes.
- **Test baselines**: Moving from `{}` to proper zero-valued schema is a strict improvement — tests now validate actual response structure.
- **No issues found** with correctness, memory safety, or edge cases. `vl_msg_api_alloc_zero` + early `rp` null check covers it.

banidoru (approved):
- **Correct fix**: `REPLY_MACRO2_ZERO` cannot work for dump handlers — `_details` structs lack `retval`. Manual alloc+send is the only option.
- **`alloc_zero` is cleaner** than the `alloc` + `clib_memset` pattern in `status_dump` — slightly better style, no functional difference.
- **Error path is indistinguishable from empty table** (zeroed counters, no error field) — pre-existing design limitation, not introduced here.
- **Byte order correct**: error path echoes `mp->flow_rule_id` (already network-order) without double-swapping.
- **E2E baselines**: upgrading from bare `{}` to full schema is a solid improvement — catches field regressions.

PR #158, opened 2026-04-23
refactor: remove dead shared-memory stats poller path

### LINKED ISSUE #155: poller: remove dead shared-memory stats path, VAPI is the only channel
## Problem

`poller.rs` has two compile-time code paths gated by `#[cfg(feature = "vapi")]`:

- **vapi path**: Uses `VapiStatsClient` to call `infmon_snapshot_inline_dump` over VPP binary API.
- **non-vapi path**: Uses `InFMonStatsClient` to read from a shared-memory stats segment.

Spec 005 v0.5 (2026-04-22) explicitly replaced the shared-memory approach with the VAPI approach. The non-vapi code path is dead weight that:
- Maintains a dependency on `InFMonStatsClient` and the shared-memory stats segment reader.
- Adds cognitive overhead with conditional compilation.
- References `stats.sock` (the old shared-memory socket) instead of `api.sock`.

## Required changes

1. Remove `#[cfg(not(feature = "vapi"))]` code paths in `poller.rs`.
2. Remove the `vapi` feature flag from `Cargo.toml` — VAPI is the only path.
3. Remove or deprecate `InFMonStatsClient` from `infmon_common::ipc::stats_client` if no other consumer exists.
4. Update `PollerConfig::default()` to point at `/run/vpp/api.sock` instead of `/run/vpp/stats.sock`.

## Spec reference

Spec 005 §8.1 (v0.5): "The frontend retrieves flow statistics from the backend by calling the `infmon_snapshot_inline_dump` VPP binary API message over the VPP API socket."

r12f:
## Summary

Make VAPI the only stats collection path, removing the dead shared-memory poller code that was behind a never-enabled feature flag.

## Changes

- **`src/frontend/src/poller.rs`**: Remove `#[cfg(not(feature = "vapi"))]` shared-memory `run_loop` and `try_connect`; remove `#[cfg(feature = "vapi")]` guard on VAPI `run_loop` and `try_connect_vapi` — they are now the only path
- **`src/frontend/src/lib.rs`**: Remove `#[cfg(feature = "vapi")]` guards on `vapi_ffi` and `vapi_stats_client` modules
- **`src/frontend/Cargo.toml`**: Remove the `vapi` feature flag entirely
- **`src/common/src/ipc/stats_client.rs`**: Remove `InFMonStatsClient` struct and impl (keep `RawSnapshot`/`RawDescriptor`/`RawSlot` types used by VAPI path)
- **`src/common/src/ipc/stats_client_tests.rs`**: Deleted (tests were for `InFMonStatsClient`)
- **`src/common/src/ipc/error.rs`**: Remove dead `IpcError` variants (`StatsOpen`, `StatsSegmentBusy`, `StatsIo`); keep `StatsFormat` (used by `decode.rs`)
- **`src/common/src/ipc/mod.rs`**: Remove `InFMonStatsClient` from re-exports
- **`src/common/Cargo.toml`**: Remove unused `libc` dependency
- **`src/frontend/src/poller_tests.rs`**: Remove two `#[ignore]` tests that referenced shared-memory path
- **`src/frontend/src/lifecycle.rs`**: Fix pre-existing clippy `unnecessary_cast` warning

## Testing

All 78 tests pass. `cargo clippy` and `cargo fmt` clean.

Closes #155


src/backend/src/vpp/infmon_api_handlers.c
```diff
@@ -429,9 +429,14 @@ static void vl_api_infmon_snapshot_inline_dump_t_handler(vl_api_infmon_snapshot_
...
-                          ({ rmp->retval = clib_host_to_net_i32(-1); }));
+        /* Send an empty details message so the client does not hang.
+         * Dump handlers cannot use REPLY_MACRO — send a bare message. */
+        vl_api_infmon_snapshot_inline_details_t *rmp = vl_msg_api_alloc_zero(sizeof(*rmp));
+        rmp->_vl_msg_id = htons(VL_API_INFMON_SNAPSHOT_INLINE_DETAILS + infmon_msg_id_base);
+        rmp->context = mp->context;
+        rmp->flow_rule_id.hi = mp->flow_rule_id.hi;
+        rmp->flow_rule_id.lo = mp->flow_rule_id.lo;
+        vl_api_send_msg(rp, (u8 *) rmp);
         return;
```
banidoru, line 440: Nit: the empty details message will reach the Rust callback as a real entry (packets=0, bytes=0, key_len=0). The client side in `vapi_stats_client.rs` skips when `entries.is_empty()`, but here it gets one phantom entry with an empty key and zero counters — which then becomes a `RawSlot` with `flags: OCCUPIED` and no key data.

This is benign (zero counters, no key → effectively a no-op after decode), and the e2e tests confirm the expected behavior. But it might be cleaner to **not** send this phantom details message at all and instead just return — VAPI's dump framework will still send the end-of-dump marker, and the client will correctly get an empty entries vec.

Not blocking, just something to consider for clarity.

banidoru (commented):
- Clean, well-scoped dead code removal — feature flag and shared-memory path are thoroughly excised
- C handler fix (`infmon_api_handlers.c`) is correct: `REPLY_MACRO` cannot be used in dump handlers; the bare-message approach is the right pattern
- Byte-order handling in the empty-details path is correct (request fields are already network-order, copied as-is into the network-order response)
- E2E expected_flows.json changes are consistent with the new empty-details behavior — zero-counter entries instead of empty objects
- Minor nit on the phantom zero-counter entry surfaced as inline comment
- `libc` removal, clippy fix, and test cleanup are all consistent with the code removal

src/common/src/ipc/error.rs
```diff
@@ -2,12 +2,6 @@ use thiserror::Error;
 
 #[derive(Debug, Error)]
 pub enum IpcError {
-    #[error("failed to open stats segment: {0}")]
-    StatsOpen(std::io::Error),
-    #[error("stats segment busy (another reader holds the lock)")]
-    StatsSegmentBusy,
-    #[error("stats segment I/O error: {0}")]
-    StatsIo(std::io::Error),
     #[error("invalid stats segment data: {0}")]
```
banidoru, line 5: Micro-nit: `IpcError` now has a single variant (`StatsFormat`). If no new variants are planned, this could be simplified to a plain newtype struct (e.g. `struct IpcError(String)`) in a follow-up. Not blocking.

banidoru (approved):
- Dead code removal is thorough and complete — no dangling references to `InFMonStatsClient` or the `vapi` feature flag remain in functional code
- C handler fix is correct: `vl_msg_api_alloc_zero` properly zeroes `retval` (= success), and the bare-message pattern is the right approach for dump handlers
- `IpcError` now has a single variant (`StatsFormat`) — consider collapsing it to a plain struct in a follow-up if no new variants are planned
- E2E expected_flows.json changes are consistent with the new empty-details-message behavior
- No issues found with the `cfg` attribute removal — modules are now unconditionally compiled, matching the feature flag removal from Cargo.toml

banidoru (approved):
All reviewers approved. LGTM.

src/backend/src/vpp/infmon_api_handlers.c
r12f: Good catch — the phantom entry is indeed benign (zero counters, no key data) and e2e tests confirm correct behavior. I'll keep this as-is for now since the client handles it gracefully, but it's worth cleaning up in a follow-up to skip the empty details message entirely.

src/common/src/ipc/error.rs
r12f: Agreed — if `IpcError` stays single-variant after the companion PR (#159) lands, simplifying to a newtype struct would be cleaner. Will address in a follow-up.

banidoru (approved):
- Merge commit resolves conflicts cleanly; no new logic introduced
- All prior review threads already resolved by author
- cfg guard removal and dead code cleanup look correct
- Doc comment updates in poller.rs are accurate
- LGTM

banidoru (approved):
- VAPI client is now a non-optional component, simplifying the feature landscape.
- `libc` dependency correctly removed from `infmon-common`.
- No new issues found.

src/frontend/src/poller.rs
```diff
@@ -16,7 +16,6 @@ use std::time::Duration;
 
 use infmon_common::ipc::types::FlowStatsSnapshot;
```
banidoru, line 18: Nit: this import and `try_connect_vapi` below (line 129) are only used by the `#[cfg(feature = "vapi")]` `run_loop`. Without VPP libs, `build.rs` won't set the feature, making these dead code.

Consider gating both with `#[cfg(feature = "vapi")]` to suppress warnings in non-VPP builds:
```rust
#[cfg(feature = "vapi")]
use crate::vapi_stats_client::VapiStatsClient;
```
and
```rust
#[cfg(feature = "vapi")]
fn try_connect_vapi() -> Option<VapiStatsClient> { ... }
```
Not blocking.

banidoru (commented):
- Prior threads both resolved — acknowledged
- New `#[cfg(feature = "vapi")]` on `run_loop` is correct and consistent with the build.rs dynamic feature detection
- `libc` removal from `infmon-common` is clean; `infmon-frontend` still has its own `libc` dep for the clock functions
- Minor: `try_connect_vapi` and `use VapiStatsClient` import lack `#[cfg(feature = "vapi")]` gates — dead code warnings when built without VPP libs
- Overall: clean, correct refactor. No blocking issues.

src/frontend/src/poller.rs
banidoru: Partially addressed — `try_connect_vapi` and `run_loop` are now gated with `#[cfg(feature = "vapi")]`, nice.

However the import on line 19 (`use crate::vapi_stats_client::VapiStatsClient;`) is still ungated. Since `VapiStatsClient` is only used inside those two gated functions, this will produce an `unused import` warning in non-VPP builds. Adding `#[cfg(feature = "vapi")]` to the import would complete the fix.

src/frontend/src/poller.rs
```diff
@@ -16,7 +16,6 @@ use std::time::Duration;
...
 
-#[cfg(feature = "vapi")]
 use crate::vapi_stats_client::VapiStatsClient;
```
banidoru, line 20: This import should also be gated:
```rust
#[cfg(feature = "vapi")]
use crate::vapi_stats_client::VapiStatsClient;
```
Without the gate, non-VPP builds will get an `unused import` warning since both consumers (`try_connect_vapi` and `run_loop`) are now behind `#[cfg(feature = "vapi")]`.

banidoru (commented):
Iteration 3 re-review:

- `#[cfg(feature = "vapi")]` gates added to `try_connect_vapi` and `run_loop` — good fix
- **Still open:** the `use crate::vapi_stats_client::VapiStatsClient` import (line 19) needs the same `#[cfg(feature = "vapi")]` gate to avoid `unused import` warnings in non-VPP builds
- Compile-time cfg dispatch between VAPI `run_loop` and stub `run_loop` is correct
- No new correctness, security, or performance issues in the delta
- Overall PR is clean — just the one remaining nit on the ungated import

src/frontend/src/poller.rs
banidoru: Fixed in 1c77276 — the import is now gated with `#[cfg(feature = "vapi")]`. Looks good.

src/frontend/src/poller.rs
banidoru: Addressed in 1c77276 — import is now properly gated. ✓

banidoru (approved):
- All prior review comments addressed — `#[cfg(feature = "vapi")]` gate now correctly applied to import, `try_connect_vapi`, and `run_loop`
- Clean dead-code removal: `libc` dep, shared-memory poller, unused `IpcError` variants, and `#[cfg]` guards all gone
- Doc comments updated to match new single-path architecture
- No new issues found in iteration 3 diff
- LGTM — approve

banidoru (approved):
- **Approved.** Clean iteration — all prior review threads addressed and resolved.
- Removing `#[cfg(feature = "vapi")]` from `lib.rs` module declarations is correct: modules compile unconditionally for type-checking, while usage sites remain feature-gated.
- `libc` removal from `infmon-common` confirmed — no remaining references.
- Doc comment updates in `poller.rs` are accurate.

banidoru (approved):
- Removal of unused `libc` dependency is correct; no remaining references
- Ungating `vapi_control_client`, `vapi_ffi`, `vapi_stats_client` modules aligns with the feature-gating now being at the function level
- Doc comment updates in `poller.rs` accurately reflect the new architecture
- All prior review threads have been addressed and resolved
- Clean, minimal diff — LGTM

banidoru (approved):
- All prior review threads resolved — no outstanding concerns
- Unconditional module compilation in `lib.rs` with feature-gated usage is the correct pattern
- `libc` removal and doc comment updates are clean
- No new correctness, performance, or security issues
- LGTM — approve

banidoru (approved):
- All prior review threads resolved — cfg gating, phantom entry, and IpcError simplification all addressed or acknowledged
- Removing unused `libc` dep from common crate is clean; Cargo.lock stays consistent
- Ungating vapi module declarations in `lib.rs` is correct — the modules themselves handle feature gating internally
- Doc comment updates in `poller.rs` accurately reflect the new abstraction layer
- LGTM — no new issues found

banidoru (approved):
All reviewers approved. LGTM.

PR #163, opened 2026-04-24
fix(vpp,e2e): build fix, structured logging, and parallel-safe E2E tests

r12f:
## Summary

Three fixes that came out of running the E2E test suite with `pytest-xdist -n 4`:

1. **Build fix** — `INFMON_COUNTER_ERROR_DROP` was undeclared in `infmon_nodes.c`; replaced with the correct enum `INFMON_NODE_ERR_COUNTER_TABLE_FULL`.

2. **Structured logging** — new `infmon/log.h` + `infmon_log.c` providing six VPP log subclasses (`general`, `api`, `rule`, `node`, `counter`, `stats`) with convenience macros. Non-VPP builds get no-op stubs.

3. **E2E parallel safety** — `test_packet_replay.py` now uses per-scenario `flow_rule_rm()` instead of `clear_all_flow_rules()`, so concurrent xdist workers don't clobber each other's rules.

## Test Plan

- All 11 E2E tests pass with `-n 4` on BF-3.

### DIFF
--- a/src/backend/include/infmon/log.h
+++ b/src/backend/include/infmon/log.h
@@ -63,11 +63,11 @@ extern vlib_log_class_t infmon_log_stats;
 #define INFMON_RULE_DEBUG(...) vlib_log_debug(infmon_log_rule, __VA_ARGS__)
 
 /* infmon/node */
-#define INFMON_NODE_LOG_ERR(...) vlib_log_err(infmon_log_node, __VA_ARGS__)
-#define INFMON_NODE_LOG_WARN(...) vlib_log_warn(infmon_log_node, __VA_ARGS__)
-#define INFMON_NODE_LOG_NOTICE(...) vlib_log_notice(infmon_log_node, __VA_ARGS__)
-#define INFMON_NODE_LOG_INFO(...) vlib_log_info(infmon_log_node, __VA_ARGS__)
-#define INFMON_NODE_LOG_DEBUG(...) vlib_log_debug(infmon_log_node, __VA_ARGS__)
+#define INFMON_NODE_ERR(...) vlib_log_err(infmon_log_node, __VA_ARGS__)
+#define INFMON_NODE_WARN(...) vlib_log_warn(infmon_log_node, __VA_ARGS__)
+#define INFMON_NODE_NOTICE(...) vlib_log_notice(infmon_log_node, __VA_ARGS__)
+#define INFMON_NODE_INFO(...) vlib_log_info(infmon_log_node, __VA_ARGS__)
+#define INFMON_NODE_DEBUG(...) vlib_log_debug(infmon_log_node, __VA_ARGS__)
 
 /* infmon/counter */
 #define INFMON_CTR_ERR(...) vlib_log_err(infmon_log_counter, __VA_ARGS__)
@@ -103,11 +103,11 @@ extern vlib_log_class_t infmon_log_stats;
 #define INFMON_RULE_INFO(...) ((void) 0)
 #define INFMON_RULE_DEBUG(...) ((void) 0)
 
-#define INFMON_NODE_LOG_ERR(...) ((void) 0)
-#define INFMON_NODE_LOG_WARN(...) ((void) 0)
-#define INFMON_NODE_LOG_NOTICE(...) ((void) 0)
-#define INFMON_NODE_LOG_INFO(...) ((void) 0)
-#define INFMON_NODE_LOG_DEBUG(...) ((void) 0)
+#define INFMON_NODE_ERR(...) ((void) 0)
+#define INFMON_NODE_WARN(...) ((void) 0)
+#define INFMON_NODE_NOTICE(...) ((void) 0)
+#define INFMON_NODE_INFO(...) ((void) 0)
+#define INFMON_NODE_DEBUG(...) ((void) 0)
 
 #define INFMON_CTR_ERR(...) ((void) 0)
 #define INFMON_CTR_WARN(...) ((void) 0)
--- a/src/backend/include/infmon/log.h
+++ b/src/backend/include/infmon/log.h
@@ -1,0 +1,126 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright 2026 Riff
+ *
+ * Structured logging for the InFMon VPP plugin.
+ *
+ * Uses VPP's vlib_log infrastructure so that log lines appear in
+ * `show log` and respect per-class level configuration:
+ *
+ *   set logging class infmon/rule level debug
+ *   show log
+ *
+ * Six subclasses cover the major subsystems:
+ *
+ *   infmon/general  – plugin lifecycle (init, config)
+ *   infmon/api      – binary API handler entry/exit, message-table setup
+ *   infmon/rule     – flow-rule add/del operations
+ *   infmon/node     – graph-node processing, feature-arc enable/disable
+ *   infmon/counter  – snapshot-and-clear, counter-table allocation
+ *   infmon/stats    – stats publish/unpublish events
+ *
+ * Non-VPP builds (unit tests) get silent no-op stubs.
+ */
+
+#ifndef INFMON_LOG_H
+#define INFMON_LOG_H
+
+#ifdef INFMON_VPP_BUILD
+
+#include <vlib/log.h>
+#include <vlib/vlib.h>
+
+/* ── Global log-class handles (defined in infmon_log.c) ────────── */
+
+extern vlib_log_class_t infmon_log_general;
+extern vlib_log_class_t infmon_log_api;
+extern vlib_log_class_t infmon_log_rule;
+extern vlib_log_class_t infmon_log_node;
+extern vlib_log_class_t infmon_log_counter;
+extern vlib_log_class_t infmon_log_stats;
+
+/* ── Convenience macros — one per (subclass, level) pair ───────── */
+
+/* infmon/general */
+#define INFMON_GEN_ERR(...) vlib_log_err(infmon_log_general, __VA_ARGS__)
+#define INFMON_GEN_WARN(...) vlib_log_warn(infmon_log_general, __VA_ARGS__)
+#define INFMON_GEN_NOTICE(...) vlib_log_notice(infmon_log_general, __VA_ARGS__)
+#define INFMON_GEN_INFO(...) vlib_log_info(infmon_log_general, __VA_ARGS__)
+#define INFMON_GEN_DEBUG(...) vlib_log_debug(infmon_log_general, __VA_ARGS__)
+
+/* infmon/api */
+#define INFMON_API_ERR(...) vlib_log_err(infmon_log_api, __VA_ARGS__)
+#define INFMON_API_WARN(...) vlib_log_warn(infmon_log_api, __VA_ARGS__)
+#define INFMON_API_NOTICE(...) vlib_log_notice(infmon_log_api, __VA_ARGS__)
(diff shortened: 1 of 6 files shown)


src/backend/include/infmon/log.h
```diff
@@ -0,0 +1,126 @@
...
+#define INFMON_NODE_LOG_WARN(...) vlib_log_warn(infmon_log_node, __VA_ARGS__)
+#define INFMON_NODE_LOG_NOTICE(...) vlib_log_notice(infmon_log_node, __VA_ARGS__)
+#define INFMON_NODE_LOG_INFO(...) vlib_log_info(infmon_log_node, __VA_ARGS__)
+#define INFMON_NODE_LOG_DEBUG(...) vlib_log_debug(infmon_log_node, __VA_ARGS__)
+
+/* infmon/counter */
+#define INFMON_CTR_ERR(...) vlib_log_err(infmon_log_counter, __VA_ARGS__)
+#define INFMON_CTR_WARN(...) vlib_log_warn(infmon_log_counter, __VA_ARGS__)
+#define INFMON_CTR_NOTICE(...) vlib_log_notice(infmon_log_counter, __VA_ARGS__)
+#define INFMON_CTR_INFO(...) vlib_log_info(infmon_log_counter, __VA_ARGS__)
```
banidoru, line 76: **Inconsistent naming**: Node macros use `INFMON_NODE_LOG_*` while every other subsystem uses the shorter `INFMON_<SUBSYS>_*` pattern (e.g. `INFMON_API_ERR`, `INFMON_CTR_ERR`). This will trip up anyone grepping for `INFMON_NODE_ERR`. Consider renaming to `INFMON_NODE_ERR`, `INFMON_NODE_WARN`, etc. for consistency.

src/backend/include/infmon/log.h
```diff
@@ -0,0 +1,126 @@
...
+extern vlib_log_class_t infmon_log_general;
+extern vlib_log_class_t infmon_log_api;
+extern vlib_log_class_t infmon_log_rule;
+extern vlib_log_class_t infmon_log_node;
+extern vlib_log_class_t infmon_log_counter;
+extern vlib_log_class_t infmon_log_stats;
+
+/* ── Convenience macros — one per (subclass, level) pair ───────── */
+
+/* infmon/general */
```
banidoru, line 44: Nit: All 60 macros (VPP + no-op stubs) follow the same mechanical pattern. A two-level macro like:
```c
#define INFMON_LOG_(cls, lvl, ...) vlib_log_##lvl(cls, __VA_ARGS__)
#define INFMON_API_ERR(...) INFMON_LOG_(infmon_log_api, err, __VA_ARGS__)
```
wouldn't help much on its own, but an `X-macro` or `#define INFMON_DECLARE_LOG_MACROS(PREFIX, cls)` generating all 5 levels per subsystem would cut this from ~60 lines to ~12 and make adding a new subsystem trivial. Not blocking, but worth considering before this list grows.

src/backend/src/vpp/infmon_log.c
```diff
@@ -0,0 +1,36 @@
...
+    infmon_log_rule = vlib_log_register_class("infmon", "rule");
+    infmon_log_node = vlib_log_register_class("infmon", "node");
+    infmon_log_counter = vlib_log_register_class("infmon", "counter");
+    infmon_log_stats = vlib_log_register_class("infmon", "stats");
+
+    vlib_log_notice(infmon_log_general, "logging subsystem initialized");
+    return 0;
+}
+
+/* Run early so log classes are available before other init functions. */
```
banidoru: `VLIB_INIT_FUNCTION` does not guarantee ordering relative to other `VLIB_INIT_FUNCTION` registrations. If any other InFMon init function (e.g. node/API registration) logs via these macros before `infmon_log_init` runs, the handles will be zero-valued. VPP treats 0 as a valid (default) log class, so it won't crash, but messages would be mis-categorized.

Consider using `VLIB_INIT_FUNCTION` with `.init_order = VLIB_INIT_ORDER_FIRST` or adding explicit `VLIB_INIT_FUNCTION_DEPENDENCY` in the other init functions to ensure ordering.

tests/e2e/test_packet_replay.py
```diff
@@ -64,8 +65,13 @@ def test_packet_replay(scenario: str, infmon_env: dict) -> None:
...
     refresh_mode = os.environ.get("INFMON_E2E_TEST_REFRESH_BASELINE", "0") == "1"
 
-    # 1. Clear all existing flow rules
-    clear_all_flow_rules()
+    # 1. Remove any stale rule from a prior run of *this* scenario.
+    #    Do NOT clear all rules — other xdist workers may be running
+    #    concurrently with their own rules.
+    try:
+        flow_rule_rm(scenario)
+    except subprocess.CalledProcessError:
```
banidoru: The bare `except subprocess.CalledProcessError: pass` silently swallows *all* failures from `flow_rule_rm`, not just "rule doesn't exist". If the API server is down or returns an unexpected error, this will hide it. Consider logging a debug message, or checking the return code / stderr for the specific "not found" condition.

banidoru (commented):
- **Build fix is correct** — enum rename in `infmon_nodes.c` looks right.
- **Inconsistent macro naming** — node subsystem uses `INFMON_NODE_LOG_*` while all others use `INFMON_<SUBSYS>_*` (e.g. `INFMON_API_ERR`). Should be `INFMON_NODE_ERR` etc. for consistency.
- **Macro boilerplate** — 60 macros (30 VPP + 30 no-op) could be collapsed with a parameterized helper macro to reduce maintenance burden.
- **Init ordering risk** — `VLIB_INIT_FUNCTION` doesn't guarantee ordering; if other init functions log before `infmon_log_init`, handles are zero-initialized (may silently misbehave).
- **E2E parallel safety looks good** — per-scenario `flow_rule_rm` instead of `clear_all_flow_rules` is the right fix for xdist concurrency.

tests/e2e/test_packet_replay.py
```diff
@@ -7,11 +7,12 @@
 import json
 import os
 import pathlib
+import subprocess
 from typing import List
 
 import pytest
```
banidoru, line 14: **Unused import**: `clear_all_flow_rules` is still imported here but is no longer used anywhere in this file after the refactor to per-scenario `flow_rule_rm`. Remove it to keep the import list clean and avoid lint warnings.

```suggestion
from helpers import flow_rule_add, flow_rule_rm, wait_for_stats
```

banidoru (changes requested):
- The new logging framework is a solid addition, but there are some subtle race conditions and robustness issues to address.
- The E2E test cleanup logic is not safe for concurrent runs (`pytest -n`). It removes its own rule but doesn’t prevent another worker from re-adding it before the test completes, leading to flakes.
- The `VLIB_INIT_FUNCTION` for logging has a potential race condition if other init functions log before it runs.
- The use of `CLIB_UNUSED` for the `vlib_main_t *` parameter in the init function is risky; a direct `(void)vm;` is safer against future VPP API changes.
- Added a note about the risks of calling these logging macros from signal-handler context, which could lead to deadlocks.

banidoru (commented):
- **Build fix is correct** — enum rename from `INFMON_COUNTER_ERROR_DROP` → `INFMON_NODE_ERR_COUNTER_TABLE_FULL` matches the definition and all other call sites.
- **Parallel-safe E2E refactor is sound** — per-scenario `flow_rule_rm` instead of `clear_all_flow_rules` is the right approach for xdist safety.
- **Unused import** — `clear_all_flow_rules` is still imported but no longer called; should be removed.
- **Structured logging setup looks good** — clean separation of log classes with proper VPP lifecycle integration.
- **`#ifdef` guard in `.c` is belt-and-suspenders** — `infmon_log.c` is only compiled for the VPP plugin target (which defines `INFMON_VPP_BUILD`), so the whole-file `#ifdef` is technically redundant. Not a blocker, just noting it.

banidoru (approved):
All reviewers approved. LGTM.

src/backend/include/infmon/log.h
r12f: Good catch — renamed `INFMON_NODE_LOG_*` to `INFMON_NODE_*` to match the `INFMON_<SUBSYS>_*` pattern used by all other subsystems. No callers needed updating since these macros weren't used in .c files yet. Fixed in 14402b0.

src/backend/include/infmon/log.h
r12f: Agreed this would be a nice cleanup as the subsystem list grows. For now with 6 subsystems (60 lines) I'd prefer to keep it explicit and avoid the macro-generation indirection, but will revisit if we add more. Filed mentally as a follow-up refactor.

src/backend/src/vpp/infmon_log.c
r12f: Valid concern. Added `.runs_before = VLIB_INITS("infmon_api_init", "infmon_node_init")` so `infmon_log_init` is guaranteed to complete before any future API or node init functions that might log. Currently there's only one `VLIB_INIT_FUNCTION` in the plugin, but this is defensive for when more are added. Fixed in 14402b0.

tests/e2e/test_packet_replay.py
r12f: Good point. Now the `except` block logs unexpected failures at debug level instead of silently swallowing them. We still proceed in both cases (rule might not exist, or the cleanup is best-effort before re-creation), but non-"not found" errors are now visible. Fixed in 14402b0.

tests/e2e/test_packet_replay.py
r12f: Removed the unused `clear_all_flow_rules` import — good catch. Fixed in 5fd1dbe.

banidoru (approved):
**Verdict: APPROVE**

- All 5 prior review threads addressed and resolved — naming consistency, init ordering, error handling, unused import all fixed
- `runs_before` dependency on `infmon_log_init` is correct defensive ordering for future init functions
- Parallel-safe E2E cleanup via per-scenario `flow_rule_rm` is a solid improvement over global `clear_all_flow_rules`
- Minor: `import logging` inside `except` block (test_packet_replay.py ~L76) is unconventional — consider moving to top-level imports
- Clean, well-structured diff overall

banidoru (approved):
LGTM, changes look solid.

tests/e2e/test_packet_replay.py
```diff
@@ -127,5 +141,8 @@ def test_packet_replay(scenario: str, infmon_env: dict) -> None:
                 f"Actual:   {json.dumps(stats, indent=2)}"
             )
     finally:
-        # Ensure flow rules are cleaned up even on failure
-        clear_all_flow_rules()
+        # Ensure this scenario's flow rule is cleaned up even on failure
+        try:
+            flow_rule_rm(scenario)
+        except subprocess.CalledProcessError:
+            pass
```
banidoru, line 148: Minor: This `except CalledProcessError: pass` still silently swallows all failures, unlike the setup block which now logs at debug level. Consider adding the same `logging.debug(...)` pattern here for consistency — helps when debugging teardown issues in CI.

banidoru (approved):
LGTM — all prior concerns addressed.

- Naming consistency fixed (`INFMON_NODE_*`)
- Init ordering secured via `.runs_before`
- Unused import removed
- Setup exception handling now logs instead of silently swallowing
- One minor nit: `finally` block teardown (line ~144) still uses bare `except: pass` — consider adding the same debug logging as the setup block for consistency, but not a blocker.

banidoru (approved):
All reviewers approved. LGTM.

PR #164, opened 2026-04-24
feat: add logging to backend API handlers and VPP nodes

r12f:
## Summary

Add comprehensive logging across the InFMon backend using the existing `infmon/log.h` logging infrastructure. Previously the log subsystem was defined but no log statements were present.

### `api_handler.c` (portable layer)
- `ctx_init` / `ctx_destroy`: log lifecycle events
- `flow_rule_add`: log rule name, fields, max_keys on entry; success with idx/key_width/workers; failure with error code
- `flow_rule_del`: log name on entry, success with remaining count, failure with error
- `flow_rule_list`: log rule count
- `snapshot_and_clear`: log flow_rule_id on entry; success with generation/retired/slots/arena stats; failure with result code
- `status`: log worker count

### `infmon_api_handlers.c` (VPP VAPI handlers)
- Every handler logs on entry with key parameters (rule name, flow_rule_id, etc.)
- `flow_rule_add` / `flow_rule_del`: log success (with publish confirmation) and failure
- `snapshot_and_clear` / `snapshot_inline_dump`: log result codes
- `flow_rule_list_dump` / `flow_rule_get` / `status_dump`: log invocations

### `infmon_nodes.c` (VPP data-plane nodes)
- `erspan-decap`: log packets processed per frame
- `flow-match`: log frame size, rule_count, and rule_set pointer (helps diagnose NULL rules)
- `counter`: log frame size, worker_id, tick, insert_retry, table_full counts

### `infmon_log.c`
- Fix include order: `vlib/vlib.h` before `vlib/log.h` (build fix)

## Log levels
- **INFO**: Rule add/del operations, snapshot requests (infrequent control-plane ops)
- **DEBUG**: Per-frame node stats, list/get/status queries (high-frequency, off by default)
- **WARN/ERR**: Failures and unexpected conditions

## Testing
- All 10 unit tests pass
- Build succeeds with no warnings

### DIFF
--- a/src/backend/src/api_handler.c
+++ b/src/backend/src/api_handler.c
@@ -5,8 +5,10 @@
 
 #include "infmon/api_handler.h"
 
+#include <inttypes.h>
 #include <string.h>
 
+#include "infmon/log.h"
 #include "infmon/snapshot.h" /* INFMON_MAX_WORKERS */
 
 /* ── Helpers ─────────────────────────────────────────────────────── */
@@ -90,12 +92,14 @@ void infmon_api_ctx_init(infmon_api_ctx_t *ctx, infmon_flow_rule_set_t *rule_set
     memset(ctx, 0, sizeof(*ctx));
     ctx->rule_set = rule_set;
     ctx->stats_reg = stats_reg;
+    INFMON_API_INFO("ctx_init: rule_set=%p stats_reg=%p", (void *) rule_set, (void *) stats_reg);
 }
 
 void infmon_api_ctx_destroy(infmon_api_ctx_t *ctx)
 {
     if (!ctx)
         return;
+    INFMON_API_INFO("ctx_destroy: ctx=%p", (void *) ctx);
     for (uint32_t w = 0; w < INFMON_MAX_WORKERS; w++) {
         for (uint32_t i = 0; i < INFMON_FLOW_RULE_SET_MAX; i++) {
             if (ctx->tables[w][i]) {
@@ -119,10 +123,15 @@ static infmon_api_result_t flow_rule_add_internal(infmon_api_ctx_t *ctx,
     if (!ctx || !rule)
         return INFMON_API_ERR_INVALID_RULE;
 
+    INFMON_RULE_INFO("flow_rule_add: name='%s' fields=%u max_keys=%u", rule->name,
+                     rule->field_count, rule->max_keys);
+
     /* 1. Insert into the rule set (validates + checks budget). */
     infmon_flow_rule_result_t rr = infmon_flow_rule_add(ctx->rule_set, rule);
-    if (rr != INFMON_FLOW_RULE_OK)
+    if (rr != INFMON_FLOW_RULE_OK) {
+        INFMON_RULE_ERR("flow_rule_add: '%s' insert failed: %d", rule->name, (int) rr);
         return map_rule_result(rr);
+    }
 
     /* 2. Find the index the rule landed at. */
     uint32_t idx = find_rule_index(ctx->rule_set, rule->name);
@@ -163,6 +172,9 @@ static infmon_api_result_t flow_rule_add_internal(infmon_api_ctx_t *ctx,
         memset(&ctx->flow_rule_ids[idx], 0, sizeof(ctx->flow_rule_ids[0]));
     }
 
+    INFMON_RULE_INFO("flow_rule_add: '%s' ok — idx=%u key_width=%u workers=%u", rule->name, idx,
+                     inserted->key_width, ctx->worker_count > 0 ? ctx->worker_count : 1);
+
     return INFMON_API_OK;
 }
 
@@ -182,6 +194,8 @@ infmon_api_result_t infmon_api_flow_rule_del(infmon_api_ctx_t *ctx, const char *
 {
     if (!ctx || !name)
         return INFMON_API_ERR_INVALID_RULE;
+
+    INFMON_RULE_INFO("flow_rule_del: name='%s'", name);
 
     /* 1. Find the rule index before removing. */
     uint32_t idx = find_rule_index(ctx->rule_set, name);
@@ -213,6 +227,8 @@ infmon_api_result_t infmon_api_flow_rule_del(infmon_api_ctx_t *ctx, const char *
         ctx->tables[w][count] = NULL;
     memset(&ctx->flow_rule_ids[count], 0, sizeof(ctx->flow_rule_ids[0]));
 
+    INFMON_RULE_INFO("flow_rule_del: '%s' ok — removed idx=%u, %u rules remain", name, idx, count);
+
     return INFMON_API_OK;
 }
 
@@ -228,10 +244,8 @@ infmon_api_result_t infmon_api_flow_rule_list(const infmon_api_ctx_t *ctx,
     if (!cb)
         return INFMON_API_OK;
 
-    /* The rule set is dense (add appends, rm compacts), so
-     * infmon_flow_rule_get() should not return NULL for i < count.
-     * The NULL guard is purely defensive. */
     uint32_t n = infmon_flow_rule_count(ctx->rule_set);
+    INFMON_API_DEBUG("flow_rule_list: %u rules", n);
     for (uint32_t i = 0; i < n; i++) {
         const infmon_flow_rule_t *r = infmon_flow_rule_get(ctx->rule_set, i);
         if (r)
@@ -267,6 +281,9 @@ infmon_api_result_t infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx,
                                                   infmon_flow_rule_id_t flow_rule_id,
                                                   infmon_api_snap_reply_t *reply)
 {
+    INFMON_CTR_DEBUG("snapshot_and_clear: id=%016" PRIx64 "-%016" PRIx64, flow_rule_id.hi,
+                     flow_rule_id.lo);
+
     if (!reply)
         return INFMON_API_ERR_INTERNAL;
 
@@ -316,8 +333,11 @@ infmon_api_result_t infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx,
 
     reply->result = map_snap_result(snap_reply.result);
 
-    if (snap_reply.result != INFMON_SNAP_OK)
+    if (snap_reply.result != INFMON_SNAP_OK) {
+        INFMON_CTR_WARN("snapshot_and_clear: snap failed for idx=%u result=%d", idx,
+                        (int) snap_reply.result);
         return reply->result;
+    }
 
     /* Build the descriptor by aggregating across all retired worker tables. */
     infmon_counter_table_t *retired = snap_reply.retired_tables[0];
@@ -354,6 +374,10 @@ infmon_api_result_t infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx,
         reply->retired_tables[w] = snap_reply.retired_tables[w];
     reply->num_retired = snap_reply.num_retired;
 
+    INFMON_CTR_DEBUG(
+        "snapshot_and_clear: ok — idx=%u gen=%" PRIu64 " retired=%u slots_len=%u arena_used=%u",
+        idx, desc->generation, snap_reply.num_retired, desc->slots_len, desc->key_arena_used);
+
     return INFMON_API_OK;
 }
 
@@ -389,5 +413,8 @@ infmon_api_result_t infmon_api_status(const infmon_api_ctx_t *ctx, infmon_api_st
     reply->workers = ctx->worker_counters;
     reply->worker_count = ctx->worker_count;
     reply->result = INFMON_API_OK;
+
+    INFMON_API_DEBUG("status: worker_count=%u", reply->worker_count);
+
     return INFMON_API_OK;
 }
--- a/src/backend/src/api_handler.c
+++ b/src/backend/src/api_handler.c
@@ -92,14 +92,14 @@ void infmon_api_ctx_init(infmon_api_ctx_t *ctx, infmon_flow_rule_set_t *rule_set
     memset(ctx, 0, sizeof(*ctx));
     ctx->rule_set = rule_set;
     ctx->stats_reg = stats_reg;
-    INFMON_API_INFO("ctx_init: rule_set=%p stats_reg=%p", (void *) rule_set, (void *) stats_reg);
+    INFMON_API_DEBUG("ctx_init: ready");
 }
 
 void infmon_api_ctx_destroy(infmon_api_ctx_t *ctx)
 {
     if (!ctx)
         return;
-    INFMON_API_INFO("ctx_destroy: ctx=%p", (void *) ctx);
+    INFMON_API_DEBUG("ctx_destroy: cleaning up");
     for (uint32_t w = 0; w < INFMON_MAX_WORKERS; w++) {
         for (uint32_t i = 0; i < INFMON_FLOW_RULE_SET_MAX; i++) {
             if (ctx->tables[w][i]) {
@@ -248,6 +248,9 @@ infmon_api_result_t infmon_api_flow_rule_list(const infmon_api_ctx_t *ctx,
     INFMON_API_DEBUG("flow_rule_list: %u rules", n);
     for (uint32_t i = 0; i < n; i++) {
         const infmon_flow_rule_t *r = infmon_flow_rule_get(ctx->rule_set, i);
+        /* The rule set is dense (add appends, rm compacts), so
+         * infmon_flow_rule_get() should not return NULL for i < count.
+         * The NULL guard is purely defensive. */
         if (r)
             cb(r, i, user);
     }
@@ -281,11 +284,11 @@ infmon_api_result_t infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx,
                                                   infmon_flow_rule_id_t flow_rule_id,
                                                   infmon_api_snap_reply_t *reply)
 {
-    INFMON_CTR_DEBUG("snapshot_and_clear: id=%016" PRIx64 "-%016" PRIx64, flow_rule_id.hi,
-                     flow_rule_id.lo);
-
     if (!reply)
         return INFMON_API_ERR_INTERNAL;
+
+    INFMON_CTR_DEBUG("snapshot_and_clear: id=%016" PRIx64 "-%016" PRIx64, flow_rule_id.hi,
+                     flow_rule_id.lo);
 
     memset(reply, 0, sizeof(*reply));
 
--- a/src/backend/src/api_handler.c
+++ b/src/backend/src/api_handler.c
@@ -172,8 +172,8 @@ static infmon_api_result_t flow_rule_add_internal(infmon_api_ctx_t *ctx,
         memset(&ctx->flow_rule_ids[idx], 0, sizeof(ctx->flow_rule_ids[0]));
     }
 
-    INFMON_RULE_INFO("flow_rule_add: '%s' ok — idx=%u key_width=%u workers=%u", rule->name, idx,
-                     inserted->key_width, ctx->worker_count > 0 ? ctx->worker_count : 1);
+    INFMON_RULE_DEBUG("flow_rule_add: '%s' ok — idx=%u key_width=%u workers=%u", rule->name, idx,
+                      inserted->key_width, ctx->worker_count > 0 ? ctx->worker_count : 1);
 
     return INFMON_API_OK;
 }
--- a/src/backend/src/api_handler.c
+++ b/src/backend/src/api_handler.c
@@ -123,7 +123,7 @@ static infmon_api_result_t flow_rule_add_internal(infmon_api_ctx_t *ctx,
     if (!ctx || !rule)
         return INFMON_API_ERR_INVALID_RULE;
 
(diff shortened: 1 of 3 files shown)


src/backend/src/vpp/infmon_nodes.c
```diff
@@ -239,6 +241,8 @@ static uword infmon_erspan_decap_node_fn(vlib_main_t *vm, vlib_node_runtime_t *n
     vlib_put_next_frame(vm, node, next_match, n_left_match);
     vlib_put_next_frame(vm, node, next_pass, n_left_pass);
 
+    INFMON_NODE_DEBUG("erspan-decap: processed %u pkts", frame->n_vectors);
```
banidoru: This runs once per frame on the data-plane hot path. `vlib_log_debug` is a real function call that checks the log level each time — at multi-Mpps rates this adds non-trivial overhead even when debug logging is off.

Consider guarding with something like:
```c
if (PREDICT_FALSE(vlib_log_is_enabled(infmon_log_node, VLIB_LOG_LEVEL_DEBUG)))
    INFMON_NODE_DEBUG(...);
```
Or skip per-frame logging entirely and rely on the existing VPP node error counters (which are already incremented).

Same concern applies to the `flow-match` and `counter` node functions below.

src/backend/src/api_handler.c
```diff
@@ -228,10 +244,8 @@ infmon_api_result_t infmon_api_flow_rule_list(const infmon_api_ctx_t *ctx,
     if (!cb)
         return INFMON_API_OK;
 
-    /* The rule set is dense (add appends, rm compacts), so
-     * infmon_flow_rule_get() should not return NULL for i < count.
-     * The NULL guard is purely defensive. */
     uint32_t n = infmon_flow_rule_count(ctx->rule_set);
```
banidoru, line 247: The removed comment (`The rule set is dense (add appends, rm compacts), so infmon_flow_rule_get() should not return NULL for i < count. The NULL guard is purely defensive.`) was valuable documentation for future readers. The new debug log line doesn't convey the same design rationale. Consider keeping the comment alongside the log statement.

src/backend/src/api_handler.c
```diff
@@ -267,6 +281,9 @@ infmon_api_result_t infmon_api_snapshot_and_clear(infmon_api_ctx_t *ctx,
                                                   infmon_flow_rule_id_t flow_rule_id,
                                                   infmon_api_snap_reply_t *reply)
 {
+    INFMON_CTR_DEBUG("snapshot_and_clear: id=%016" PRIx64 "-%016" PRIx64, flow_rule_id.hi,
```
banidoru: Nit: this log statement executes before the `if (!reply) return ...` check on line 287. It's harmless (just logs on an invalid call), but placing it after the null guard would be more consistent with the pattern used in other functions.

banidoru (commented):
- **Overall**: Clean, well-structured logging addition. Good level choices (INFO for control-plane, DEBUG for hot-path). No correctness issues.
- **Hot-path concern**: `INFMON_NODE_DEBUG` calls in the three graph-node functions run per-frame. Even at DEBUG (off by default), `vlib_log_debug` still incurs a function-call + level-check per invocation — could add up at high pps. Consider a `PREDICT_FALSE` guard or batch stats.
- **Removed comment**: The deleted comment in `flow_rule_list` about the rule set being dense and the NULL guard being defensive was useful documentation that the new log line doesn't replace.
- **Log-before-null-check**: `snapshot_and_clear` logs `flow_rule_id` before checking `if (!reply)` — harmless but logs on invalid calls.
- **Pointer values in logs**: Raw `%p` for `ctx`, `rule_set`, `rs` is fine for debug but leaks ASLR info if logs are exposed. Minor for a VPP plugin.

src/backend/src/api_handler.c
```diff
@@ -90,12 +92,14 @@ void infmon_api_ctx_init(infmon_api_ctx_t *ctx, infmon_flow_rule_set_t *rule_set
     memset(ctx, 0, sizeof(*ctx));
     ctx->rule_set = rule_set;
     ctx->stats_reg = stats_reg;
+    INFMON_API_INFO("ctx_init: rule_set=%p stats_reg=%p", (void *) rule_set, (void *) stats_reg);
```
banidoru: Logging raw pointer values (`rule_set=%p`, `stats_reg=%p`) at **INFO** level is a security concern — it leaks kernel/ASLR address layout to anyone with log access. Same issue in `ctx_destroy` below.

These lifecycle events are also relatively low-value; consider demoting to DEBUG, or dropping the pointer values entirely and logging something more meaningful (e.g. rule capacity, stats registry name).

src/backend/src/vpp/infmon_api_handlers.c
```diff
@@ -161,7 +162,9 @@ static void vl_api_infmon_flow_rule_add_t_handler(vl_api_infmon_flow_rule_add_t
 
     infmon_vpp_api_ctx_ensure();
 
-    /* Convert API message to infmon_flow_rule_t */
+    INFMON_RULE_INFO("vapi: flow_rule_add name='%.*s' fields=%u max_keys=%u",
```
banidoru, line 165: `(int) sizeof(mp->name)` gives the **full buffer size** (e.g. 64), not the actual string length. If the name is shorter, `%.*s` will print trailing null bytes or padding garbage on some platforms.

Use `strnlen(mp->name, sizeof(mp->name))` as the precision instead:
```c
INFMON_RULE_INFO("vapi: flow_rule_add name='%.*s' ...",
                 (int) strnlen(mp->name, sizeof(mp->name)), mp->name, ...);
```

src/backend/src/vpp/infmon_api_handlers.c
```diff
@@ -378,6 +400,10 @@ static void vl_api_infmon_snapshot_and_clear_t_handler(vl_api_infmon_snapshot_an
 
     infmon_vpp_api_ctx_ensure();
 
+    INFMON_CTR_INFO("vapi: snapshot_and_clear id=%016" PRIx64 "-%016" PRIx64,
```
banidoru: This logs at **INFO** but the same `snapshot_and_clear` entry-point log in `api_handler.c` (line 284) uses **DEBUG**. Since snapshots can be frequent (polled by the control plane), INFO here risks log spam. Suggest aligning both to DEBUG, or keeping INFO only in one layer.

Same inconsistency exists for `snapshot_inline_dump` (INFO here vs. no log in the portable layer).

banidoru (commented):
- Pointer values logged at INFO level (`ctx=%p`, `rule_set=%p`, `rs=%p`) leak ASLR addresses — demote to DEBUG or remove
- Inconsistent log levels: `snapshot_and_clear` is DEBUG in api_handler.c but INFO in the VAPI handler for the same operation
- `%.*s` with `sizeof(mp->name)` prints full buffer width including padding; use `strnlen` for the precision instead
- Overall: good coverage of control-plane paths; data-plane hot-path concern already noted

src/backend/src/vpp/infmon_api_handlers.c
banidoru, line 165: Both this VAPI handler and the core `flow_rule_add_internal()` in `api_handler.c` (line ~126) log the same entry info (name, fields, max_keys) at INFO. Every `flow_rule_add` call produces two nearly identical log lines. Consider making the inner (core) layer DEBUG and keeping only the VAPI layer at INFO, or vice versa — but not both at INFO.

src/backend/src/vpp/infmon_api_handlers.c
```diff
@@ -428,6 +456,10 @@ static void vl_api_infmon_snapshot_inline_dump_t_handler(vl_api_infmon_snapshot_
 
     infmon_vpp_api_ctx_ensure();
 
+    INFMON_CTR_INFO("vapi: snapshot_inline_dump id=%016" PRIx64 "-%016" PRIx64,
```
banidoru: `snapshot_inline_dump` logs entry at `INFMON_CTR_INFO` here, matching the `snapshot_and_clear` VAPI handler (line 403). But the core implementation in `api_handler.c` uses `INFMON_CTR_DEBUG` for the same operation. The mismatch means enabling INFO shows the outer call but not the inner details — either promote the core to INFO or demote these to DEBUG.

src/backend/src/vpp/infmon_nodes.c
```diff
@@ -278,6 +282,9 @@ static uword infmon_flow_match_node_fn(vlib_main_t *vm, vlib_node_runtime_t *nod
     const infmon_flow_rule_t *rules = rs ? rs->rules : NULL;
     uint32_t rule_count = rs ? rs->count : 0;
 
+    INFMON_NODE_DEBUG("flow-match: frame %u pkts, rule_count=%u rs=%p", frame->n_vectors,
```
banidoru: Logging `rs=%p` (raw pointer) in the data-plane node adds no diagnostic value — it's a module-scoped pointer that rarely changes. Drop it or replace with something actionable like the published generation number.

banidoru (commented):
- Dual-layer logging (VAPI handler + core API) means every operation produces two near-identical log lines — consider logging only at one layer or making the inner layer DEBUG
- `snapshot_inline_dump` and `snapshot_and_clear` VAPI handlers both log at INFO, but the core `api_handler.c` counterpart uses DEBUG — pick one level consistently
- Node hot-path logging (all three nodes) will generate one log call per frame even at DEBUG; VPP's rate limiter helps but these should be guarded or removed for production builds

banidoru (approved):
All reviewers approved. LGTM.

src/backend/src/vpp/infmon_nodes.c
r12f: Good catch — removed all three per-frame `INFMON_NODE_DEBUG` calls from the erspan-decap, flow-match, and counter node functions. The existing VPP node error counters already track the interesting events (no-match, insert-retry, table-full) at zero per-packet cost.

src/backend/src/api_handler.c
r12f: Agreed — restored the original design-rationale comment alongside the debug log. The comment explains *why* the NULL guard exists, which the log line alone doesn't convey.

src/backend/src/api_handler.c
r12f: Fixed — moved the `INFMON_CTR_DEBUG` call below the `if (!reply) return` null guard for consistency with the other functions.

src/backend/src/api_handler.c
r12f: Good security point. Demoted both `ctx_init` and `ctx_destroy` to DEBUG and removed the raw pointer values entirely. The lifecycle events now log without leaking address layout.

src/backend/src/vpp/infmon_api_handlers.c
r12f: Nice catch — switched to `strnlen(mp->name, sizeof(mp->name))` so `%.*s` uses the actual string length rather than the full buffer size.

src/backend/src/vpp/infmon_api_handlers.c
r12f: Agreed — demoted all three snapshot VAPI entry/result logs (`snapshot_and_clear` entry, `snapshot_and_clear` result, `snapshot_inline_dump` entry) from INFO to DEBUG, aligning with the portable layer which already uses DEBUG.

src/backend/src/vpp/infmon_api_handlers.c
r12f: Good point — demoted the core layer entry log to `INFMON_RULE_DEBUG` so only the VAPI handler logs at INFO. The core success/failure logs remain at INFO since those fire once per completed operation.

src/backend/src/vpp/infmon_api_handlers.c
r12f: Already addressed in the previous commit — all three snapshot VAPI logs (`snapshot_and_clear` entry/result and `snapshot_inline_dump` entry) are now `INFMON_CTR_DEBUG`, matching the core layer.

src/backend/src/vpp/infmon_nodes.c
r12f: Already addressed — all per-frame `INFMON_NODE_DEBUG` calls (including the one logging `rs=%p`) were removed entirely in the prior commit. The existing VPP node error counters cover the actionable signals.

src/backend/src/vpp/infmon_api_handlers.c
banidoru: Addressed — all three snapshot VAPI logs are now at DEBUG, consistent with the core layer.

src/backend/src/vpp/infmon_nodes.c
banidoru: Addressed — the per-frame node debug calls (including rs=%p) have been removed entirely.

src/backend/src/vpp/infmon_api_handlers.c
banidoru: Still open — both the VAPI handler (line 165, INFMON_RULE_INFO) and core flow_rule_add_internal (line ~175, INFMON_RULE_INFO on success) log at INFO. Every flow_rule_add call still produces two near-identical INFO lines. Suggest demoting the core layer success log to DEBUG.

banidoru (commented):
Re-review (iteration 2):

- Prior concerns largely addressed: hot-path node logs removed, pointer leaks fixed, strnlen fix applied, comment restored, null-guard ordering fixed, snapshot logs demoted to DEBUG.
- **Still open:** duplicate INFO logging in both VAPI and core layers for `flow_rule_add` — every add produces two near-identical INFO lines.
- Minor nit: continuation-line alignment in `snapshot_and_clear` and `snapshot_inline_dump` VAPI handlers is off by one space vs. other log calls in the same file.
- Overall quality is solid. One substantive issue remains (duplicate INFO).

src/backend/src/vpp/infmon_api_handlers.c
```diff
@@ -378,6 +400,10 @@ static void vl_api_infmon_snapshot_and_clear_t_handler(vl_api_infmon_snapshot_an
 
     infmon_vpp_api_ctx_ensure();
 
+    INFMON_CTR_DEBUG("vapi: snapshot_and_clear id=%016" PRIx64 "-%016" PRIx64,
```
banidoru, line 403: Nit: continuation lines here use 20-space indent, but the other log macros in this file (e.g. `INFMON_RULE_INFO` at line 165) align to the opening paren (21 spaces). Same at the `snapshot_inline_dump` log below. Trivial but inconsistent.

src/backend/src/vpp/infmon_api_handlers.c
banidoru: Still open in the current diff. Core success log at line 175 (`INFMON_RULE_INFO`) and VAPI success log at line 201 (`INFMON_RULE_INFO`) both fire at INFO on every successful flow_rule_add.

banidoru (commented):
- Prior indent inconsistency (20 vs 21 spaces in continuation lines) still present at lines 403 and 459.
- Duplicate INFO logging for flow_rule_add (VAPI entry + core success) still unresolved — both fire per successful call.
- All other prior concerns have been addressed.
- New diff looks clean — no new issues found.

src/backend/src/vpp/infmon_api_handlers.c
banidoru: Still present — lines 403 and 459 use 20-space indent while line 165 uses 21 (paren-aligned). Minor but worth fixing for consistency.

banidoru: Still present in the latest diff — lines 403 and 459 use 20-space continuation indent while line 165 uses 21 (paren-aligned). Trivial nit, not blocking.

banidoru (approved):
- **Verdict: approved** — all prior concerns addressed, code is clean
- Hot-path node logs removed ✓, pointer values stripped ✓, log levels aligned ✓
- `strnlen` fix for `%.*s` precision ✓, design-rationale comment restored ✓
- Core entry log demoted to DEBUG, snapshot logs aligned to DEBUG ✓
- Minor: continuation indent inconsistency at lines 403/459 in `infmon_api_handlers.c` (20 vs 21 spaces) — not blocking

src/backend/src/vpp/infmon_api_handlers.c
r12f: Fixed in e977d3f — all continuation lines now use 21-space paren-aligned indent consistently.

src/backend/src/vpp/infmon_nodes.c
```diff
@@ -24,6 +25,7 @@
 
 #include "infmon/counter_table.h"
 #include "infmon/graph_node.h"
+#include "infmon/log.h"
```
banidoru: Both `#include <inttypes.h>` (line 18) and `#include "infmon/log.h"` (line 28) were added in support of `INFMON_NODE_DEBUG` calls that have since been removed. These are now unused includes — please remove them to keep the diff minimal.

src/backend/src/api_handler.c
```diff
@@ -163,6 +172,9 @@ static infmon_api_result_t flow_rule_add_internal(infmon_api_ctx_t *ctx,
         memset(&ctx->flow_rule_ids[idx], 0, sizeof(ctx->flow_rule_ids[0]));
     }
 
+    INFMON_RULE_INFO("flow_rule_add: '%s' ok — idx=%u key_width=%u workers=%u", rule->name, idx,
```
banidoru: This `INFMON_RULE_INFO` fires on every successful `flow_rule_add` — but the VAPI handler also logs success at INFO (line 201 in `infmon_api_handlers.c`). Net effect: 2 INFO lines per successful add. Consider demoting this one to DEBUG since the VAPI layer already covers it. Non-blocking nit.

banidoru (commented):
Iteration 3 re-review:

- All 10 prior review threads are resolved — fixes look correct
- Unused `#include "infmon/log.h"` and `#include <inttypes.h>` in `infmon_nodes.c` — all `INFMON_NODE_DEBUG` calls were removed but the includes stayed
- Duplicate INFO on `flow_rule_add` success persists: core (line 175, `INFMON_RULE_INFO`) + VAPI (line 201, `INFMON_RULE_INFO`) — 2 INFO lines per successful add. Non-blocking but noisy
- Everything else looks good — log levels, null guards, `strnlen` fix, comment restoration all landed correctly

src/backend/src/vpp/infmon_nodes.c
banidoru: Still present in the latest diff — `#include <inttypes.h>` and `#include "infmon/log.h"` remain in `infmon_nodes.c` with no remaining callers after the NODE_DEBUG removal. Please remove these unused includes.

src/backend/src/api_handler.c
banidoru: Still present — `INFMON_RULE_INFO` at line 175 in `api_handler.c` and the VAPI handler success log at line 201 both fire at INFO on every successful add. Two INFO lines per operation. Non-blocking nit but worth demoting one to DEBUG.

banidoru (commented):
- Most prior feedback addressed well — pointer leaks fixed, log levels aligned, hot-path logging removed, comment restored, strnlen fix applied
- **Unused includes still present:** `infmon_nodes.c` keeps `<inttypes.h>` and `"infmon/log.h"` after all NODE_DEBUG calls were removed — dead includes should be cleaned up
- **Duplicate INFO on flow_rule_add success:** core layer (line 175) and VAPI layer (line 201) both log at INFO — demote one to DEBUG
- Overall: good logging additions with consistent patterns; two minor items remain from prior rounds

banidoru (approved):
- All major issues from prior iterations addressed: hot-path logging removed, pointer leaks fixed, log levels aligned, strnlen fix applied, comment restored, null-guard ordering fixed
- Two non-blocking nits remain: unused includes in `infmon_nodes.c` (`inttypes.h`, `log.h`) and duplicate INFO on `flow_rule_add` success (core + VAPI both at INFO)
- Continuation indent inconsistency fixed in latest commit
- Overall logging coverage is solid — good entry/exit/error tracing across API and VAPI layers
- LGTM — approve with minor nits noted above

banidoru (approved):
All reviewers approved. LGTM.

src/backend/src/vpp/infmon_nodes.c
r12f: Removed both unused includes (`#include <inttypes.h>` and `#include "infmon/log.h"`) in 5f29fd7 — they were leftover from the removed NODE_DEBUG calls.

src/backend/src/api_handler.c
r12f: Demoted the core-layer success log from INFMON_RULE_INFO to INFMON_RULE_DEBUG in 5f29fd7. Now only the VAPI handler logs at INFO, avoiding the duplicate per-operation INFO lines.

banidoru (approved):
$Iteration 4 — all prior feedback addressed. Clean diff.\n\n- Core-layer success log correctly demoted to DEBUG, eliminating duplicate INFO noise\n- Unused includes removed from infmon_nodes.c\n- Log levels well-chosen: INFO for control-plane ops, DEBUG for queries/internals, WARN/ERR for failures\n- No new issues found\n\nLGTM.

banidoru (approved):
All review comments have been addressed. The changes look good.

PR #166, opened 2026-04-24
fix(frontend): wire poller snapshots to ControlState for stats show

r12f:
## Problem

`infmonctl stats show` always returns zeros even when VPP nodes are actively processing and counting flows.

**Root cause:** `ControlState.latest_snapshot` is initialized to `None` and never written to. The poller produces `FlowStatsSnapshot` via VAPI `snapshot_inline_dump` every polling interval, but only forwards snapshots to exporter channels (e.g. Kafka, file). The control server — which serves `infmonctl stats show` — reads from `ControlState.latest_snapshot` which stays `None` forever, so it always returns zero counters.

## Fix

- Add `ControlState::update_snapshot()` method to write a new snapshot into the `Mutex<Option<Arc<FlowStatsSnapshot>>>`.
- Create a `sync_channel(1)` whose receiver drives a small `snapshot-to-control` thread that calls `update_snapshot()` on each snapshot.
- Append the sender to the poller's `raw_senders` list so it participates in the existing broadcast mechanism alongside exporters.
- Clone the `control_state` Arc before passing it to the control server so the forwarding thread can share ownership.

## Testing

- `cargo build --release -p infmon-frontend` passes
- Pre-commit hooks (rustfmt, clippy, DCO) pass

### DIFF
--- a/src/frontend/src/control.rs
+++ b/src/frontend/src/control.rs
@@ -52,6 +52,14 @@ impl ControlState {
         }
     }
 
+    /// Replace the latest snapshot (called by the snapshot‐forwarding thread).
+    pub fn update_snapshot(&self, snap: Arc<FlowStatsSnapshot>) {
+        *self
+            .latest_snapshot
+            .lock()
+            .unwrap_or_else(|e| e.into_inner()) = Some(snap);
+    }
+
     /// Create a ControlState with a VAPI control client for backend forwarding.
     #[cfg(feature = "vapi")]
     pub fn with_vapi(initial_rules: Vec<FlowRule>, client: VapiControlClient) -> Self {
--- a/src/frontend/src/control.rs
+++ b/src/frontend/src/control.rs
@@ -53,6 +53,9 @@ impl ControlState {
     }
 
     /// Replace the latest snapshot (called by the snapshot‐forwarding thread).
+    ///
+    /// Intentionally recovers from mutex poison (`into_inner()`) — losing a
+    /// snapshot is preferable to panicking on the stats path.
     pub fn update_snapshot(&self, snap: Arc<FlowStatsSnapshot>) {
         *self
             .latest_snapshot
--- a/src/frontend/src/lifecycle.rs
+++ b/src/frontend/src/lifecycle.rs
@@ -12,6 +12,8 @@ use crate::exporter::{
 };
 use crate::poller::{self, PollerConfig, PollerHandle};
 
+use infmon_common::ipc::types::FlowStatsSnapshot;
+
 /// Errors during lifecycle operations.
 #[derive(Debug)]
 pub enum LifecycleError {
@@ -143,13 +145,7 @@ impl Frontend {
             interval: Duration::from_millis(frontend_cfg.polling_interval_ms),
         };
 
-        let raw_senders: Vec<_> = exporter_senders
-            .iter()
-            .map(|s| s.as_raw_sender().clone())
-            .collect();
-        let poller_handle = poller::spawn(poller_config, raw_senders);
-
-        // Spawn control server for CLI RPCs
+        // Spawn control server for CLI RPCs (before poller so we can wire the snapshot channel)
         let control_socket = PathBuf::from(&frontend_cfg.control_socket);
         #[cfg(feature = "vapi")]
         let control_state = {
@@ -166,7 +162,7 @@ impl Frontend {
         };
         #[cfg(not(feature = "vapi"))]
         let control_state = Arc::new(ControlState::new(config.flow_rules.clone()));
-        let control_handle = match control::spawn(&control_socket, control_state) {
+        let control_handle = match control::spawn(&control_socket, control_state.clone()) {
             Ok(h) => {
                 tracing::info!("control server listening on {}", control_socket.display());
                 Some(h)
@@ -176,6 +172,30 @@ impl Frontend {
                 None
             }
         };
+
+        // Create a snapshot channel that forwards to the control state so
+        // `infmonctl stats show` can read the latest counters.
+        let (stats_tx, stats_rx) = std::sync::mpsc::sync_channel::<Arc<FlowStatsSnapshot>>(1);
+        {
+            let cs = control_state.clone();
+            std::thread::Builder::new()
+                .name("snapshot-to-control".into())
+                .spawn(move || {
+                    while let Ok(snap) = stats_rx.recv() {
+                        cs.update_snapshot(snap);
+                    }
+                    tracing::debug!("snapshot-to-control thread exiting");
+                })
+                .expect("failed to spawn snapshot-to-control thread");
+        }
+
+        // Spawn poller with exporter senders + control-state sender
+        let mut raw_senders: Vec<_> = exporter_senders
+            .iter()
+            .map(|s| s.as_raw_sender().clone())
+            .collect();
+        raw_senders.push(stats_tx);
+        let poller_handle = poller::spawn(poller_config, raw_senders);
 
         Ok(Frontend {
             poller_handle: Some(poller_handle),
--- a/src/frontend/src/lifecycle.rs
+++ b/src/frontend/src/lifecycle.rs
@@ -176,20 +176,28 @@ impl Frontend {
         };
 
         // Create a snapshot channel that forwards to the control state so
-        // `infmonctl stats show` can read the latest counters.
+        // `infmonctl stats show` can read the latest counters.  Only useful
+        // when the control server is running — skip if it failed to start.
         // Capacity 1: poller drops stale snapshots via try_send, we only need the latest.
-        let (stats_tx, stats_rx) = std::sync::mpsc::sync_channel::<Arc<FlowStatsSnapshot>>(1);
-        let snapshot_thread = {
+        let (snapshot_thread, stats_tx) = if control_handle.is_some() {
+            let (tx, stats_rx) = std::sync::mpsc::sync_channel::<Arc<FlowStatsSnapshot>>(1);
             let cs = control_state.clone();
-            std::thread::Builder::new()
+            match std::thread::Builder::new()
                 .name("snapshot-to-control".into())
                 .spawn(move || {
                     while let Ok(snap) = stats_rx.recv() {
                         cs.update_snapshot(snap);
                     }
                     tracing::debug!("snapshot-to-control thread exiting");
-                })
-                .expect("failed to spawn snapshot-to-control thread")
+                }) {
+                Ok(h) => (Some(h), Some(tx)),
+                Err(e) => {
+                    tracing::warn!("failed to spawn snapshot-to-control thread: {e}");
+                    (None, None)
+                }
+            }
+        } else {
+            (None, None)
         };
 
         // Spawn poller with exporter senders + control-state sender
@@ -197,7 +205,9 @@ impl Frontend {
             .iter()
             .map(|s| s.as_raw_sender().clone())
             .collect();
-        raw_senders.push(stats_tx);
+        if let Some(tx) = stats_tx {
+            raw_senders.push(tx);
+        }
         let poller_handle = poller::spawn(poller_config, raw_senders);
 
         Ok(Frontend {
@@ -205,7 +215,7 @@ impl Frontend {
             exporter_handles,
             exporter_senders,
             control_handle,
-            snapshot_thread: Some(snapshot_thread),
+            snapshot_thread,
             config_path: config_path.to_path_buf(),
             shutdown,
         })
--- a/src/frontend/src/lifecycle.rs
+++ b/src/frontend/src/lifecycle.rs
@@ -3,6 +3,7 @@
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
+use std::thread::JoinHandle;
 use std::time::Duration;
 
 use crate::control::{self, ControlHandle, ControlState};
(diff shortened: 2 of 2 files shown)


src/frontend/src/lifecycle.rs
```diff
@@ -177,6 +173,30 @@ impl Frontend {
...
 
+        // Create a snapshot channel that forwards to the control state so
+        // `infmonctl stats show` can read the latest counters.
+        let (stats_tx, stats_rx) = std::sync::mpsc::sync_channel::<Arc<FlowStatsSnapshot>>(1);
+        {
+            let cs = control_state.clone();
+            std::thread::Builder::new()
+                .name("snapshot-to-control".into())
+                .spawn(move || {
+                    while let Ok(snap) = stats_rx.recv() {
```
banidoru, line 188: The `JoinHandle` returned by `spawn()` is silently dropped here. During `Frontend::stop()`, the poller is stopped first (dropping `stats_tx`), which will cause this thread to exit — but there's no `join()` to wait for it. In practice it's harmless (the thread does trivial work), but for deterministic shutdown you should store this handle in `Frontend` and join it in `stop()` between steps 1 and 4, similar to how exporter handles are joined.

src/frontend/src/lifecycle.rs
```diff
@@ -177,6 +173,30 @@ impl Frontend {
             }
         };
 
+        // Create a snapshot channel that forwards to the control state so
```
banidoru, line 178: Capacity of 1 is fine here since the poller uses `try_send` and we only care about the latest snapshot. Consider adding a brief comment like `// capacity 1: poller drops stale snapshots via try_send, we only need the latest` to make the intent explicit for future readers.

banidoru (commented):
- **Overall:** Clean, correct fix for the root cause. The bridging approach via `sync_channel` + forwarding thread fits the existing architecture well.
- **Orphaned thread:** The `snapshot-to-control` thread's `JoinHandle` is dropped immediately — it won't be joined during `Frontend::stop()`. Consider storing it so shutdown is fully deterministic.
- **Channel capacity:** `sync_channel(1)` with `try_send` on the poller side means stale snapshots are silently dropped — correct for "latest stats" semantics, but worth a brief comment noting this is intentional.
- **Poisoned mutex recovery:** `unwrap_or_else(|e| e.into_inner())` is a good defensive choice — the snapshot data is read-only/replaceable so recovering from poison is safe here.

src/frontend/src/control.rs
```diff
@@ -52,6 +52,14 @@ impl ControlState {
         }
     }
 
+    /// Replace the latest snapshot (called by the snapshot‐forwarding thread).
+    pub fn update_snapshot(&self, snap: Arc<FlowStatsSnapshot>) {
+        *self
+            .latest_snapshot
```
banidoru, line 61: Nit: Poison recovery here (`into_inner()`) is the right call for a stats path — losing a snapshot is better than panicking. Just worth a brief doc comment noting this is intentional so future readers don't "fix" it to `.unwrap()`.

src/frontend/src/lifecycle.rs
```diff
@@ -177,6 +173,30 @@ impl Frontend {
...
+                    while let Ok(snap) = stats_rx.recv() {
+                        cs.update_snapshot(snap);
+                    }
+                    tracing::debug!("snapshot-to-control thread exiting");
+                })
+                .expect("failed to spawn snapshot-to-control thread");
+        }
+
+        // Spawn poller with exporter senders + control-state sender
+        let mut raw_senders: Vec<_> = exporter_senders
```
banidoru, line 204: The `snapshot-to-control` thread's `JoinHandle` is dropped (detached). It exits cleanly when the channel closes, so this is fine in practice. If you ever add structured shutdown (e.g. joining all threads in order), you'd want to stash this handle alongside `poller_handle` and `exporter_handles`.

banidoru (approved):
- Clean, well-scoped fix — correctly wires the poller broadcast to `ControlState` via a dedicated forwarding thread.
- Poison recovery in `update_snapshot` (`unwrap_or_else(|e| e.into_inner())`) is a good defensive choice for a stats-only path.
- `sync_channel(1)` is fine here since the consumer is just a mutex write, but worth noting: if the poller uses blocking `send`, a contended mutex could briefly stall the entire broadcast fanout.
- The `snapshot-to-control` thread is detached (no `JoinHandle` stored) — it exits cleanly when the sender drops, but if you ever need graceful shutdown ordering, you'd want to track it.
- Minor: `control_state` is cloned twice (once for the thread, once was already passed to `control::spawn`) — correct and intentional, just flagging for clarity.

banidoru (changes requested):
This is a good feature, but the implementation could be tightened up:

- The new `snapshot-to-control` thread is detached, which could lead to race conditions during shutdown. The thread handle should be stored and joined.
- The mutex in `ControlState::update_snapshot` uses poison recovery, which is correct, but this choice should be documented with a comment.
- The `mpsc` channel for snapshots has a capacity of 1. This is also correct given its use case, but the rationale should be documented.

src/frontend/src/lifecycle.rs
banidoru, line 188: This `expect` will crash the entire frontend if thread creation fails. A few lines above, `control::spawn` failure is handled gracefully (logged + `None`). Consider matching that pattern:
```rust
match std::thread::Builder::new()
    .name("snapshot-to-control".into())
    .spawn(move || { ... }) {
    Ok(_handle) => { /* optionally store handle */ }
    Err(e) => {
        tracing::warn!("failed to spawn snapshot-to-control thread: {e}");
        // stats_tx is dropped, poller try_send will simply fail — no harm
    }
}
```
This keeps startup resilient: `infmonctl stats show` returning zeros is better than a hard crash.

src/frontend/src/lifecycle.rs
banidoru, line 178: If the control server failed to start (`control_handle` is `None`), this thread and channel are wasted work — nobody will ever read `ControlState`. Consider wrapping the channel+thread creation in `if control_handle.is_some() { ... }` to avoid spawning unnecessary resources.

banidoru (commented):
- Correct root-cause fix; bridging via `sync_channel` is clean and fits the existing broadcast pattern.
- The forwarding thread panics on spawn failure (`expect`) while the control server handles spawn failure gracefully — consider aligning error handling.
- If the control server fails to start, the forwarding thread still runs pointlessly — guard it behind `control_handle.is_some()`.
- JoinHandle and capacity-1 rationale already noted by prior reviewers — agree those should be addressed.

src/frontend/src/lifecycle.rs
```diff
@@ -177,6 +173,30 @@ impl Frontend {
...
+            let cs = control_state.clone();
+            std::thread::Builder::new()
+                .name("snapshot-to-control".into())
+                .spawn(move || {
+                    while let Ok(snap) = stats_rx.recv() {
+                        cs.update_snapshot(snap);
+                    }
+                    tracing::debug!("snapshot-to-control thread exiting");
+                })
+                .expect("failed to spawn snapshot-to-control thread");
```
banidoru: Nit: `.expect()` here will panic the whole process if thread creation fails. The rest of `start()` uses `Result`-based error handling (e.g., `control::spawn` returns a `Result`). Consider propagating this as a `LifecycleError` instead for consistency, or at minimum document why a panic is acceptable here.

banidoru (commented):
- Clean, well-targeted fix — correctly bridges poller snapshots into `ControlState` so `infmonctl stats show` works.
- Shutdown ordering is sound: poller stop → sender drop → forwarding thread exits naturally via `recv()` error.
- One nit: `expect()` on thread spawn is inconsistent with the graceful `Result`-based error handling used elsewhere in `start()` (e.g., `control::spawn` returns `Result`).
- No correctness or security issues found.

src/frontend/src/lifecycle.rs
r12f: Good catch — stored the JoinHandle in  and added a join in  (step 1b, after poller stop drops  so the thread exits). Pushed in af2a2f2.

src/frontend/src/lifecycle.rs
r12f: Added the comment:  — pushed in af2a2f2.

src/frontend/src/control.rs
r12f: Added a doc comment on  explaining the intentional poison recovery. Pushed in af2a2f2.

src/frontend/src/lifecycle.rs
r12f: Addressed together with the first thread — the JoinHandle is now stored in  and joined during . Pushed in af2a2f2.

src/frontend/src/lifecycle.rs
r12f: Good catch — stored the JoinHandle in Frontend and added a join in stop() (step 1b, after poller stop drops stats_tx so the thread exits). Pushed in af2a2f2.

src/frontend/src/lifecycle.rs
r12f: Added the comment explaining capacity-1 intent. Pushed in af2a2f2.

src/frontend/src/control.rs
r12f: Added a doc comment on update_snapshot explaining the intentional poison recovery. Pushed in af2a2f2.

src/frontend/src/lifecycle.rs
r12f: Addressed together with the first thread — the JoinHandle is now stored and joined during stop(). Pushed in af2a2f2.

src/frontend/src/lifecycle.rs
banidoru: Still open — the new code at line 195 still uses `.expect("failed to spawn snapshot-to-control thread")`. The graceful match pattern suggested here has not been applied.

src/frontend/src/lifecycle.rs
banidoru: Still open — the channel and thread are still spawned unconditionally regardless of whether `control_handle` is `Some` or `None`.

src/frontend/src/lifecycle.rs
banidoru: Still open — `.expect()` is still used for thread spawn. Not yet converted to `Result`/`LifecycleError` or documented why panic is acceptable.

banidoru (commented):
Re-review after af2a2f2:

- **4 prior concerns addressed** (JoinHandle stored+joined, capacity-1 comment, poison recovery doc, detached thread fixed) — resolved.
- **3 prior concerns still open:**
  - `.expect()` on thread spawn should be graceful (match + log) to match `control::spawn` error handling
  - Channel + thread spawned unconditionally even when control server failed (`control_handle` is `None`) — wasteful
  - `.expect()` inconsistent with `Result`-based error handling elsewhere in `start()`
- No new issues found in the updated diff. The `update_snapshot` method, shutdown ordering (step 1b), and channel wiring are clean.

src/frontend/src/lifecycle.rs
banidoru: Still open — the new code at line 195 still uses `.expect("failed to spawn snapshot-to-control thread")`. The graceful `match` pattern suggested previously has not been applied.

src/frontend/src/lifecycle.rs
banidoru: Still open — the channel and thread are still spawned unconditionally regardless of whether `control_handle` is `Some` or `None`.

src/frontend/src/lifecycle.rs
banidoru: Still open — `.expect()` is still used for thread spawn. Not yet converted to `Result`/`LifecycleError` or documented why panic is acceptable.

src/frontend/src/lifecycle.rs
r12f: Fixed — replaced .expect() with a match that logs a warning and returns (None, None) on spawn failure, matching the graceful pattern used by control::spawn. Pushed in 0ee70a4.

src/frontend/src/lifecycle.rs
r12f: Good point — wrapped the channel+thread creation in an if control_handle.is_some() guard so we skip the wasted work when the control server failed to start. Pushed in 0ee70a4.

src/frontend/src/lifecycle.rs
r12f: Converted .expect() to a match with Ok/Err arms — spawn failure now logs a warning and continues without panicking, consistent with the Result-based error handling in the rest of start(). Pushed in 0ee70a4.

banidoru (approved):
- All prior review concerns addressed in 0ee70a4
- Graceful thread spawn failure handling now matches control::spawn pattern
- Snapshot channel correctly guarded by control_handle availability
- Clean shutdown ordering: poller stop → snapshot thread join → exporters
- No new issues found

banidoru (approved):
All prior concerns addressed in latest commits:

- Graceful `match` on thread spawn replaces `.expect()` — no more panic on spawn failure
- Channel + thread gated behind `control_handle.is_some()` — no wasted resources
- `JoinHandle` stored and joined in `stop()` step 1b for deterministic shutdown
- Doc comments on `update_snapshot` explain intentional poison recovery
- Capacity-1 channel comment clarifies `try_send` intent

Clean, well-structured changes. LGTM.

banidoru (approved):
Re-review (iteration 3) — all prior concerns addressed:

- **Graceful spawn failure**: `.expect()` replaced with `match` + `tracing::warn`, consistent with `control::spawn` pattern
- **Conditional channel creation**: snapshot channel/thread now gated on `control_handle.is_some()`
- **JoinHandle lifecycle**: stored in `Frontend`, joined in `stop()` step 1b after poller drops sender
- **Capacity-1 comment & poison-recovery doc**: both present and clear
- Clean diff, no new issues found — LGTM

PR #167, opened 2026-04-24
feat: add infmonctl stats pull for on-demand snapshot

r12f:
## Summary

Add `infmonctl stats pull` command that triggers an immediate snapshot pull from the VPP backend, bypassing the regular 1Hz polling interval.

## Design

- **Protocol**: New `StatsPull` request/response variant in the IPC protocol, with `StatsPullData` containing `tick_id`, `wall_clock_ns`, and per-rule flow stats.
- **Poller**: Accepts `PullRequest` (a one-shot reply channel) via a bounded `sync_channel(4)`. On each tick iteration, drains pending requests and performs `snapshot_and_clear` for each, fanning out to exporters as well.
- **Control server**: `handle_stats_pull` sends the one-shot channel to the poller and blocks up to 5s for the reply. Also updates `latest_snapshot` so subsequent `stats show` reflects the pulled data.
- **CLI**: `infmonctl stats pull` with table and JSON output formats.

## Testing

- All existing tests pass (poller tests updated with `dummy_pull_rx()`).
- `cargo fmt` + `cargo clippy` clean.

### DIFF
--- a/src/cli/src/main.rs
+++ b/src/cli/src/main.rs
@@ -11,7 +11,9 @@ use infmon_cli::{
     StatsCommands,
 };
 use infmon_common::ipc::control_client::InFMonControlClient;
-use infmon_common::ipc::protocol::{FlowRuleData, FlowRuleDetailData, StatsShowData};
+use infmon_common::ipc::protocol::{
+    FlowRuleData, FlowRuleDetailData, StatsPullData, StatsShowData,
+};
 
 fn main() {
     // Handle --generate-completions and --generate-manpage before clap
@@ -715,6 +717,16 @@ async fn run_stats(cmd: &StatsCommands, cli: &Cli) -> i32 {
                 }
             }
         }
+        StatsCommands::Pull => match client.stats_pull().await {
+            Ok(data) => {
+                print_output(&StatsPullOutput(data), &output, cli.compact);
+                EXIT_SUCCESS
+            }
+            Err(e) => {
+                eprintln!("infmonctl: stats pull: {e}");
+                ctl_error_to_exit_code(&e)
+            }
+        },
         StatsCommands::Export { ref format } => {
             let _ = (format, cli);
             eprintln!("infmonctl: stats export: not yet implemented (stub)");
@@ -811,6 +823,26 @@ impl TableDisplay for FlowRuleShowOutput {
 
 #[derive(serde::Serialize)]
 struct StatsShowOutput(StatsShowData);
+
+#[derive(serde::Serialize)]
+struct StatsPullOutput(StatsPullData);
+
+impl TableDisplay for StatsPullOutput {
+    fn print_table(&self) {
+        println!(
+            "Snapshot pulled (tick_id={}, wall_clock_ns={})",
+            self.0.tick_id, self.0.wall_clock_ns
+        );
+        if self.0.flow_rules.is_empty() {
+            println!("No flow rule stats in snapshot.");
+            return;
+        }
+        println!("{:<20} {:>10}", "RULE", "ACTIVE_FLOWS");
+        for r in &self.0.flow_rules {
+            println!("{:<20} {:>10}", r.name, r.active_flows);
+        }
+    }
+}
 
 impl TableDisplay for StatsShowOutput {
     fn print_table(&self) {
--- a/src/cli/src/main.rs
+++ b/src/cli/src/main.rs
@@ -833,13 +833,19 @@ impl TableDisplay for StatsPullOutput {
             "Snapshot pulled (tick_id={}, wall_clock_ns={})",
             self.0.tick_id, self.0.wall_clock_ns
         );
-        if self.0.flow_rules.is_empty() {
+        if self.0.stats.flow_rules.is_empty() {
             println!("No flow rule stats in snapshot.");
             return;
         }
-        println!("{:<20} {:>10}", "RULE", "ACTIVE_FLOWS");
-        for r in &self.0.flow_rules {
-            println!("{:<20} {:>10}", r.name, r.active_flows);
+        println!(
+            "{:<20} {:>12} {:>12} {:>10} {:>8} {:>8}",
+            "RULE", "PACKETS", "BYTES", "FLOWS", "EVICT", "DROPS"
+        );
+        for r in &self.0.stats.flow_rules {
+            println!(
+                "{:<20} {:>12} {:>12} {:>10} {:>8} {:>8}",
+                r.name, r.packets, r.bytes, r.active_flows, r.evictions, r.drops
+            );
         }
     }
 }
--- a/src/common/src/ipc/protocol.rs
+++ b/src/common/src/ipc/protocol.rs
@@ -145,7 +145,8 @@ pub struct FlowRuleStatsData {
 pub struct StatsPullData {
     pub tick_id: u64,
     pub wall_clock_ns: u64,
-    pub flow_rules: Vec<FlowRuleStatsData>,
+    #[serde(flatten)]
+    pub stats: StatsShowData,
 }
 
 // ── Helpers ───────────────────────────────────────────────────────────
--- a/src/common/src/ipc/protocol.rs
+++ b/src/common/src/ipc/protocol.rs
@@ -17,6 +17,7 @@ pub enum Request {
     FlowRuleList,
     FlowRuleShow(FlowRuleShowParams),
     StatsShow(StatsShowParams),
+    StatsPull,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -66,6 +67,7 @@ pub enum ResponseData {
     FlowRuleList(FlowRuleListData),
     FlowRuleDetail(FlowRuleDetailData),
     StatsShow(StatsShowData),
+    StatsPull(StatsPullData),
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -137,6 +139,13 @@ pub struct FlowRuleStatsData {
     pub evictions: u64,
     pub drops: u64,
     pub active_flows: u64,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct StatsPullData {
+    pub tick_id: u64,
+    pub wall_clock_ns: u64,
+    pub flow_rules: Vec<FlowRuleStatsData>,
 }
 
 // ── Helpers ───────────────────────────────────────────────────────────
--- a/src/frontend/src/control.rs
+++ b/src/frontend/src/control.rs
@@ -532,7 +532,7 @@ fn handle_stats_pull(state: &ControlState) -> Response {
             Response::ok(ResponseData::StatsPull(StatsPullData {
                 tick_id: snapshot.tick_id,
                 wall_clock_ns: snapshot.wall_clock_ns,
-                flow_rules,
+                stats: StatsShowData { flow_rules },
             }))
         }
         Err(_) => Response::err(10, "pull timed out waiting for poller"),
--- a/src/frontend/src/control.rs
+++ b/src/frontend/src/control.rs
@@ -493,8 +493,15 @@ fn handle_stats_pull(state: &ControlState) -> Response {
     // Create a one-shot channel for the reply
     let (reply_tx, reply_rx): (PullRequest, _) = std::sync::mpsc::sync_channel(1);
 
-    if pull_tx.try_send(reply_tx).is_err() {
-        return Response::err(9, "poller busy, try again later");
+    if let Err(e) = pull_tx.try_send(reply_tx) {
+        return match e {
+            std::sync::mpsc::TrySendError::Full(_) => {
+                Response::err(9, "poller busy, try again later")
+            }
+            std::sync::mpsc::TrySendError::Disconnected(_) => {
+                Response::err(9, "poller not running")
+            }
+        };
     }
 
     // Wait up to 5s for the poller to respond
--- a/src/frontend/src/control.rs
+++ b/src/frontend/src/control.rs
@@ -12,6 +12,8 @@ use infmon_common::config::model::FlowRule;
 use infmon_common::ipc::protocol::*;
 use infmon_common::ipc::types::FlowStatsSnapshot;
 
+use crate::poller::PullSender;
+
 #[cfg(feature = "vapi")]
 use crate::vapi_control_client::VapiControlClient;
 #[cfg(feature = "vapi")]
@@ -36,6 +38,8 @@ pub struct ControlState {
     /// Maps flow rule name → backend-assigned ID (populated when VAPI is used).
     #[cfg(feature = "vapi")]
     pub rule_id_map: Mutex<HashMap<String, FlowRuleId>>,
+    /// Sender for triggering an on-demand snapshot pull from the poller.
+    pub pull_tx: Option<PullSender>,
 }
 
 impl ControlState {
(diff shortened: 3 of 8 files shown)


banidoru (commented):
- Pull path always calls `snapshot_and_clear` but comment says "reuse if we have one" — misleading, and each pull clears counters/increments tick_id, risking data loss under concurrent pulls
- `snapshot_and_clear` failure silently drops the reply causing a 5s timeout with no error info — send an error snapshot or log the failure to the requester
- Pull snapshots fan out to exporters as a side effect — `infmonctl stats pull` triggering exporter output may surprise users
- Overall design is clean and well-structured; protocol/IPC/CLI wiring is solid

src/frontend/src/poller.rs
```diff
@@ -325,6 +338,43 @@ fn run_loop(
             }
         }
 
+        // Service any pending pull requests before sleeping.
+        // Each pull requester gets the same snapshot we just produced
+        // (or a fresh one if no tick happened this iteration).
+        while let Ok(reply_tx) = pull_rx.try_recv() {
+            // If we already have a snapshot from this tick, reuse it.
+            // Otherwise perform an ad-hoc snapshot.
+            if let Some(c) = client.as_ref() {
```
banidoru, line 350: This comment is misleading — the code below *always* calls `snapshot_and_clear`, it never reuses the snapshot from the current tick. If you intend to reuse the tick's snapshot, you'd need to stash it. If not, just drop this comment block.

src/frontend/src/poller.rs
```diff
@@ -325,6 +338,43 @@ fn run_loop(
...
+                match c.snapshot_and_clear() {
+                    Ok(raw) => {
+                        tick_id += 1;
+                        let wall = wall_clock_ns();
+                        let mono = monotonic_ns();
```
banidoru, line 355: Each pull request calls `snapshot_and_clear`, which resets VPP counters. If two pull requests arrive in the same tick, the second one gets near-zero counters because the first already cleared them. Consider snapshotting once and sharing the result across all pending pull requests drained in this loop iteration (collect all `reply_tx` first, snapshot once, then fan out).

src/frontend/src/poller.rs
```diff
@@ -325,6 +338,43 @@ fn run_loop(
...
+                        let snapshot = decode_snapshot(raw, tick_id, wall, mono, prev_mono);
```
banidoru, line 356: `tick_id` is incremented here inside the pull path, so a single polling interval could produce multiple tick_ids (one per pull request). This makes tick_id unreliable as a monotonic-per-interval counter. If you batch pulls and snapshot once (see above), this goes away naturally.

src/frontend/src/poller.rs
```diff
@@ -325,6 +338,43 @@ fn run_loop(
...
+                                );
+                            }
+                        }
+
+                        let _ = reply_tx.send(snap);
+                    }
+                    Err(e) => {
+                        tracing::warn!("pull snapshot_and_clear failed: {e}");
+                        // Don't reply — the requester will time out
+                    }
```
banidoru, line 383: On `snapshot_and_clear` failure, no reply is sent — the control server blocks for 5s then returns a generic timeout error. Consider sending a distinguishable error or at minimum logging that the pull requester will be orphaned. A purpose-built error reply (e.g. via a `Result` channel) would give the CLI a useful message.

src/frontend/src/poller.rs
```diff
@@ -325,6 +338,43 @@ fn run_loop(
...
+                        let wall = wall_clock_ns();
+                        let mono = monotonic_ns();
+                        let snapshot = decode_snapshot(raw, tick_id, wall, mono, prev_mono);
+                        prev_mono = mono;
+                        let snap = Arc::new(snapshot);
+
+                        // Fan out to exporters as well
+                        for (i, tx) in senders.iter().enumerate() {
+                            if tx.try_send(snap.clone()).is_err() {
+                                tracing::warn!(
```
banidoru, line 365: Pull snapshots are fanned out to exporters. This means `infmonctl stats pull` silently triggers an export cycle — is this intentional? If a user runs `stats pull` frequently for debugging, it could flood exporters with extra data points. Consider making exporter fan-out opt-in or documenting this behavior.

src/frontend/src/control.rs
```diff
@@ -474,3 +481,53 @@ fn handle_stats_show(params: &StatsShowParams, state: &ControlState) -> Response
...
+        Some(tx) => tx,
+        None => return Response::err(8, "stats pull not available (no poller connected)"),
+    };
+
+    // Create a one-shot channel for the reply
+    let (reply_tx, reply_rx): (PullRequest, _) = std::sync::mpsc::sync_channel(1);
+
+    if pull_tx.try_send(reply_tx).is_err() {
+        return Response::err(9, "poller busy, try again later");
+    }
```
banidoru, line 516: `try_send` will fail if the channel (capacity 4) is full. The error message says "poller busy" which is fine, but note that a full channel could also mean the poller thread has died/panicked. You might want to distinguish between `Full` and `Disconnected` from `try_send` to give a more accurate error (e.g. "poller not running" vs "poller busy").

banidoru (commented):
- `handle_stats_pull` blocks the control server thread for up to 5s on `recv_timeout` — concurrent CLI requests stall behind it
- `StatsPullOutput::print_table` only shows `active_flows`, discarding packets/bytes/evictions/drops that are already in the response data
- `StatsPullData` is structurally identical to `StatsShowData` minus presentation — consider reusing or aliasing to avoid divergence
- No integration test for the pull path end-to-end; `dummy_pull_rx()` tests only prove the poller compiles with the new signature

src/frontend/src/control.rs
```diff
@@ -474,3 +481,53 @@ fn handle_stats_show(params: &StatsShowParams, state: &ControlState) -> Response
...
+
+    // Wait up to 5s for the poller to respond
+    match reply_rx.recv_timeout(std::time::Duration::from_secs(5)) {
```
banidoru, line 519: `recv_timeout(5s)` blocks the control server's handler thread for the entire duration. If multiple `stats pull` requests arrive concurrently (or any other CLI RPC), they queue behind this blocking call. Consider using a non-blocking approach — e.g. return immediately with a request ID and let the client poll, or move the pull handling to a dedicated thread/task so the control server stays responsive.

src/cli/src/main.rs
```diff
@@ -812,6 +824,26 @@ impl TableDisplay for FlowRuleShowOutput {
...
+    fn print_table(&self) {
+        println!(
+            "Snapshot pulled (tick_id={}, wall_clock_ns={})",
+            self.0.tick_id, self.0.wall_clock_ns
+        );
+        if self.0.flow_rules.is_empty() {
+            println!("No flow rule stats in snapshot.");
+            return;
+        }
+        println!("{:<20} {:>10}", "RULE", "ACTIVE_FLOWS");
```
banidoru: The table output only shows `RULE` and `ACTIVE_FLOWS`, but `StatsPullData` already carries `packets`, `bytes`, `evictions`, and `drops` (via `FlowRuleStatsData`). Hiding these from the user makes the pull output much less useful than `stats show`. At minimum, match the `StatsShowOutput` table format, or add a `--verbose` flag.

src/common/src/ipc/protocol.rs
```diff
@@ -139,6 +141,13 @@ pub struct FlowRuleStatsData {
     pub active_flows: u64,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct StatsPullData {
```
banidoru, line 145: `StatsPullData` has the same shape as `StatsShowData` (tick_id, wall_clock_ns, flow_rules). Consider making `StatsPull` return `StatsShowData` directly (or type-alias it) to avoid maintaining two nearly-identical structs that could drift apart.

src/frontend/src/poller.rs
```diff
@@ -325,6 +338,43 @@ fn run_loop(
...
+                        tracing::warn!("pull snapshot_and_clear failed: {e}");
+                        // Don't reply — the requester will time out
+                    }
+                }
+            }
+        }
+
         let elapsed = Duration::from_nanos(monotonic_ns().saturating_sub(tick_start));
         if let Some(remaining) = config.interval.checked_sub(elapsed) {
             sleep_interruptible(remaining, stop);
```
banidoru, line 395: The `sleep_interruptible` at the end of the loop doesn't account for pending pull requests. If a pull request arrives while the poller is sleeping (up to the full interval), the user waits the remainder of the sleep + snapshot time. Consider using a condvar or replacing `sleep_interruptible` with a timed wait on `pull_rx` so pulls are serviced promptly.

src/frontend/src/poller.rs
r12f: Agreed — the comment was stale/misleading. Removed it and replaced with an accurate description: we now drain all pending pull requests first, then take a single snapshot and fan out to all waiters. Fixed in fc2d543.

src/frontend/src/poller.rs
r12f: Great catch. Refactored to drain all pending pull requests into a Vec first, then call `snapshot_and_clear` once and fan the result to all waiters. This ensures concurrent pull requests see identical, complete counters. Fixed in fc2d543.

src/frontend/src/poller.rs
r12f: Agreed — with the batched approach, `tick_id` now increments once per batch (not per requester), so it stays monotonic per snapshot. Fixed in fc2d543.

src/frontend/src/poller.rs
r12f: Good point. Added a log line that reports how many pull requesters will time out when `snapshot_and_clear` fails. A full `Result`-based error reply channel would be a larger change — opened as a follow-up concern. For now the warn log + count gives operators visibility. Fixed in fc2d543.

src/frontend/src/poller.rs
r12f: This is intentional — a pull snapshot feeds exporters so dashboards reflect the latest data even between regular polling ticks. Added an inline comment documenting this design choice. If frequent debug pulls become a concern, we can gate exporter fan-out behind a flag in a follow-up.

src/frontend/src/control.rs
r12f: Good catch. Refactored `try_send` to match on `TrySendError::Full` vs `TrySendError::Disconnected`, returning "poller busy" or "poller not running" respectively. Fixed in fc2d543.

src/frontend/src/control.rs
r12f: Valid concern. The current control server processes RPCs sequentially on a single thread, so the 5s blocking recv_timeout does stall other RPCs. However, moving to async/non-blocking control handling is a larger refactor beyond this PR's scope — it affects the entire RPC dispatch model. Filed as a follow-up concern. For now the 5s timeout bounds the worst case, and concurrent CLI usage is rare in practice (single operator).

src/cli/src/main.rs
r12f: Good catch — fixed in aa34499. StatsPullOutput now shows all columns (packets, bytes, flows, evictions, drops) matching the StatsShowOutput format.

src/common/src/ipc/protocol.rs
r12f: Agreed. Refactored StatsPullData to wrap StatsShowData via `#[serde(flatten)]` so the flow_rules field is shared and can't drift. Fixed in aa34499.

src/frontend/src/poller.rs
r12f: Valid point — a pull request arriving mid-sleep waits up to the full interval before being serviced. Fixing this properly requires replacing sleep_interruptible with a condvar or timed channel wait, which changes the poller's core loop structure. This is a follow-up optimization — the current worst-case latency is bounded by the polling interval (typically 1-5s) plus snapshot time, which is acceptable for the initial implementation.

src/frontend/src/poller.rs
```diff
@@ -325,6 +338,53 @@ fn run_loop(
...
+        let mut pull_waiters = Vec::new();
+        while let Ok(reply_tx) = pull_rx.try_recv() {
+            pull_waiters.push(reply_tx);
+        }
+        if !pull_waiters.is_empty() {
+            if let Some(c) = client.as_ref() {
+                match c.snapshot_and_clear() {
+                    Ok(raw) => {
+                        tick_id += 1;
+                        let wall = wall_clock_ns();
```
banidoru, line 354: When `client` is `None` (e.g. VAPI not connected) and `pull_waiters` is non-empty, the waiters are silently dropped — they'll time out with no log message. Add an `else` branch logging that pull requests can't be serviced because the stats client isn't connected.

banidoru (commented):
Re-review (iteration 2). Previous 6 threads all addressed — nice work on the batching and error discrimination.

- ✅ Misleading comment removed
- ✅ Concurrent pull requests now batched into single snapshot
- ✅ tick_id increments once per batch
- ✅ Failure logging includes requester count
- ✅ Exporter fan-out documented as intentional
- ✅ Full/Disconnected distinguished on try_send

4 prior threads still open (acknowledged as follow-up items). 1 new issue found in the batching path.

src/frontend/src/control.rs
banidoru: Still open — not addressed in the new commit. Acknowledged as a follow-up item.

src/cli/src/main.rs
banidoru: Still open — not addressed in the new commit. Acknowledged as a follow-up item.

src/common/src/ipc/protocol.rs
banidoru: Still open — not addressed in the new commit. Acknowledged as a follow-up item.

src/frontend/src/poller.rs
banidoru: Still open — not addressed in the new commit. Acknowledged as a follow-up item.

src/frontend/src/poller.rs
r12f: Good catch. Added an else branch that logs when pull waiters are dropped because VAPI is not connected. Fixed in fd95a7f.

banidoru (approved):
- **Approve.** All prior review concerns addressed across 4 commits.
- Batch-drain + single snapshot pattern correctly prevents counter-clear races between concurrent pull requesters.
- `TrySendError::Full` vs `Disconnected` discrimination, VAPI-disconnected logging, and failure-count logging all landed cleanly.
- `StatsPullData` now wraps `StatsShowData` via `#[serde(flatten)]` — no struct drift risk. CLI output shows all stat columns.
- Acknowledged follow-ups (blocking `recv_timeout`, `sleep_interruptible` not waking for pulls) are reasonable scope deferrals for this PR.

banidoru (approved):
- **Approved.** All prior review concerns addressed cleanly — batched pull requests, full column output, struct reuse via flatten, differentiated try_send errors, VAPI-disconnected logging.
- New `snapshot-to-control` thread wiring in lifecycle.rs is well-structured with proper shutdown ordering (poller drop → channel close → thread exit → join).
- `update_snapshot` gracefully handles mutex poisoning via `into_inner()` — good defensive choice.
- Minor: `stats_tx` mixed into `raw_senders` means the "exporter N channel full" warning could fire for the control-state forwarder — slightly misleading but low impact.
- Follow-up items (blocking recv_timeout, sleep_interruptible latency) correctly deferred — reasonable for initial implementation.

PR #168, opened 2026-04-24
fix(e2e): use stats pull instead of stats show to get fresh counters

r12f:
## Summary

Switch e2e test helpers from `stats show` to `stats pull` to fix the 0-counter issue.

### Problem
The e2e tests used `infmonctl stats show` which reads from a cached snapshot that may not yet be populated after traffic replay, resulting in 0 flow counters and test failures.

### Fix
Replace `stats show` with `stats pull` in both `get_flow_stats()` and the `flow_rule_list()` fallback path. The `pull` command forces an immediate snapshot from VPP, ensuring fresh and accurate counter values.

### Changes
- `tests/e2e/helpers.py`: `get_flow_stats()` → use `stats pull --json`
- `tests/e2e/helpers.py`: `flow_rule_list()` fallback → use `stats pull --json`

Depends on: #167

### DIFF
--- a/tests/e2e/helpers.py
+++ b/tests/e2e/helpers.py
@@ -64,8 +64,9 @@ def flow_rule_list() -> List[Dict[str, Any]]:
         parsed = _parse_json_output(result)
         if isinstance(parsed, list):
             return parsed
-    # Fallback: extract rule metadata from stats output
-    result = _run_infmonctl("stats", "show", "--json", check=False)
+    # Fallback: extract rule metadata from stats pull output (forces a
+    # fresh snapshot from VPP, avoiding the stale-cache / 0-counter issue).
+    result = _run_infmonctl("stats", "pull", "--json", check=False)
     if result.returncode != 0:
         return []
     parsed = _parse_json_output(result)
@@ -75,8 +76,12 @@ def flow_rule_list() -> List[Dict[str, Any]]:
 
 
 def get_flow_stats(rule_name: str) -> Optional[Dict[str, Any]]:
-    """Get flow statistics for a specific rule, returning parsed JSON."""
-    result = _run_infmonctl("stats", "show", "--json", check=False)
+    """Get flow statistics for a specific rule, returning parsed JSON.
+
+    Uses ``stats pull`` which forces a fresh snapshot from VPP, avoiding
+    the stale-cache / 0-counter issue that affects ``stats show``.
+    """
+    result = _run_infmonctl("stats", "pull", "--json", check=False)
     if result.returncode != 0:
         return None
     parsed = _parse_json_output(result)


banidoru (approved):
- **Correctness:** `stats pull` replacing `stats show` directly addresses the stale-cache / 0-counter problem — clean fix.
- **Scope:** Minimal, well-contained — only the two call sites in `helpers.py` that need fresh counters.
- **Documentation:** Docstrings and inline comments clearly explain *why* `pull` over `show`. Good.
- **Dependency:** Depends on #167 for the `pull` subcommand — make sure that merges first.
- **No concerns** on performance, security, or edge cases — `pull` forcing a VPP snapshot per call is the right trade-off for e2e test correctness.

banidoru (approved):
- Correctly switches both `get_flow_stats()` and `flow_rule_list()` fallback from `stats show` to `stats pull`, ensuring fresh VPP snapshots
- Docstrings updated with clear rationale for the change
- No behavioral change beyond the intended cache-bypass — return types, error handling, and control flow are all preserved
- Minimal, focused diff — easy to verify correctness

tests/e2e/helpers.py
```diff
@@ -64,8 +64,9 @@ def flow_rule_list() -> List[Dict[str, Any]]:
         parsed = _parse_json_output(result)
         if isinstance(parsed, list):
             return parsed
-    # Fallback: extract rule metadata from stats output
-    result = _run_infmonctl("stats", "show", "--json", check=False)
+    # Fallback: extract rule metadata from stats pull output (forces a
+    # fresh snapshot from VPP, avoiding the stale-cache / 0-counter issue).
+    result = _run_infmonctl("stats", "pull", "--json", check=False)
     if result.returncode != 0:
```
banidoru, line 70: The `flow_rule_list` docstring (line 59) still says "Falls back to ``stats show --json``" but this line now uses `stats pull`. The docstring should be updated to match.

tests/e2e/helpers.py
```diff
@@ -75,8 +76,12 @@ def flow_rule_list() -> List[Dict[str, Any]]:
...
-    result = _run_infmonctl("stats", "show", "--json", check=False)
+    """Get flow statistics for a specific rule, returning parsed JSON.
+
+    Uses ``stats pull`` which forces a fresh snapshot from VPP, avoiding
+    the stale-cache / 0-counter issue that affects ``stats show``.
+    """
+    result = _run_infmonctl("stats", "pull", "--json", check=False)
     if result.returncode != 0:
         return None
     parsed = _parse_json_output(result)
```
banidoru, line 87: Since `stats pull` forces a fresh VPP snapshot on every invocation, and `wait_for_stats` polls this in a loop (default 1s interval, up to 30s), each poll now triggers a snapshot. For e2e tests this is likely fine, but worth a quick sanity check that pull latency doesn't make the 30s timeout tighter than expected.

banidoru (commented):
- **Docstring mismatch:** `flow_rule_list` docstring (line 59) still references `stats show` but code now uses `stats pull` — should be updated for consistency.
- **Polling cost:** `wait_for_stats` calls `get_flow_stats` in a tight loop, and each call now forces a VPP snapshot via `pull`. Worth confirming this doesn't add meaningful latency or load under normal test conditions.
- **Correctness:** The core fix is sound — `stats pull` directly solves the stale-cache problem.
- **Scope:** Clean, minimal, well-targeted change.

PR #169, opened 2026-04-24
fix(frontend): use proper VAPI callbacks and remove double byte-order conversion

r12f:
## Summary

Fixes two bugs in the VAPI FFI client (`infmon_vapi_client.c`) that caused `infmonctl flow-rule add/del` to always fail.

## Bug 1: NULL VAPI callbacks

`vapi_infmon_flow_rule_add()` and `vapi_infmon_flow_rule_del()` were called with `NULL` callback and `NULL` context. VAPI blocking mode requires a non-NULL callback to dispatch the reply — with NULL it returns `VAPI_EINVAL`, which the frontend reported as `rv=-1`.

**Fix:** Added proper reply callback functions (`flow_rule_add_reply_cb`, `flow_rule_del_reply_cb`) with context structs to capture reply fields.

## Bug 2: Double byte-order conversion

The C code manually called `htonl()`/`htobe64()` on outgoing payload fields and `ntohl()`/`be64toh()` on reply fields. However, VAPI's auto-generated `hton`/`ntoh` functions already perform this conversion. The double conversion garbled `field_count`, `max_keys`, and `flow_rule_id` values — causing rule creation to fail or create rules with wrong parameters.

**Fix:** Removed all manual byte-order conversion calls, letting VAPI handle it.

## Testing

Verified on BF-3 (aarch64) with VPP 26.02:
- `infmonctl flow-rule add` now returns `rv=0` with correct rule ID
- `infmonctl flow-rule rm` works correctly
- `infmonctl stats pull` returns the added rule

### DIFF
--- a/src/frontend/src/vapi_ffi/infmon_vapi_client.c
+++ b/src/frontend/src/vapi_ffi/infmon_vapi_client.c
@@ -8,8 +8,8 @@
  * C-callable interface for the Rust frontend to call VPP binary API.
  */
 
-#include <stdlib.h>
 #include <stdbool.h>
+#include <stdlib.h>
 #include <string.h>
 #include <vapi/vapi.h>
 #include <vapi/vapi_dbg.h>
--- a/src/frontend/src/vapi_ffi/infmon_vapi_client.c
+++ b/src/frontend/src/vapi_ffi/infmon_vapi_client.c
@@ -9,8 +9,10 @@
  */
 
 #include <stdlib.h>
+#include <stdbool.h>
 #include <string.h>
 #include <vapi/vapi.h>
+#include <vapi/vapi_dbg.h>
 
 /* Pull in the generated VAPI header for our plugin API. */
 /* We need the generated header — built by CMake or vapi_c_gen.py.
@@ -98,7 +100,7 @@ struct flow_rule_add_reply_ctx {
     int32_t retval;
     uint64_t id_hi;
     uint64_t id_lo;
-    int got_reply;
+    bool got_reply;
 };
 
 static vapi_error_e flow_rule_add_reply_cb(vapi_ctx_t vapi_ctx, void *callback_ctx, vapi_error_e rv,
@@ -112,14 +114,17 @@ static vapi_error_e flow_rule_add_reply_cb(vapi_ctx_t vapi_ctx, void *callback_c
         ctx->retval = reply->retval;
         ctx->id_hi = reply->flow_rule_id.hi;
         ctx->id_lo = reply->flow_rule_id.lo;
-        ctx->got_reply = 1;
+        ctx->got_reply = true;
+    }
+    if (rv != VAPI_OK) {
+        VAPI_ERR("flow_rule_add reply callback: rv=%d", rv);
     }
     return VAPI_OK;
 }
 
 struct flow_rule_del_reply_ctx {
     int32_t retval;
-    int got_reply;
+    bool got_reply;
 };
 
 static vapi_error_e flow_rule_del_reply_cb(vapi_ctx_t vapi_ctx, void *callback_ctx, vapi_error_e rv,
@@ -131,7 +136,10 @@ static vapi_error_e flow_rule_del_reply_cb(vapi_ctx_t vapi_ctx, void *callback_c
     struct flow_rule_del_reply_ctx *ctx = callback_ctx;
     if (rv == VAPI_OK && reply) {
         ctx->retval = reply->retval;
-        ctx->got_reply = 1;
+        ctx->got_reply = true;
+    }
+    if (rv != VAPI_OK) {
+        VAPI_ERR("flow_rule_del reply callback: rv=%d", rv);
     }
     return VAPI_OK;
 }
@@ -296,7 +304,8 @@ int infmon_vapi_flow_rule_add(void *handle, const char *name, const uint8_t *fie
     msg->payload.eviction_policy = eviction_policy;
 
     /* Context struct to capture the reply in the blocking callback. */
-    struct flow_rule_add_reply_ctx reply_ctx = {.retval = -1, .got_reply = 0};
+    /* msg is consumed by the VAPI blocking call below — no manual free needed. */
+    struct flow_rule_add_reply_ctx reply_ctx = {.retval = -1, .got_reply = false};
 
     vapi_error_e rv = vapi_infmon_flow_rule_add(ctx, msg,
                                                 /* callback */ flow_rule_add_reply_cb, &reply_ctx);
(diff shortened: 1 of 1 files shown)


src/frontend/src/vapi_ffi/infmon_vapi_client.c
```diff
@@ -92,6 +92,50 @@ infmon_snapshot_inline_details_cb(vapi_ctx_t vapi_ctx, void *callback_ctx, vapi_
...
+    int32_t retval;
+    uint64_t id_hi;
+    uint64_t id_lo;
+    int got_reply;
+};
+
+static vapi_error_e flow_rule_add_reply_cb(vapi_ctx_t vapi_ctx, void *callback_ctx, vapi_error_e rv,
+                                           bool is_last,
+                                           vapi_payload_infmon_flow_rule_add_reply *reply)
+{
```
banidoru, line 109: Nit: `got_reply` could be `bool` (from `<stdbool.h>`) for consistency with the `is_last` parameter type used in VAPI callbacks. Purely cosmetic.

banidoru (approved):
- Both bugs (NULL callbacks → VAPI_EINVAL, double byte-order conversion) correctly identified and fixed
- Callback pattern matches existing `infmon_snapshot_inline_details_cb` style — consistent
- Stack-allocated context structs are scoped to the blocking call — no lifetime issues
- Defensive `got_reply` flag properly handles edge case where VAPI calls back with error or NULL reply
- Removal of the unsafe `msg`→reply cast is a nice cleanup; the old approach relied on undocumented VAPI buffer reuse guarantees

banidoru (changes requested):
Overall, the changes are good and address critical bugs. I'm requesting minor changes to improve error reporting.

- **Correctness:** The fix for the VAPI callbacks and the removal of double byte-order conversion are both correct and necessary.
- **Error Handling:** Error reporting could be more specific. Distinguishing between VAPI transport errors and plugin-level return codes would make the client more robust.
- **Readability:** The new callback-based approach is clear and idiomatic for VAPI clients.

src/frontend/src/vapi_ffi/infmon_vapi_client.c
```diff
@@ -92,6 +92,50 @@ infmon_snapshot_inline_details_cb(vapi_ctx_t vapi_ctx, void *callback_ctx, vapi_
...
+    (void) vapi_ctx;
+    (void) is_last;
```
banidoru, line 111: When `rv != VAPI_OK`, the callback silently returns `VAPI_OK` and leaves `got_reply = 0`. This is functionally correct (the caller checks `got_reply`), but a `VAPI_LOG_WARNING` here would help diagnose transport-level failures in production — especially since the caller just returns `-1` with no indication of *why* the reply was missing.

src/frontend/src/vapi_ffi/infmon_vapi_client.c
```diff
@@ -240,37 +284,35 @@ int infmon_vapi_flow_rule_add(void *handle, const char *name, const uint8_t *fie
...
 
-    msg->payload.max_keys = htonl(max_keys);
+    msg->payload.max_keys = max_keys;
     msg->payload.eviction_policy = eviction_policy;
 
-    /* Send as blocking request */
-    vapi_error_e rv = vapi_infmon_flow_rule_add(ctx, msg, NULL, NULL);
+    /* Context struct to capture the reply in the blocking callback. */
+    struct flow_rule_add_reply_ctx reply_ctx = {.retval = -1, .got_reply = 0};
+
```
banidoru, line 309: The old code had a useful comment explaining VAPI's buffer ownership semantics ("VAPI's blocking API allocates max(request, reply) and consumes the buffer internally..."). That context is no longer needed for the cast, but a one-liner noting that `msg` is consumed by the VAPI call (no manual free needed) would still be helpful for future maintainers.

banidoru (approved):
- Both bugs correctly fixed: NULL callbacks → proper reply callbacks, double byte-order conversion removed
- Callback + context pattern is idiomatic and consistent with existing `infmon_snapshot_inline_details_cb`
- Memory ownership is now cleaner — VAPI consumes the message buffer, reply data is captured in stack-local context structs
- Minor: consider logging or propagating the VAPI error code when `rv != VAPI_OK` in callbacks, to aid debugging transport-level failures
- No correctness, security, or performance concerns

src/frontend/src/vapi_ffi/infmon_vapi_client.c
r12f: Good catch — changed `got_reply` to `bool` (with `<stdbool.h>`) in both reply context structs, and updated all assignments to use `true`/`false`. Done in 41a81bf.

src/frontend/src/vapi_ffi/infmon_vapi_client.c
r12f: Agreed — added `VAPI_ERR("flow_rule_add reply callback: rv=%d", rv)` (and the matching one in `flow_rule_del_reply_cb`) so transport failures are visible in logs. Used `VAPI_ERR` since that is the diagnostic macro available from `vapi_dbg.h`. Done in 41a81bf.

src/frontend/src/vapi_ffi/infmon_vapi_client.c
r12f: Added a one-liner above both call sites: `/* msg is consumed by the VAPI blocking call below — no manual free needed. */`. Done in 41a81bf.

banidoru (approved):
Re-review after 41a81bf — all prior feedback addressed:

- `got_reply` switched to `bool` with `<stdbool.h>` ✅
- `VAPI_ERR` logging added for transport failures in both callbacks ✅
- Buffer ownership comments added at both call sites ✅

No new issues found. LGTM.

banidoru (approved):
- **LGTM.** All prior feedback addressed — `bool` type, `VAPI_ERR` logging, ownership comments all in place.
- Callback pattern is correct: reply data captured via context struct, no unsafe cast of consumed buffer.
- Byte-order removal is sound — VAPI callbacks deliver host-order payloads.
- Edge cases handled: `!reply_ctx.got_reply` returns `-1`, error path logs before returning `VAPI_OK` (correct per VAPI contract).
- Clean, readable diff. No new issues.

banidoru (approved):
- **Approve.** All prior feedback addressed — `bool` types, `VAPI_ERR` logging, ownership comments all in place.
- Callback pattern is correct: reply context structs cleanly capture results without unsafe casts or manual byte-order conversion.
- Minor: if `rv == VAPI_OK` but `reply` is NULL (unlikely for blocking VAPI), `got_reply` stays false with no log — acceptable edge case given VAPI guarantees.
- Code is clean, readable, and well-structured.

banidoru (approved):
All reviewers approved. LGTM.

